PR #107, opened 2026-04-22
chore(skills): regenerate pp-* skill mirrors

pejmanjohn:
## Problem

`skills/pp-*` is generated plugin output that mirrors the upstream `library/**/SKILL.md` files. The checked-in mirrors had drifted from the library skills, so people installing the plugin could receive older agent guidance than the guidance maintained beside each CLI.

That matters because these skills are the user-facing entrypoints for the plugin. In this drift, the focused skills were missing newer operational guidance such as agent-mode behavior, output filtering with `--select`, stale `@latest` install fallback text, and response-envelope notes for some tools. The repo's own `AGENTS.md` says these mirrors should be regenerated alongside library skill changes, and that plugin content changes need a patch version bump.

## What changed

- Re-ran `go run ./tools/generate-skills/main.go` so all 22 focused `skills/pp-*` mirrors match their upstream library `SKILL.md` sources.
- Committed only the generated mirror output plus the plugin manifest version bump.
- Bumped `.claude-plugin/plugin.json` from `1.2.0` to `1.2.1` for the plugin content update.

## Scope

This is intentionally a generated-output maintenance PR. It does not change any library CLI source, generator behavior, registry metadata, or workflow automation.

The automation gap that made this easy to miss is handled separately in #108, so this PR stays reviewable as the mirror refresh itself.

## Verification

- `cd tools/generate-skills && go test ./...`
- Ran `.github/scripts/verify-skill/verify_skill.py` against all 22 CLI directories with `SKILL.md` files.
- The verifier completed with 0 blocking failures. HubSpot still reports 2 likely false positives for `associations companies` examples, which matches the verifier's documented false-positive pattern.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "printing-press-library",
-  "version": "1.2.0",
+  "version": "1.2.1",
   "description": "Discover, install, and use CLI tools and MCP servers for hundreds of APIs",
   "author": {
     "name": "mvanhorn"
--- a/skills/pp-agent-capture/SKILL.md
+++ b/skills/pp-agent-capture/SKILL.md
@@ -95,6 +95,29 @@ Parse `$ARGUMENTS`:
 
 Run any command with `--help` for full flag documentation.
 
+## Agent Mode
+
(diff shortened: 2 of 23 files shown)


tmchow:
@pejmanjohn this is different now, the skills live in `cli-skills/`

PR #108, opened 2026-04-22
fix(skills): automate pp-* skill sync

pejmanjohn:
## Problem

The repo already documents a manual workaround for keeping plugin skills in sync: when a library `SKILL.md` or a CLI's `internal/cli/**` code changes, contributors are expected to run the skill generator and manually bump the plugin patch version.

Two pieces made that easy to miss:

- `.github/workflows/generate-skills.yml` only ran for registry, manifest, or generator changes. It did not run when the source skill prose or CLI command source changed.
- `tools/generate-skills/main.go` only bumped `.claude-plugin/plugin.json` when the set of `skills/pp-*` directories changed. If the same skill files changed content, the generator could refresh the mirrors without bumping the plugin version.

The result is stale plugin output or stale plugin versions after content-only changes, even though the repository treats `skills/pp-*` as generated mirrors.

## What changed

- Added `library/**/SKILL.md` and `library/**/internal/cli/**` to the generate-skills workflow path filters.
- Replaced the old directory-only version-bump check with a generated-output snapshot that tracks both:
  - the sorted set of `pp-*` skill directories
  - the SHA-256 hash of each generated `SKILL.md`
- Kept the existing behavior for added or removed skill directories.
- Added regression tests for content-only plugin version bumps, snapshot comparison, and the workflow path filters.

## Scope

This PR changes the automation and generator behavior only. It intentionally does not include regenerated `skills/pp-*` output; that generated-output refresh is isolated in #107 so each PR has one job.

## Verification

- `cd tools/generate-skills && go test -count=1 ./...`
- `git diff --check`

The new tests cover the failure mode directly: a `pp-*` skill with the same directory name but different `SKILL.md` content now causes the plugin patch version to bump.

### DIFF
--- a/.github/workflows/generate-skills.yml
+++ b/.github/workflows/generate-skills.yml
@@ -6,6 +6,8 @@ on:
     paths:
       - 'registry.json'
       - 'library/**/.printing-press.json'
+      - 'library/**/SKILL.md'
+      - 'library/**/internal/cli/**'
       - 'tools/generate-skills/**'
 
 jobs:
--- a/tools/generate-skills/main.go
+++ b/tools/generate-skills/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"crypto/sha256"
 	"encoding/json"
 	"fmt"
 	"log"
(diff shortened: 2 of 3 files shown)


tmchow:
@pejmanjohn can you rebase and resolve conflcits?

pejmanjohn:
@tmchow Rebased onto latest main and resolved the conflicts. PR is mergeable now.

PR #170, opened 2026-05-01
chore(wikipedia): library template sweep via regen-merge

tmchow:
Same pattern as #166-#169. 3 stale-template TEMPLATED-WITH-ADDITIONS; took fresh wholesale. Verified: build clean, 87 tests pass, `page get-summary 'Albert Einstein'` returns real summary from live Wikipedia API.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #166, opened 2026-05-01
chore(pypi): library template sweep via regen-merge

tmchow:
## Summary

First end-to-end dogfood of the [\`printing-press regen-merge\`](https://github.com/mvanhorn/cli-printing-press/pull/461) subcommand against a published library CLI. Brings \`developer-tools/pypi\` up to the current generator template.

## What landed (auto-applied by regen-merge)

- Context-aware store: \`OpenWithContext\`, \`withMigrationLock\`, \`execWithBusyRetry\`, \`retryOnBusy\`, \`isSQLiteBusy\`
- New \`internal/cliutil/\` package split: \`probe\`, \`ratelimit\`, \`verifyenv\`
- New \`internal/mcp/cobratree/\` subpackage centralizing shellout-to-CLI registration
- Helper additions: \`dryRunOK\`, \`printJSONFiltered\`, \`LookupFieldValue\`, more
- 32 templated files refreshed; 9 new template files added; 1 stale auth scaffold deleted

## Manual merge decisions (4 TEMPLATED-WITH-ADDITIONS files)

These files had hand-customizations on top of templated content; regen-merge correctly refused to auto-apply them. After review:

| File | Pub addition | Decision |
|---|---|---|
| \`helpers.go\` | \`wrapResultsWithFreshness\` | Dropped — 0 callers, dead code |
| \`store.go\` | \`setSchemaVersion\` | Dropped — fresh performs the same write inline in migrate() |
| \`tools.go\` | \`handleSync\` + shellout helpers | Helpers obsoleted by cobratree refactor. **\`handleSync\` MCP tool dropped** — flagged below |
| \`client.go\` | \`adaptiveLimiter\` (8 decls) | **Dropped** — flagged below |

## Behavior regressions (worth surfacing)

1. **Sync MCP tool removed.** \`handleSync\` was an MCP-callable tool that triggered a sync run. Not load-bearing for a public read-only API; users who want sync can shell out to \`pypi-pp-cli sync\`. If it's actually used by an agent flow, this needs to come back as a cobratree-style auto-registered tool.

2. **Adaptive rate limiter removed.** The hand-written \`adaptiveLimiter\` watched 429 responses and reduced the configured rate when throttled. pypi.org doesn't enforce rate limits, so the static cap from \`--rate-limit\` is sufficient. If reinstated, should probably move into the templated client as a generator-level feature, not a per-CLI customization.

## Verification

- \`go build ./...\` clean
- \`go test ./...\` — 87 pass
[...]

PR #171, opened 2026-05-01
chore(open-meteo): library template sweep via regen-merge

tmchow:
Same pattern as #166-#170. Build/test/vet clean, 86 tests pass.

Surfaces a pre-existing spec gap: `spec.yaml` has no `servers:` block, so `BaseURL` defaults to `https://api.example.com`. Pub had the same default; the binary needs a user `config.toml` override to actually reach open-meteo.com. Spec surgery to add the servers block is a separate follow-up.


tmchow:
## Tighter verification matrix

| Check | Result |
|---|---|
| version | open-meteo-pp-cli 1.0.0 |
| doctor --json (with config override) | config=ok api=reachable |
| agent-context | 14 commands surfaced |
| MCP binary build | clean |
| All 8 top-level commands load `--help` | clean |

Pre-existing limitations (NOT sweep-introduced):
1. `spec.yaml` has no `servers:` section, so `BaseURL` defaults to `https://api.example.com`. Override via `OPEN_METEO_CONFIG=/path/to/config.toml` with `base_url` set to reach the real API.
2. The `forecast` command is wired with hardcoded `/v1/forecast` path and an empty `params` map — the generator didn't translate the spec's query parameters (latitude, longitude, hourly, etc.) into cobra flags. Pub had the same issue. This is a generator-level issue separate from this sweep.

I noted both in the original PR. Sweep is clean for what it covers (template currency).

PR #175, opened 2026-05-01
chore(cal-com): library template sweep via regen-merge

tmchow:
Unblocked by [cli-printing-press#467](https://github.com/mvanhorn/cli-printing-press/pull/467). Auth-resource customization that previously caused this CLI to be deferred is no longer load-bearing — the generator now omits the dead `cal_com_auth` table, so take-fresh-wholesale drops both pub's typed Upsert and the table cleanly. Auth still syncs through the generic resources path.

Two manual fixups beyond the standard 4-file overwrite:
1. Pre-injected the 4 NOVEL analytics sub-command `AddCommand` calls into fresh's `analytics.go` so regen-merge's restoration had a host function to inject into.
2. Same arg-shape fix as #174 (`&flags` -> `flags` on restored `AddCommand` calls in root.go).

Verified: build/test/vet clean (180 tests), doctor OK, `analytics` exposes all 4 NOVEL sub-commands. Real API call deferred (needs CAL_COM_TOKEN).


tmchow:
## Tighter verification — caught and fixed a regression

User-supplied API key let me run a real end-to-end matrix.

**Regression caught:** my take-fresh wholesale on `internal/client/client.go` silently dropped pub's hand-added `requiredHeadersForPath()` call — the cal-api-version header logic from `calcom_versions.go`. The decl-set classifier missed it because the change was inside an existing function body, not a new top-level decl. **This is exactly the "in-place body modification" failure mode the regen-merge plan called out as a known limitation.**

Fix amended into this PR: restored the per-path required-headers loop in `do()`. Without it, every Cal.com v2 endpoint that requires `cal-api-version` (bookings, event-types, schedules, slots, teams, organizations) silently degrades or returns wrong-shape responses.

## Verification matrix (post-fix)

| Check | Result |
|---|---|
| `me get` | ✓ returns Trevin Chow |
| `event-types list` | ✓ 3 event types (15min, 30min, secret) |
| `bookings list` | ✓ 11 bookings returned with correct shape |
| `schedules list` | ✓ 1 schedule (Working Hours) |
| Real booking via CLI | ✓ created 5 bookings (IDs 19059923, 19059960, 19059996, 19060013, 19060038) |
| Drove stats on public links | ✓ HTTP 200 on `/15min`, `/30min`, `/secret` slugs |
| NOVEL `book` lifecycle | ✓ slot-check + create flow works end-to-end |
| Pre-existing config-precedence issue | Found: `auth_header` in `config.toml` shadows `CAL_COM_TOKEN` env var. Worked around by emptying the auth_header line. Worth a separate generator fix — config Load should clear `AuthHeaderVal` when env-var token is set. |

## Lesson for regen-merge

The take-fresh-wholesale strategy assumed the 6 known-stale files (helpers, client, store, tools, schema_version_test, upsert_batch_test) had no in-body customizations. Pub's `client.go` proved that wrong for cal-com. Two improvements worth filing:

1. **Body-diff probe** before take-fresh: scan pub's stale-listed file for non-template function bodies; flag if any function body diffs significantly from fresh's.
2. **Per-CLI take-fresh blocklist** in `.printing-press.json`: let CLIs opt specific files out of the "always take fresh" rule when they have known in-body customizations the decl-set comparison can't see.

Cal.com's `calcom_versions.go` had a self-documenting `pp-note` comment flagging it as a generator gap; that file survived (TEMPLATED-WITH-ADDITIONS preservation worked). The integration call into `do()` was the one that vanished.

PR #303, opened 2026-05-08
fix(cli): fix slack sync for private channels and scope errors

neektza:
## Problem

Three bugs in `slack-pp-cli sync` prevented private channel messages from being pulled, even when the bot token had access to those channels.

### Bug 1: `mpim,im` types cause `missing_scope` → silent store corruption

`syncResourcePath` requested `types=public_channel,private_channel,mpim,im`. Bot tokens typically lack `mpim:read`, so Slack returned `{ok:false, error:"missing_scope"}`. This was stored as the single conversation record instead of surfacing as an error. Result: 1 corrupt record, 0 messages.

**Fix:** Removed `mpim,im` from the types param.

### Bug 2: No `ok:false` detection in `syncResource`

The generic sync loop had no Slack-specific error detection. Any `{ok:false}` response was treated as a valid record and stored, masking real failures.

**Fix:** Added `ok:false` check immediately after each API call.

### Bug 3: Wrong pagination cursor param (`after` → `cursor`)

`determinePaginationDefaults()` returned `cursorParam: "after"`, but Slack's `conversations.list` uses `cursor`. Pagination silently restarted from page 1 on every request.

**Fix:** Changed `cursorParam` to `"cursor"`.

## Validation

- Before: 1 conversation (corrupt error object), 0 messages
- After: **56 conversations (public + private), 5,956 messages, 7,106 total records**

## Files Changed

- `internal/cli/sync.go` — three targeted fixes, each marked with `// PATCH:` comment
- `.printing-press-patches.json` — patch manifest per AGENTS.md convention

---
*— HermesVonNikica*

### DIFF
--- a/library/productivity/slack/.printing-press-patches.json
+++ b/library/productivity/slack/.printing-press-patches.json
@@ -3,5 +3,25 @@
   "applied_at": "2026-05-08",
   "base_run_id": "20260409-073625",
   "base_printing_press_version": "1.2.1",
-  "patches": []
+  "patches": [
+    {
+      "id": "fix-conversations-types-scope",
+      "summary": "Remove mpim,im from conversations.list types to avoid missing_scope errors.",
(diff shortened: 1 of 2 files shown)


tmchow:
nice @neektza!

PR #358, opened 2026-05-09
fix(linear): linear-pp-cli auth set-token to persist api_key

giacaglia:
## Problem

`linear-pp-cli auth set-token` called `SaveTokens`, which stored the value in `access_token`. `AuthHeader()` only read `api_key` (and `LINEAR_API_KEY`), so `doctor` still reported auth as missing and GraphQL returned 401 after a successful "Token saved" message when users only had an OAuth access token on disk.

## Changes

- Add `SaveLinearAPIKey` to write the `api_key` TOML field and clear OAuth token fields; use `auth set-api-key` for Linear personal keys (`lin_api_…`).
- Keep `auth set-token` for OAuth access tokens; `AuthHeader()` now sends `Bearer <access_token>` when `api_key` is unset (per Linear OAuth), so status and GraphQL match stored OAuth tokens.
- Set `AuthSource` to `config:api_key` or `config:access_token` when the value comes from the file and env is unset.
- Clear `api_key` on `auth logout` (along with OAuth fields).
- Add `internal/config` tests (including api_key precedence over `access_token` when both are present).
- Document Cursor setup in new `library/project-management/linear/CURSOR.md`, correct the Linear UI link in README (Security & access), link from SKILL files, and drop the inaccurate claim that `CURSOR.md` is mirrored under `cli-skills/pp-linear/`.

## Verification

`go test ./...` in `library/project-management/linear`.


library/project-management/linear/internal/cli/auth.go
```diff
@@ -54,19 +58,42 @@ func newAuthStatusCmd(flags *rootFlags) *cobra.Command {
...
+}
+
 func newAuthSetTokenCmd(flags *rootFlags) *cobra.Command {
 	return &cobra.Command{
-		Use:     "set-token <token>",
-		Short:   "Save an API token to the config file",
-		Example: "  linear-pp-cli auth set-token sk_live_abc123",
+		Use:   "set-token <token>",
+		Short: "Save an OAuth access token to the config file (advanced)",
+		Long: "Stores the value in access_token for OAuth-style token flows. " +
```
tmchow, line 87: Required: `set-token` still saves into `access_token`, but the actual auth path only reads `api_key`/`auth_header` through `Config.AuthHeader()`. I reproduced this locally: after `linear-pp-cli --config <tmp>/config.toml auth set-token oauth_access_example`, `auth status` immediately reports `Not authenticated`, so API requests would still be sent without Authorization. Please either wire `access_token` into the real auth header/client behavior, or stop exposing `set-token` as an auth setup path.

library/project-management/linear/SKILL.md
```diff
@@ -38,6 +38,10 @@ go install github.com/mvanhorn/printing-press-library/library/project-management
 
 If `--version` reports "command not found" after install, the install step did not put the binary on `$PATH`. Do not proceed with skill commands until verification succeeds.
 
+## Cursor
+
+Personal API keys are created under **Account → Security & access** in Linear, not under Integrations. For step-by-step Cursor setup (keys, `auth set-api-key`, optional MCP tradeoffs), see [CURSOR.md](https://github.com/mvanhorn/printing-press-library/blob/main/library/project-management/linear/CURSOR.md) in the Linear module (mirrored verbatim under `cli-skills/pp-linear/`).
```
tmchow: Small docs fix while you are here: this says `CURSOR.md` is mirrored verbatim under `cli-skills/pp-linear/`, but this PR only mirrors `SKILL.md`; there is no `cli-skills/pp-linear/CURSOR.md`. The GitHub link is fine, so I would just drop the mirrored-verbatim clause.

tmchow (changes requested):
Thanks for tightening up the Linear API-key path. `auth set-api-key` itself looks good and the targeted verifier/tests pass, but I found one required behavior fix before merge: `auth set-token` is still advertised as usable while the client never authenticates from `access_token`.

tmchow:
conventional commit format please :)

library/project-management/linear/internal/cli/auth.go
giacaglia: You’re right! On the earlier revision SaveTokens / set-token only filled access_token, while AuthHeader() only considered auth_header and api_key, so your repro (set-token → auth status → “Not authenticated”) was accurate.

We went with your first option: wire access_token into the same auth path the client uses.

library/project-management/linear/SKILL.md
giacaglia: Good catch! That wording was wrong. I removed the “mirrored verbatim under cli-skills/pp-linear/” bit and kept the link to CURSOR.md in the Linear module only

tmchow (approved):

PR #394, opened 2026-05-10
fix(podscan): populate resourceIDFieldOverrides so sync stores rows

### LINKED ISSUE #383: podscan-pp-cli: `sync` for `categories` resource silently stores 0/163 records ("no extractable ID field found")
## Summary

`podscan-pp-cli sync` (v1.0.0) successfully fetches 163 records for the `categories` resource over the wire, but stores **0** in the local SQLite cache. The sync logs flag this as a non-critical anomaly (`reason: all_items_failed_id_extraction`) and exits 0 under the new default exit policy, so the failure is silent unless you read the JSON event stream.

## Reproduction

Fresh install of `podscan-pp-cli` v1.0.0 via the recommended path:

```bash
npx -y @mvanhorn/printing-press install podscan
podscan-pp-cli auth set-token <bearer>
podscan-pp-cli sync
```

Observed sync output (relevant lines):

```
{"event":"sync_start","resource":"categories"}
warning: 163/163 categories items skipped (no extractable ID field found)
{"event":"sync_anomaly","resource":"categories","consumed":163,"stored":0,"reason":"all_items_failed_id_extraction"}
{"event":"sync_progress","resource":"categories","fetched":0}
{"event":"sync_complete","resource":"categories","total":0,"duration_ms":564}
```

Direct API check (works fine):

```bash
curl -H "Authorization: Bearer $TOK" -H "User-Agent: curl/8.7.1" \
  https://podscan.fm/api/v1/categories
# Returns: {"categories": [{"category_id": "...", "category_name": "...", "category_display_name": "..."}, ...163 items...]}
```

## Root cause hypothesis

PodScan's `/categories` endpoint returns objects shaped:

```json
{ "category_id": "ct_xxx", "category_name": "...", "category_display_name": "..." }
```

The CLI's ID extractor appears to look for a field literally named `id`. Categories use `category_id` (consistent with `episode_id`, `podcast_id` elsewhere in the API — no top-level `id` field anywhere I've probed). The extraction template needs to either:

1. Accept resource-specific ID field names (`{resource}_id`), or
2. Be configurable via the spec annotation that drove codegen, or
3. Fall back to a deterministic field-name heuristic (`{singular_resource_name}_id`) when no plain `id` exists

## Impact

- **Local cache for `categories` is permanently empty** — every `categories list` call hits the live API even though `sync` was run
- **Other generated CLIs in the library may have the same template gap** if their resources use `{resource}_id` instead of `id` (worth grepping the codegen tree)
- **Failure is silent under the new exit policy** — `--strict` would surface it but most users will run sync without that flag and assume success because exit code is 0

## Workaround

None at the user level. Either skip `sync` for the `categories` resource (no flag for that visible in `--help`), or use `categories list --agent` (live, works fine) and manage the cache logic outside the CLI.

## Suggested next steps

- Confirm the ID-extraction template is the actual root cause (vs. a podscan-specific spec annotation issue)
- Audit other library CLIs for the same pattern: `grep -r "no extractable ID field found"` against the generated trees
- Consider promoting `all_items_failed_id_extraction` from "warning" to "critical" in the default exit policy when it affects the entire resource (not just some records)

## Environment

- `podscan-pp-cli --version`: 1.0.0
- Install path: `npx -y @mvanhorn/printing-press install podscan`
- OS: macOS (Darwin 25.3.0)
- Token tier: standard (not paid; for context, `exports` returns 403 and `alerts` returns 404 on the same token, but `categories` is reachable and returns full data)

gregvanhorn:
## Summary

Fixes #383. Thanks @jaredirish for the detailed repro.

`podscan-pp-cli sync` was consuming items over the wire but storing 0 rows for any resource without a plain `id` field. The generator's profiler shipped both `resourceIDFieldOverrides` maps empty (one in `internal/cli/sync.go`, one in `internal/store/store.go`), so `extractID` and `UpsertBatch` fell through to the generic fallback list — `id, ID, name, uuid, slug, key, code, uid` — none of which match PodScan's `{resource}_id` naming.

Visible failure: `categories` consumed 163, stored 0. The same shape was latent on `alerts`, `exports`, and `exports-podcasts`.

## Fix

Populate both override maps from the spec's response schemas:

| resource | id field | source |
|---|---|---|
| alerts | `alert_id` | `Alert.alert_id` |
| categories | `category_id` | `Category.category_id` |
| episodes | `episode_id` | `Episode.episode_id` |
| podcasts | `podcast_id` | `Podcast.podcast_id` |
| exports | `episode_id` | `/exports/episodes` returns Episode shape |
| exports-podcasts | `podcast_id` | `/exports/podcasts` returns Podcast shape |
| mentions | `episode_id` | `Mention` natural-keys on episode_id |

Both maps are kept in sync (UpsertBatch and extractID must resolve fields the same way).

## Why both files

`internal/cli/sync.go::extractID` and `internal/store/store.go::UpsertBatch` each maintain their own copy of the override map. A divergence between them produces silent drops on heterogeneous payloads — see the comment block at `store.go:629`.

## Upstream follow-up

The real fix belongs in `cli-printing-press`: the profiler should populate `resourceIDFieldOverrides` from response-schema `{resource}_id` fields when the spec lacks an `x-resource-id` annotation. Will file a separate issue against the generator. This PR is the library-side patch so users on v1.0.0 get a working sync today; recorded in `.printing-press-patches.json` so the next regen preserves the customization.

## Validation

- `go build ./...` ✓
- `go vet ./...` ✓
- `printing-press publish validate` ✓ (all 11 checks pass)

## Test plan

- [ ] `npx -y @mvanhorn/printing-press install podscan` (re-install)
- [ ] `podscan-pp-cli auth set-token <token>`
- [ ] `podscan-pp-cli sync` → categories should show `stored: 163` (not `stored: 0`)
- [ ] `podscan-pp-cli sync` → alerts/exports/exports-podcasts also store > 0 if account has them

### DIFF
--- a/library/media-and-entertainment/podscan/.printing-press-patches.json
+++ b/library/media-and-entertainment/podscan/.printing-press-patches.json
@@ -1,0 +1,19 @@
+{
+  "schema_version": 1,
+  "applied_at": "2026-05-10",
+  "base_run_id": "20260509-110700",
+  "base_printing_press_version": "4.1.0",
+  "patches": [
+    {
+      "id": "id-field-overrides",
+      "summary": "Populate resourceIDFieldOverrides for every podscan resource so sync stores rows instead of skipping them.",
+      "reason": "PodScan uses {resource}_id naming (alert_id, category_id, episode_id, podcast_id) and has no plain `id` field. The generator's profiler emitted both override maps empty, so extractID and UpsertBatch fell through to the generic fallback list (id, ID, name, uuid, slug, key, code, uid), found nothing, and dropped every item. Reported externally in #383: `sync` for `categories` consumed 163 items and stored 0.",
+      "files": [
+        "internal/cli/sync.go",
+        "internal/store/store.go"
+      ],
+      "validated_outcome": "Both override maps now match the spec's response schemas (Alert.alert_id, Category.category_id, Episode.episode_id, Podcast.podcast_id; exports/* keyed off the shape they return). go build and go vet pass.",
+      "upstream_issue": "cli-printing-press: profiler should populate resourceIDFieldOverrides from response-schema {resource}_id fields when the spec lacks x-resource-id annotations"
+    }
+  ]
+}
--- a/library/media-and-entertainment/podscan/internal/cli/sync.go
+++ b/library/media-and-entertainment/podscan/internal/cli/sync.go
@@ -1070,7 +1070,19 @@ func syncDependentResource(c interface {
 // Includes both flat resources and dependent (parent-child) resources so
 // annotations on a child path-item are honored at runtime, not just on
 // flat paths.
+//
+// PATCH: PodScan uses {resource}_id naming (no plain "id" field anywhere),
+// so without these overrides every flat resource hits the generic fallback
+// list, fails ID extraction, and stores 0 rows. The generator's profiler
+// shipped this map empty for podscan; populate it from the spec schemas.
 var resourceIDFieldOverrides = map[string]string{
+	"alerts":           "alert_id",
+	"categories":       "category_id",
+	"episodes":         "episode_id",
+	"podcasts":         "podcast_id",
+	"exports":          "episode_id",
+	"exports-podcasts": "podcast_id",
+	"mentions":         "episode_id",
 }
 
 // genericIDFieldFallbacks is the runtime safety net for resources that did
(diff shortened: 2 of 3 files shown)


gregvanhorn:
Filed upstream generator issue: mvanhorn/cli-printing-press#915 — long-term fix is having the profiler derive overrides from `{resource}_id` response-schema fields when `x-resource-id` annotations are absent.

tmchow:
Hey @gregvanhorn, this PR has unresolved review feedback and a low review score from over 6 hours back. Mind taking another pass? If you've hit a blocker or want to hand it off, just leave a note here — otherwise I'll close for inactivity in about 3 days.

tmchow:

PR #399, opened 2026-05-10
fix(table-reservation-goat): walk Chrome dirs directly in cookie discovery

pejmanjohn:
### Bug (user report)

`table-reservation-goat-pp-cli auth login --chrome` returned 0 or partial cookies on a machine where Chrome was signed in to both networks:

```json
{
  "notes": [
    "Tock cookie read failed: open /…/Profile 1/Network/Cookies: no such file or directory (and others)"
  ],
  "opentable_imported": 14,
  "opentable_logged_in": false,
  "tock_imported": 0,
  "tock_logged_in": false
}
```

Filesystem inspection showed:

```
~/Library/Application Support/Google/Chrome/Default/Cookies
~/Library/Application Support/Google/Chrome/Profile 1/Cookies
~/Library/Application Support/Google/Chrome/Local State
```

— old-layout `<profile>/Cookies`, no `Network/Cookies`. Manual workaround with symlinks `Default/Network/Cookies → ../Cookies` and `Profile 1/Network/Cookies → ../Cookies` recovered:

```json
{ "opentable_imported": 52, "opentable_logged_in": true, "tock_imported": 38, "tock_logged_in": true }
```

### Root cause

kooky v0.2.9's Chrome cookie discovery (`internal/chrome/find/find.go`) iterates only profiles listed in `<root>/Local State`'s `profile.info_cache` map. For each listed profile it tries both Chrome 96+ (`<profile>/Network/Cookies`) and pre-96 (`<profile>/Cookies`) layouts, so the layout fallback exists in theory. But a profile directory present on disk that's missing from `info_cache` is silently skipped — along with both layouts. The user's `info_cache` happened to omit at least one profile where the Tock cookies lived. Their symlinks happened to satisfy a `Network/Cookies` path kooky was already trying for some other (listed) profile.

### Fix

Supplement kooky's auto-discovery with a direct filesystem walk of every Chrome user-data root we know about. For each subdirectory that isn't a known non-profile sibling (`Crashpad`, `GrShaderCache`, `Subresource Filter`, etc.), try both cookie layouts via `kooky/browser/chrome.ReadCookies(path, ...)`. Merge with kooky's auto-discovery results, dedupe by `(Domain, Name, Path)`, prefer the entry with the later `Expires`.

Per-platform Chrome roots covered:
- **darwin**: `Google/Chrome`, `Google/Chrome Beta`, `Google/Chrome Canary`, `Google/Chrome Dev`, `Chromium`, `BraveSoftware/Brave-Browser`, `Arc/User Data`
- **linux**: `google-chrome`, `google-chrome-beta`, `google-chrome-unstable`, `chromium`, `BraveSoftware/Brave-Browser`
- **windows**: kooky-only (no reliable cross-version path map here)

Same supplementary walk applied to `RefreshAkamaiCookies` (the per-call freshness re-read), so cross-network commands benefit too.

### Tests

- `TestChromeCookieCandidatePathsIn` — temp-dir Chrome-like layout with mixed Default-only-old, Profile-1-only-new, Profile-2-both-layouts, plus `Crashpad`/`GrShaderCache` siblings that must be excluded
- `TestChromeCookieCandidatePathsIn_MissingRoot` — non-existent roots silently tolerated
- `TestDedupeCookies_KeysOnDomainNamePath` — same `(Domain, Name, Path)` collapsed; later-`Expires` wins
- `TestDedupeCookies_NilEntriesIgnored` — kooky's occasional nil entries don't panic

### Validation

| Check | Result |
|---|---|
| `go build ./...` | PASS |
| `go test ./...` | PASS (4 new tests, all packages green) |

### Out of scope

- Updating `.printing-press-patches.json`'s `validated_outcome` ledger — deferred to avoid a recurring conflict with PR #387 (still open). The patches manifest can be brought up to date in a follow-up commit once the open PRs settle.

### DIFF
--- a/library/food-and-dining/table-reservation-goat/internal/source/auth/chrome.go
+++ b/library/food-and-dining/table-reservation-goat/internal/source/auth/chrome.go
@@ -9,12 +9,196 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"time"
 
 	"github.com/browserutils/kooky"
 	_ "github.com/browserutils/kooky/browser/chrome"
+	chromebrowser "github.com/browserutils/kooky/browser/chrome"
 )
+
+// chromeRoots returns the per-platform candidate Chrome (and Chrome-derivative)
+// user-data directories. Used by the supplementary filesystem walk below to
+// supplement kooky's info_cache-driven discovery — kooky only iterates profiles
+// listed in <root>/Local State's profile.info_cache, so a profile dir present
+// on disk but missing from info_cache (corrupted, stale, fresh-install edge
+// cases) gets silently skipped along with its <profile>/Cookies file.
+func chromeRoots() []string {
+	cfgDir, err := os.UserConfigDir()
+	if err != nil {
+		return nil
+	}
+	switch runtime.GOOS {
+	case "darwin":
+		return []string{
+			filepath.Join(cfgDir, "Google", "Chrome"),
+			filepath.Join(cfgDir, "Google", "Chrome Beta"),
+			filepath.Join(cfgDir, "Google", "Chrome Canary"),
+			filepath.Join(cfgDir, "Google", "Chrome Dev"),
+			filepath.Join(cfgDir, "Chromium"),
+			filepath.Join(cfgDir, "BraveSoftware", "Brave-Browser"),
+			filepath.Join(cfgDir, "Arc", "User Data"),
+		}
+	case "linux":
+		return []string{
+			filepath.Join(cfgDir, "google-chrome"),
+			filepath.Join(cfgDir, "google-chrome-beta"),
+			filepath.Join(cfgDir, "google-chrome-unstable"),
+			filepath.Join(cfgDir, "chromium"),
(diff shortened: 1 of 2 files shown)


pejmanjohn:
Addressed both Greptile findings in `128cc8d`:

**Finding 1 (`dedupeCookies` Expires logic)** — fixed per Greptile's suggestion. The guard now keeps the existing entry when it has a concrete `Expires` AND the challenger is session-only (zero `Expires`) OR the existing's expiry is later. Added `TestDedupeCookies_PersistentBeatsSession` covering both encounter orders (kooky-first/walk-second and the reverse).

**Finding 2 (`nonProfileDirs` denylist staleness)** — replaced denylist with an allowlist regex `^(Default|Profile \d+|Guest Profile)$`. Future Chrome additions (SegmentationPlatform, MEIPreload, Snapshots, etc.) are rejected by structure rather than enumeration. Test grew to include SegmentationPlatform and System Profile siblings to verify the allowlist rejects unknowns it has never seen by name.

`go test ./internal/source/auth/` passes; full suite still green.

PR #401, opened 2026-05-10
fix(table-reservation-goat): doctor probe — use source clients, not dead http path

pejmanjohn:
### Bug (user report)

```
table-reservation-goat-pp-cli doctor reports:
FAIL API: unreachable: GET /: Get "https://www.opentable.com/": stream error: stream ID 7; INTERNAL_ERROR; received from peer
base_url: https://www.opentable.com
version: 1.0.0
```

`curl` confirms the upstream behavior: `403 Akamai access denied` over HTTP/1.1, `HTTP/2 stream not closed cleanly: INTERNAL_ERROR` over HTTP/2. Akamai's WAF rejects naive clients on the bare root.

### Root cause

Same family as PR #389 — `doctor` was probing via the vestigial `internal/client.Client` against `BaseURL = "https://www.opentable.com"` with `c.Get("/", nil)`. No real CLI command uses that path; the working code routes through `internal/source/{opentable,tock}/` with Surf's Chrome TLS fingerprint, GraphQL persisted-queries, SSR scraping, kooky-imported cookies, and Akamai-aware cooldown handling. `doctor`'s probe was lying about what the CLI actually does.

The HTTP/2 `INTERNAL_ERROR` doesn't fit the `*client.APIError` shape (no response body), so it falls through to the bare transport-error path → `unreachable: …` even though Akamai's edge clearly responded.

### Fix

Replace `c.Get("/", nil)` with `opentable.New(session).Bootstrap(ctx)` — the same Bootstrap call `goat`, `earliest`, `watch`, etc. use. It:
- Navigates `/restaurant/profile/100` (Akamai-whitelisted; the cooldown patch's documented escape hatch)
- Reads kooky-imported session cookies via Surf's Chrome fingerprint
- Short-circuits with `*BotDetectionError` when a prior-403 cooldown is active
- Caches CSRF token so warm probes are essentially free

Distinct rendering for the three failure modes:
- transport reached, Akamai blocked → `OpenTable: blocked by Akamai (cooldown until 2026-05-10T08:30Z) — try \`auth login --chrome\` to refresh cookies`
- transport failed → `OpenTable: unreachable: <err>`
- success → `OpenTable: reachable`

New `Auth:` line replaces the previous `Auth: not required` placeholder with actual cookie counts:

```
Config:    ok
Auth:      opentable: 24 cookies, tock: 38 cookies
OpenTable: reachable
```

…or, if not signed in:

```
Auth:      no cookies imported (run `auth login --chrome` after signing in to opentable.com and exploretock.com in Chrome)
OpenTable: skipped (no session loaded)
```

### Tock probe

Skipped in this PR — Tock has no equivalent cheap stable probe; each venue is its own SSR page. Cookie-jar count is the Tock readiness signal. Future PR can add a Tock probe (city-search SSR with a known stable metro) if useful.

### Removed

- `credentials` check — was guarded on `cfg.AuthHeader() != ""`, which is always empty for this cookie-auth CLI. Code path was dead user-visible.
- `internal/client` import in `doctor.go` — `data_source.go` still references it via the read-cache pipeline, so the package itself stays.

### Validation

| Check | Result |
|---|---|
| `go build ./...` | PASS |
| `go test ./...` | PASS |
| Manual `doctor --help` smoke | PASS |

### Out of scope

- `.printing-press-patches.json` ledger update — deferred (same reason as #399, avoiding manifest conflict with #387).
- Full removal of `internal/client/client.go` — still referenced by `data_source.go` via typed-endpoint subcommands' read-cache pipeline. Bigger refactor; separate PR.

### Related

- #389 (merged) — removed the broken typed-tool MCP layer that hit the same dead client
- #399 — companion fix for `auth login --chrome` cookie discovery on machines where Chrome's `Local State` is missing profile entries

### DIFF
--- a/library/food-and-dining/table-reservation-goat/internal/cli/doctor.go
+++ b/library/food-and-dining/table-reservation-goat/internal/cli/doctor.go
@@ -6,17 +6,17 @@ package cli
 import (
 	"context"
 	"database/sql"
-	"errors"
 	"fmt"
 	"io"
 	"os"
 	"strings"
 	"time"
 
-	"github.com/mvanhorn/printing-press-library/library/food-and-dining/table-reservation-goat/internal/client"
-	"github.com/mvanhorn/printing-press-library/library/food-and-dining/table-reservation-goat/internal/config"
-	"github.com/mvanhorn/printing-press-library/library/food-and-dining/table-reservation-goat/internal/store"
 	"github.com/spf13/cobra"
+	"github.com/mvanhorn/printing-press-library/library/food-and-dining/table-reservation-goat/internal/config"
+	"github.com/mvanhorn/printing-press-library/library/food-and-dining/table-reservation-goat/internal/source/auth"
+	"github.com/mvanhorn/printing-press-library/library/food-and-dining/table-reservation-goat/internal/source/opentable"
+	"github.com/mvanhorn/printing-press-library/library/food-and-dining/table-reservation-goat/internal/store"
 )
 
 // looksLikeDoctorInterstitial reports whether the response body matches a known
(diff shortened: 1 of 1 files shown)


pejmanjohn:
Addressed both Greptile findings in `a1d94e6`:

**Finding 1 (`"skipped (no session loaded)"` is dead code)** — confirmed via `auth/auth.go` that `auth.Load()` returns a non-nil empty `*Session` on missing file, never `(nil, nil)`. Changed the gate from `session != nil` to `otCookies > 0` (we already counted cookies for the `auth:` line, so it's free). Now when no OpenTable cookies are imported, doctor reports the honest "skipped (no opentable cookies — run `auth login --chrome`)" instead of triggering Bootstrap, hitting Akamai cold, and getting a misleading "blocked by Akamai" report that hides the real cause.

**Finding 2 (duplicate `auth login --chrome` hint)** — removed the redundant suffix. `bde.Error()` already names the recovery action.

`go build ./... && go test ./...` both pass.

pejmanjohn:
Addressed in `c3c93e1` per Greptile's suggestion. `"client init: %s"` → `"error: client init: %s"` so both the indicator switch and `doctorExitForFailOn` classify a client-init failure as FAIL instead of falling through to green OK on Go errors that don't happen to contain the word "error" (e.g. `"no cookie jar"`, `"connection refused"`).

pejmanjohn:
Addressed in `440d78b` per Greptile's suggestion. `"no cookies imported (...)"` → `"not configured: no cookies imported (...)"` so both the indicator switch and `doctorExitForFailOn` classify it as FAIL. The previous string contained "no" but not "not " (with trailing space) and rendered as a contradictory green OK alongside the recovery prompt.

pejmanjohn:
Caught both findings I'd missed in earlier fixups, addressed in `ad31ece`:

**P1 — `"blocked by Akamai"` was rendering as green OK** — added `strings.Contains(s, "blocked")` to both the indicator switch and `doctorExitForFailOn` predicates, matching the existing `"unreachable"` pattern (visual FAIL + exit-code FAIL). A bot-blocked state now renders red and trips `--fail-on=error` so CI doesn't silently pass.

**P2 — `looksLikeDoctorInterstitial` is dead code** — removed. Was wired to the old `client.Get("/", nil)` reachability probe that `a39b21b` replaced with `opentable.Bootstrap`; nothing else parses HTML response bodies for vendor challenge markers anymore.

`go build ./... && go test ./...` both pass. Confidence-rating regression should clear on the next review pass.

pejmanjohn:
Addressed in `85ddd0c` per Greptile's exact suggestion. Added `"not configured"` to `doctorExitForFailOn`'s predicate so it matches the indicator switch — the no-cookies state now correctly trips both the visual FAIL and `--fail-on=error` exit code.

Same class of bug as the previous "blocked" finding (indicator and exit predicates drifted apart). Auditing both lists for other gaps revealed none — they now share the full set: `error`, `unreachable`, `invalid`, `missing`, `blocked`, `not configured`.

pejmanjohn:
@greptileai please re-review — the confidence score on the landing page is anchored to a summary from 5 fixup commits ago. Current head is `85ddd0c`. All inline findings from prior reviews have been addressed:

- `client init: …` → prefixed with `error:` so FAIL fires (`c3c93e1`)
- `no cookies imported …` → prefixed with `not configured:` (`440d78b`)
- `blocked by Akamai …` → "blocked" added to indicator + exit predicates (`ad31ece`)
- `looksLikeDoctorInterstitial` dead code → removed (`ad31ece`)
- `doctorExitForFailOn` missing `not configured` → added (`85ddd0c`)

Indicator switch and `doctorExitForFailOn` now share the full FAIL set: `error`, `unreachable`, `invalid`, `missing`, `blocked`, `not configured`.

Issue #406, opened 2026-05-10
table-reservation-goat-pp-cli: slug resolver, ID resolution, and Tock coverage failures (real-world Mother's Day query)


mvanhorn:
## TL;DR

Three independent failures hit on a real-world `/pp-table-reservation-goat` query. Net effect: the agent had to abandon the CLI mid-task and finish the job by scraping OpenTable through Chrome MCP.

1. **Slug resolver matches globally, not by geography.** `joey-bellevue` resolved to "Joey's Bold Flavors" (different state). Multiple slugs misresolve. The CLI returns availability for the wrong venue with no warning.
2. **Numeric IDs from `restaurants list` aren't accepted by `availability check`.** No documented bridge from list output to availability input.
3. **Tock has zero Bellevue, WA coverage in the index.** The cross-network value prop silently degrades to OpenTable-only across the entire Seattle Eastside.

Secondary: `earliest` returns `{}` on valid slugs, Akamai 403s persist after `auth login --chrome`, and the auth flow logs a non-fatal "Network/Cookies: no such file" warning.

`/cc @pejmanjohn`

## Original prompt

Invoked via `/pp-table-reservation-goat` on 2026-05-10 (Mother's Day):

> can i get reservation for mothers day for 6 people between now and 5 pm in bellevue

## Environment

- `table-reservation-goat-pp-cli 1.0.0`
- macOS (Darwin 25.3.0)
- Fresh `auth login --chrome`: imported 25 OpenTable cookies + 18 Tock cookies. Both networks reported `logged_in: true`.
- User logged into Tock in Chrome immediately before the auth import.

## Failure 1 — slug resolver matches globally instead of by geography

`availability check opentable:<slug>` resolves slugs against a global index. When a Bellevue-suffixed slug doesn't exist as that exact string in the source-of-truth, the resolver picks the **closest fuzzy match anywhere in the world** and returns its availability with no warning. This is "confidently wrong" — the worst failure mode for an agent CLI:

| Input slug | Resolved venue | Where it actually is | Expected venue |
|---|---|---|---|
| `joey-bellevue` | Joey's Bold Flavors (id=1434169) | not Bellevue, WA | JOEY Bellevue (Seattle Eastside) |
| `13-coins-bellevue` | Via 13 | not Bellevue, WA | 13 Coins - Bellevue (id=1227073) |
| `carmines-bellevue` | Carmines Ybor City | Tampa, FL | Carmine's - Bellevue (id=187612) |
[...]

PR #423, opened 2026-05-10
fix(table-reservation-goat): accept numeric IDs in availability resolver

pejmanjohn:
## table-reservation-goat — accept numeric IDs in availability resolver

Fixes failure 2 from issue #406. First of four planned PRs to address the issue.

### The bug

`restaurants list` emits clean rows with numeric OpenTable IDs:

```json
{ "id": "3688", "name": "Daniel's Broiler - Bellevue", "url": "https://www.opentable.com/restaurant/profile/3688", ... }
```

But `availability check`, `availability multi-day`, and `earliest` all rejected those IDs:

```bash
$ availability check 3688 --party 6 --date 2026-05-10 --agent
{"available": false, "reason": "opentable: could not resolve \"3688\" ...", "venue": "3688"}

$ availability check opentable:3688 --party 6 --date 2026-05-10 --agent
{"available": false, "reason": "opentable: could not resolve \"3688\" ..."}
```

The Autocomplete-based slug resolver does name-similarity search, not ID lookup. Numeric strings don't match any venue name, so the resolver bails. **Agents that tried the natural `list → check` composition got a dead end** — the user in #406 fell back to scraping OpenTable through Chrome MCP instead.

### Fix

In `resolveEarliestForVenue` (the shared helper that powers all three commands), detect `^\d+$` in the slug arg and pass it directly to `RestaurantsAvailability` as the restaurantId. The Autocomplete slug-resolver is skipped entirely on the numeric path — which has three downstream benefits:

1. **Composition restored** — `restaurants list` IDs work as inputs to `availability check` / `multi-day` / `earliest`. Agents get the natural workflow.
2. **Bypasses the global-fuzzy-match bug (#406 failure 1)** — the numeric path doesn't touch the resolver at all, so the well-known "joey-bellevue → Joey's Bold Flavors" misfire can't trigger.
3. **Bypasses Akamai's `Autocomplete` WAF rule** — agents hitting op-specific 403s on slug resolution can pivot to numeric IDs as a recovery strategy.

### Tock side

Numeric IDs aren't a Tock convention (venues use domain-name slugs like `canlis`, `farzi-cafe-bellevue`). `tock:<digits>` now returns a typed category error pointing the caller at the correct shape, instead of running a doomed Calendar fetch:

```
{
  "available": false,
  "network": "tock",
  "reason": "tock: \"3688\" looks like a numeric ID, but Tock venues are addressed by domain-name slug (e.g. 'canlis', 'farzi-cafe-bellevue'). Numeric IDs are an OpenTable-only convention; try 'opentable:3688' instead."
}
```

### Verify echo

`row.URL = /restaurant/profile/<id>` always resolves canonically from the numeric ID. Agents can verify the venue from the URL even when `restName` is empty (no Autocomplete hydration). Reason strings fall back to `restaurant #<id>` so the output reads naturally.

### Tests

`internal/cli/earliest_test.go` (new): 4 cases.

- `TestResolveEarliestForVenue_TockNumericRejected`:
  - `tock:3688` → typed rejection
  - `tock:1183597` → typed rejection
  - `tock:42` (small numeric) → typed rejection
- `TestResolveEarliestForVenue_BareNumericIsAmbiguous`:
  - bare `3688` does NOT trip the Tock rejection (only explicit `tock:` prefix does)

### Help text

Updated `availability check` long-help with all three input shapes documented and an example showing the `list → numeric ID → check` composition.

### Validation

| Check | Result |
|---|---|
| `go build ./...` | PASS |
| `go test ./...` | PASS (4 new tests + full suite) |

### Out of scope (deferred to follow-up PRs)

- **Failure 1** (slug resolver geo-bias) — PR C will fix this with metro-aware ranking + dynamic Tock metroArea hydration. **Note**: this PR partially mitigates failure 1 by giving agents a numeric-ID escape hatch; the slug resolver itself still has the global-fuzzy bug for any caller that doesn't have the numeric ID handy.
- **Failure 3** (zero Tock Bellevue coverage) — PR C, same fix as the slug resolver geo bias.
- **Failure 4** (`earliest` returns `{}`) — PR B, dedicated structured envelope.
- **Failure 5** (op-specific Akamai 403) — PR D, typed signal + agent recovery docs.

### Related

- Issue: #406
- Companion PRs (in flight): #387 (Greptile follow-ups for v0.2.1), #399 (Chrome cookie walk fallback), #401 (doctor probe via source clients)

### DIFF
--- a/library/food-and-dining/table-reservation-goat/internal/cli/availability_check.go
+++ b/library/food-and-dining/table-reservation-goat/internal/cli/availability_check.go
@@ -26,8 +26,17 @@ func newAvailabilityCheckCmd(flags *rootFlags) *cobra.Command {
 		Use:   "check <restaurant>",
 		Short: "Check open slots for a restaurant on a specific date and party size",
 		Long: "Per-venue availability across both networks. Resolves the venue on OpenTable " +
-			"or Tock and returns the earliest matching slot per the requested date/party.",
-		Example:     "  table-reservation-goat-pp-cli availability check 'tock:alinea' --party 2 --date 2026-06-15 --json",
+			"or Tock and returns the earliest matching slot per the requested date/party.\n\n" +
+			"Restaurant identifier accepts three shapes:\n" +
+			"  • Bare slug — 'canlis' (searches both networks)\n" +
+			"  • Network-prefixed slug — 'opentable:le-bernardin', 'tock:alinea'\n" +
+			"  • Numeric OpenTable ID — '3688' or 'opentable:3688'. IDs come from\n" +
+			"    `restaurants list --json` (the `id` field) and bypass the\n" +
+			"    name-based slug resolver entirely, so they're the most\n" +
+			"    reliable input shape for agents composing `list → check`.\n" +
+			"    Tock has no numeric-ID convention; use the domain-name slug.",
+		Example: "  table-reservation-goat-pp-cli availability check 'tock:alinea' --party 2 --date 2026-06-15 --json\n" +
+			"  table-reservation-goat-pp-cli availability check 3688 --party 6 --date 2026-12-25 --agent",
 		Annotations: map[string]string{"mcp:read-only": "true"},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) == 0 {
--- a/library/food-and-dining/table-reservation-goat/internal/cli/earliest.go
+++ b/library/food-and-dining/table-reservation-goat/internal/cli/earliest.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"os"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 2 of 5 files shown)


pejmanjohn:
Addressed in `7be651f`: renamed the misleading "trailing whitespace tolerated" sub-case to "small numeric rejected" and added a comment explaining what the case actually pins (digit-shape predicate, not whitespace handling).

pejmanjohn:
@greptileai please re-review — head is now `7be651f`. The "trailing whitespace tolerated" rename addresses your only inline finding.

pejmanjohn:
@greptileai please re-review — head is now `ac4249f`.

Addressed the carry-over P1 you flagged on the prior review about `restSlug=""` getting passed to ChromeAvailability on the numeric-ID path. Two changes:

1. Documented at the call site that the empty-slug pass-through is intentional — ChromeAvailability already handles it via `/restaurant/profile/<id>` fallback URL (Akamai treats both routes as legitimate user navigation).

2. Extracted `chromeAvailPageURL` helper so the fallback URL shape is testable. New `chrome_avail_url_test.go` pins three cases including the empty-slug case, so any future refactor that produces a broken `/r/?covers=...` URL will fail tests immediately.

PR #424, opened 2026-05-10
fix(table-reservation-goat): earliest — structured envelope vs `{}`

pejmanjohn:
## table-reservation-goat — `earliest` structured envelope vs `{}`

Fixes failure 4 from issue #406. Second of four planned PRs.

### The bug

```bash
$ earliest 'daniels-broiler-bellevue,joey-bellevue,fogo-de-chao-bellevue,...' \
    --party 6 --within 1d --agent --select earliest.venue,earliest.network,earliest.slot_at
{}
```

`{}` with no envelope, no error. Indistinguishable from "checked, no slots." The reporter couldn't tell whether their 8 slugs failed to resolve, the API was down, or there were genuinely zero slots.

### Two compounding causes

1. **Wrong `--select` path** — the response had `results[]`, not `earliest`. `--select earliest.X` matched nothing and the filter stripped to `{}`. Help text didn't document the response shape.
2. **No structural distinction between "zero resolved" and "resolved but no slots"** — even with the right selector, both rendered as `results: []` (with a `--select` that didn't pull meta out, indistinguishable).

### Fix

Added a `meta` envelope at the top level of `earliestResponse` that always carries summary counts, plus an `unresolved[]` companion to `results[]`:

```json
{
  "venues":     ["daniels-broiler-bellevue", "joey-bellevue", ...],
  "party":      6,
  "within_days": 1,
  "meta": {
    "venues_requested": 8,
    "resolved":         3,
    "unresolved":       5,
    "available":        1
  },
  "results":    [{"venue": "daniels-broiler-bellevue", ...}, ...],
  "unresolved": [{"venue": "joey-bellevue", "reason": "..."}, ...],
  "queried_at": "..."
}
```

A row counts as **unresolved** when its `Network` is empty or `"unknown"` — the slug→ID resolver short-circuited before any source call. Resolved-but-blocked rows (Network set, Available=false, Reason mentions Akamai or no-slots) stay in `results` and count toward `resolved` but not `available`.

Agents can now branch on `meta.resolved == 0` to detect "nothing resolved, try different inputs" distinctly from `meta.available == 0` meaning "checked, no slots."

### Help text

`earliest --help` now enumerates the actual response shape and lists common `--select` paths:

> **Common `--select` paths:** `results.venue`, `results.network`, `results.slot_at`, `results.bookable_times`, `meta.resolved`, `meta.available`, `unresolved.venue`, `unresolved.reason`.

So the next agent that reaches for `--select earliest.X` will check the help and see the right path.

### Tests

`internal/cli/earliest_test.go` grows by 2 new test functions:

- `TestSummarizeEarliest` — 5 sub-cases covering:
  - All resolved, none available
  - All resolved, all available
  - All unresolved
  - Mixed (some resolve, some don't, some available)
  - Empty input
- `TestEarliestResponse_JSONShapeContractsMeta` — pins the JSON-shape invariants the meta envelope is supposed to guarantee (meta always present, venues_requested/resolved/unresolved/available all serialized, results+unresolved arrays present).

### Smoke verify

```bash
$ ./table-reservation-goat-pp-cli earliest 'foo,bar' --party 2 --within 1d --dry-run --agent \
    --select 'results.venue,results.network,meta.resolved,meta.unresolved'
{
  "meta": {"resolved": 2, "unresolved": 0},
  "results": [{"network": "opentable", "venue": "foo"}, {"network": "opentable", "venue": "bar"}]
}
```

The `--select earliest.X` case still produces `{}` (selector engine has no match for `earliest`), but agents reading the help will use the right path.

### Validation

| Check | Result |
|---|---|
| `go build ./...` | PASS |
| `go test ./...` | PASS (2 new test functions, 6 new sub-cases; full suite green) |
| Live JSON-shape smoke (dry-run) | PASS |

### PR stacking

This PR is **stacked on top of #423** (numeric-ID acceptance). `internal/cli/earliest_test.go` includes both PR A's `TockNumericRejected` / `BareNumericIsAmbiguous` tests and PR B's `SummarizeEarliest` / `JSONShape` tests because the test file as a unit moves together. If #423 lands first, this PR rebases cleanly to a 2-file diff.

If you'd prefer to review them in either order, the changes are conceptually independent — both touch `earliest.go` but in disjoint regions (PR A in the resolver, PR B in the response shape).

### Related

- Issue: #406
- PR A (stacked under this): #423 — numeric IDs in availability resolver
- PR C (planned, stacks above this): dynamic Tock metroArea + slug-resolver geo bias
- PR D (planned): op-specific WAF typed signal + agent recovery docs

### DIFF
--- a/library/food-and-dining/table-reservation-goat/internal/cli/earliest.go
+++ b/library/food-and-dining/table-reservation-goat/internal/cli/earliest.go
@@ -376,7 +376,15 @@ func resolveEarliestForVenue(ctx context.Context, s *auth.Session, venue string,
 	if tryOT {
 		c, err := opentable.New(s)
 		if err == nil {
-			row.Network = "opentable"
+			// NOTE: `row.Network = "opentable"` is deliberately NOT set
+			// here. PR #424 round-3 Greptile finding: setting Network
+			// before slug resolution caused slug-resolve failures to be
+			// miscounted as `meta.resolved` (Network was already
+			// "opentable" when summarizeEarliest's partition ran). The
+			// assignment moves AFTER we have a confirmed valid restID
+			// so the partition correctly categorizes failures as
+			// `meta.unresolved`.
+
 			// Numeric-ID short-circuit (issue #406, failure 2): `restaurants
 			// list` emits numeric OpenTable IDs (e.g. id=3688 for "Daniel's
 			// Broiler - Bellevue") but the Autocomplete-based slug resolver
@@ -410,11 +418,19 @@ func resolveEarliestForVenue(ctx context.Context, s *auth.Session, venue string,
 				var rerr error
 				restID, restName, restSlug, rerr = c.RestaurantIDFromQuery(ctx, slug, 40.7128, -74.0060)
 				if rerr != nil {
+					// Slug-resolve failed. row.Network stays empty so
+					// summarizeEarliest partitions this row into
+					// `unresolved[]` (PR #424 round-3 fix).
 					row.Available = false
 					row.Reason = fmt.Sprintf("opentable: could not resolve %q (%v)", slug, rerr)
 					return row
 				}
 			}
+			// Slug resolution succeeded (numeric path or named path).
+			// Claim the row for OpenTable so downstream partitioning
+			// counts this venue as resolved, even if the subsequent
+			// availability fetch is blocked by Akamai.
+			row.Network = "opentable"
 			row.URL = fmt.Sprintf("%s/restaurant/profile/%d", opentable.Origin, restID)
 			// New OT gateway (May 2026) returns single-day availability per
 			// call (forwardDays=0); scan multi-day windows by looping the
--- a/library/food-and-dining/table-reservation-goat/internal/cli/earliest.go
+++ b/library/food-and-dining/table-reservation-goat/internal/cli/earliest.go
@@ -48,12 +48,64 @@ type earliestRow struct {
 	Source   string `json:"source,omitempty"`
 }
 
+type earliestMeta struct {
+	// VenuesRequested is the count of input slugs (including duplicates).
+	VenuesRequested int `json:"venues_requested"`
+	// Resolved is the count of slugs that the source-client successfully
+	// mapped to a real network (OpenTable or Tock). Resolved-but-blocked
+	// counts here too — they got past slug→ID resolution, just couldn't
+	// fetch slots.
+	Resolved int `json:"resolved"`
+	// Unresolved is VenuesRequested - Resolved. Issue #406 failure 4:
+	// without this field, a request that resolved zero slugs returned
+	// `{}` (under `--select results.X` with no rows), indistinguishable
+	// from "resolved fine, just no slots open." Surfacing the count
+	// always lets agents branch on the case.
+	Unresolved int `json:"unresolved"`
+	// Available is the count of resolved venues with at least one bookable
+	// slot in the window.
+	Available int `json:"available"`
+}
+
+type unresolvedRow struct {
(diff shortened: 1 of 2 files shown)


pejmanjohn:
Addressed both findings in `0c130bc`:

1. Dropped `omitempty` from `Unresolved` so the field always emits (as `[]` when empty, matching `Results`). Initialized the slice as `[]unresolvedRow{}` in `summarizeEarliest` so Go marshals it as `[]` not `null`. New test `TestEarliestResponse_UnresolvedEmittedWhenEmpty` pins both the nil-slice and explicit-empty-slice cases.
2. Doc-commented `summarizeEarliest`'s PRECONDITION that callers pass `rows` and `venues` of the same length with positional correspondence — `Resolved + Unresolved == VenuesRequested` only holds under that condition.

pejmanjohn:
@greptileai please re-review — head is now `0c130bc`. Both inline findings (Unresolved omitempty + summarizeEarliest precondition) are addressed; new test pins the JSON-shape symmetry.

pejmanjohn:
Addressed in `ef7d1c0`. Good catch — the `Results: rows` assignment was indeed duplicating unresolved venues into both arrays. Refactored `summarizeEarliest` to return `(meta, results, unresolved)` so the partition happens in one place and the caller passes `Results: results` (not `Results: rows`). Applied to both dry-run and live paths.

`TestSummarizeEarliest` now asserts the three partition invariants:
- `len(results) == meta.Resolved`
- no row in `results[]` has `Network ∈ {"", "unknown"}`
- no venue appears in BOTH `results[]` and `unresolved[]`

Live smoke (`earliest 'foo,bar' --dry-run --agent`) shows results_len=2, unresolved_len=0, no duplication.

pejmanjohn:
@greptileai please re-review — head is now `ef7d1c0`. The round-2 finding about `Results: rows` duplicating unresolved venues has been addressed: `summarizeEarliest` now returns `(meta, results, unresolved)` so the partition happens in one place and the caller assigns `Results: results` (not `Results: rows`). The test asserts the partition is disjoint.

pejmanjohn:
@greptileai please re-review — head is now `9907a06`.

Addressed the carried-over P1 you flagged in the summary: `row.Network = "opentable"` was being set BEFORE the slug resolver ran, so slug-resolve failures kept Network="opentable" and got miscounted as `meta.resolved` instead of `meta.unresolved`.

Fix: moved the `row.Network = "opentable"` assignment to AFTER successful slug resolution (numeric-ID path always succeeds; named-path returns early on failure with Network unset). The Tock branch was already correct — it only sets Network="tock" inside the successful Calendar-fetch block.

`TestSummarizeEarliest`'s partition-invariant assertions (added in `ef7d1c0`) now act as the regression guard: any future code that sets Network before resolution succeeds will produce rows where unresolved venues leak into results[], and the tests will catch it.

pejmanjohn:
@greptileai please re-review — head is now `7cd3420`.

Addressed both round-2 findings:

1. **Carry-over P1 (`row.Network = "opentable"` set before slug resolution)** — fixed in `9907a06`: moved the Network assignment to AFTER successful slug resolution. Slug-resolve failures now correctly partition into `meta.unresolved` instead of being miscounted as `meta.resolved`.

2. **P2 (misleading test comment + weak `null || []` assertion)** — fixed in `7cd3420`: restructured `TestEarliestResponse_UnresolvedEmittedWhenEmpty` to assert each contract separately:
   - Baseline: nil slice marshals to `null` (Go semantics, documented as what we DON'T want)
   - Contract: empty slice marshals to `[]` (what we enforce)
   - End-to-end: `summarizeEarliest` returns non-nil even when no venues are unresolved

PR #398, opened 2026-05-10
feat(google-search-console): add google-search-console

heyimjames:
Re-submission of [#283](https://github.com/mvanhorn/printing-press-library/pull/283) addressing all review feedback. The previous PR was closed before the final fixes landed; opening this one against the same branch.

## Changes since #283

1. **`fix(google-search-console): inline novel-command flag declarations`** (934cedbb) — addressed the 16 verify-skill `flag-commands` errors flagged in #283 by inlining `--site`/`--window`/`--db` declarations on each of the 11 transcendence + 3 absorb commands. Filed [mvanhorn/cli-printing-press#718](https://github.com/mvanhorn/cli-printing-press/issues/718) on the upstream drift between this repo's verify-skill script and the printing-press binary's bundled verifier; the resync to that script is now in.
2. **`fix(google-search-console): restore public-library import path for store`** (6c9b91e5) — the inline-flags fix had inadvertently reverted the publish-package import-path rewrite for `internal/store`. Restored.
3. **`fix(google-search-console): improve sync_correctness scorer signals`** (7bb45c03) — addresses the scorer-token-presence finding ([mvanhorn/cli-printing-press#717](https://github.com/mvanhorn/cli-printing-press/issues/717)).
4. **`feat(google-search-console): add AGENTS.md per #728 convention`** (99a7f113) — adds the generated-CLI AGENTS.md the rest of the library now ships, rendered from the template introduced in [mvanhorn/cli-printing-press#728](https://github.com/mvanhorn/cli-printing-press/pull/728).

## Original description

**Every Search Console feature, plus a local SQLite store that turns one-shot API calls into time-series insights no other GSC tool ships.**

Sync your search analytics, sites, sitemaps, and URL inspection state into a local store, then run cross-time queries the live API cannot express: query decay, keyword cannibalization, coverage drift, opportunity-with-baseline, and a cross-property mover board. Every command supports `--json`, `--select`, `--csv`, `--dry-run`, and ships an MCP server out of the box.

**API:** google-search-console | **Category:** developer-tools | **Press version:** 4.0.2
**Spec:** https://api.apis.guru/v2/specs/googleapis.com/searchconsole/v1/openapi.yaml

### 11 transcendence commands no other GSC tool can express

- **`book`** — Cross-property mover board (UNION across all sites + delta + per-site rollup)
- **`cannibalize`** — Same-site SERP competition (`GROUP BY (site, query) HAVING COUNT(DISTINCT page) > 1`)
[...]

PR #450, opened 2026-05-11
feat(google-search-console): Indexing API write surface

heyimjames:
## Summary

Adds three write commands that hit Google's Indexing API (indexing.googleapis.com), so an agent can ask Google to crawl/recrawl URLs found by the diagnostic transcendence commands (`opportunity`, `decay`, `new-queries`, etc.) instead of only reporting them.

**Drafted as the first half of a planned write-surface expansion.** Local-store writes (annotations, triage state, watches) will land in a separate follow-up PR to keep review surfaces small.

## Commands added

- `index-submit <url> [--type URL_UPDATED|URL_DELETED]` — single URL notification (counts toward the 200/day project quota).
- `index-status <url>` — read the latest notification metadata Google has on file for the URL. Useful for dedupe before re-submitting.
- `index-bulk-submit --from-file urls.txt [--type ...] [--rate N] [--max N]` — batch with per-second rate limit and per-run cap.

## Infrastructure changes

- `client.do()` now treats path values beginning with `http://` or `https://` as absolute URLs, bypassing the `BaseURL` prepend. This lets one `Client` speak to both `webmasters.googleapis.com` and `indexing.googleapis.com` without spinning up a parallel client. One-line change in the URL-resolution branch; relative paths still concatenate as before.
- `auth.go` requests the `https://www.googleapis.com/auth/indexing` OAuth scope alongside the existing webmasters scopes. **Users will be prompted to re-auth on first use after upgrade** — flagged for release notes.

## MCP annotations

- `index-status` → `mcp:read-only=true` (no permission prompt; safe for agent loops).
- `index-submit`, `index-bulk-submit` → `mcp:read-only=false` (MCP hosts prompt before each run, which is correct since the call produces a real GSC-side notification).

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./...` passes
- [x] `verify_skill.py` passes — flag-names, flag-commands, positional-args, unknown-command all clean
- [x] Scorecard unchanged at 80/A (this PR adds write capability rather than score signal)
- [ ] Live exercise once auth re-grant has fired in a test account (cannot run from this dev machine without OAuth setup)

## Dependency

**This PR builds on top of #398.** Targeting `main` so once #398 merges the diff naturally shrinks to just these four files. While #398 is open, this PR's diff will show its commits too. Draft until #398 merges.


tmchow:
Looks abandoned. Happy to relook at this if you want to reopen.

PR #283, opened 2026-05-07
feat(google-search-console): add google-search-console

heyimjames:
## google-search-console

Search Console API — search analytics, sites, sitemaps, and URL inspection. **Every Search Console feature, plus a local SQLite store that turns one-shot API calls into time-series insights no other GSC tool ships.**

Sync your search analytics, sites, sitemaps, and URL inspection state into a local store, then run cross-time queries the live API cannot express: query decay, keyword cannibalization, coverage drift, opportunity-with-baseline, and a cross-property mover board. Every command supports `--json`, `--select`, `--csv`, `--dry-run`, and ships an MCP server out of the box.

**API:** google-search-console | **Category:** developer-tools | **Press version:** 4.0.2
**Spec:** https://api.apis.guru/v2/specs/googleapis.com/searchconsole/v1/openapi.yaml

### What This CLI Does

This is the GSC ecosystem's first agent-native CLI with a real local store. It absorbs every command exposed by the leading GSC tools (Bin-Huang's npm CLI, ahonn/AminForou MCPs, joshcarty's Python SDK), then adds **11 transcendence commands** that no other tool can express because they require persistent storage:

- **`book`** — Cross-property mover board (UNION across all sites + delta + per-site rollup)
- **`cannibalize`** — Same-site SERP competition (`GROUP BY (site, query) HAVING COUNT(DISTINCT page) > 1`)
- **`decay`** — Linear-fit slope per query over 16 months of daily rows
- **`coverage-drift`** — Diff successive `url_inspections` snapshots; catches silent deindexings
- **`opportunity`** — Quick-wins joined to prior snapshots for entry-date baseline
- **`new-queries`** (with `--lost`) — Anti-join surface for "what's new since last visit"
- **`triage`** — Non-INDEXED pages × historical impressions; fix broken-but-trafficked first
- **`momentum`** — Recent vs trailing baseline, both directions
- **`territory`** — Cross-time `(query × country × device)` share shifts
- **`appearance`** — searchAppearance dimension across windows
- **`sitemap-health`** — Cross-snapshot diff with regression flags

Plus the spec-derived `webmasters` (sites + sitemaps + searchanalytics) and `url-inspection` / `url-testing-tools` shortcuts, plus `quickwins`, `compare`, and `inspect-batch` absorbs that wrap competitor shapes with our local-store backing.

[...]

PR #451, opened 2026-05-11
feat(google-search-console): agent state — annotations + triage state

heyimjames:
## Summary

Stacks on top of #398 and #450. Adds an agent-owned notebook layer in the local store: pages, queries, sites, and triage items can carry persistent notes and tags. Triage-state mutations (resolve / snooze / unresolve) are sugar over the annotations API with reserved tags.

Together with #450, this completes the planned write-surface expansion. After both merge, the GSC MCP goes from "read agent with a small write surface" (6 writes) to "stateful agent surface" (15 writes — 5 GSC-state-touching writes from #283 and #450, plus 10 local-state writes from this PR).

## Schema

```sql
annotations(id, target_type, target, note, tags, expires_at,
            created_at, updated_at)
idx_annotations_target on (target_type, target)
idx_annotations_tags on (tags)
```

`CREATE TABLE IF NOT EXISTS` makes this forward-compatible with older binaries that don't know about the table — no `StoreSchemaVersion` bump needed.

## Commands added

| Command | What | MCP annotation |
|---|---|---|
| `annotate <type> <target> [--note] [--tag ...]` | Add an annotation | write |
| `annotations [--type] [--target] [--tag] [--include-expired]` | List annotations | read-only |
| `annotation-remove <id>` | Delete by id | write |
| `triage-resolve <id> [--note]` | Tag triage item "resolved" | write |
| `triage-snooze <id> --until 7d [--note]` | Tag "snoozed" with expiry | write |
| `triage-unresolve <id>` | Strip resolved/snoozed tags | write |

`type` ∈ `{page, query, site, triage}`.

## Why this matters

Before: an agent that diagnoses a problem (e.g., `decay` flags a page losing rank) has no persistent way to mark "I've handled this" or "snooze for a week." Every new session re-flags the same items.

After: the agent can take notes, mark items resolved, snooze them with an expiry, and recall its own state across runs. That changes the character of the MCP from per-call diagnostic to a stateful agent surface that supports recurring weekly/daily loops.

## MCP safety

All writes here are **local-store only** — no Google API impact, nothing visible in Search Console. So `mcp:read-only=false` but not destructiveHint. MCP hosts will prompt once and then allow the agent to manage its own workspace freely.

The `annotations` (list) command stays `mcp:read-only=true` so reads don't prompt at all.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./...` passes
- [x] `verify_skill.py` clean (flag-names, flag-commands, positional-args, unknown-command)
- [ ] Integration: an agent loop that uses `triage` → `triage-resolve` → re-run `triage` should see the resolved item gone. *This requires modifying the `triage` command to honor the resolved-state filter, which lands in a follow-up PR to keep this diff focused on the data layer.*

## Stacking

Targets `main`. While #398 and #450 are open, this PR's diff includes their commits. Order of intended merge: #398 → #450 → this. Each auto-retargets and shrinks as predecessors merge.


tmchow:
Looks abandoned. Happy to relook at this if you want to reopen.

PR #466, opened 2026-05-11
feat(google-search-console): add search analytics workflow shortcuts

andreslfdz:
## Summary
- add brand-vs-nonbrand-split with branded-split and brand-split aliases
- add page-queries with page-query-breakdown alias for page-specific query rows
- add which-index discovery and generated-tree patch metadata for the GSC CLI customization

## Validation
- cd library/marketing/google-search-console && go test ./...
- cd library/marketing/google-search-console && go vet ./...
- cd library/marketing/google-search-console && go build ./...

## Notes
- Both commands are read-only live Search Analytics shortcuts and are exposed to MCP through the existing Cobra command mirror.
- Live authenticated GSC API calls were not run in this PR.


tmchow:

PR #478, opened 2026-05-11
feat(flight-goat): add watch command group for purchased-flight price monitoring

gitaaronta:
## Summary

Adds a new `flight-goat-pp-cli watch` command group for monitoring purchased-flight prices. Users register a flight they've already booked; `watch check` re-queries Google Flights and alerts when the *same exact itinerary* (airline + flight number + date + route + cabin) appears at a lower price than `paid - threshold`. State persists in its own SQLite file (decoupled from the generator-owned `internal/store` schema) and every alert carries a stable JSON envelope plus a mandatory rebooking safety notice.

## What Changed

- New package `library/travel/flight-goat/internal/watch/` — `Watch` model with validation, SQLite store (modernc.org/sqlite, schema v2 with forward-only migration), match logic, webhook + stdout dispatch with dedup via `last_alerted_price`.
- New hand-written extension `library/travel/flight-goat/internal/cli/watch.go` — Cobra commands `watch add | list | show | check | remove | alert-test`. Mirrors the existing `primary.go` extension pattern.
- One-line hook into `internal/cli/root.go` (`registerWatchCommands(rootCmd, flags)`), tagged with a `// PATCH(library):` comment and recorded in `.printing-press-patches.json`.
- Tests covering: validation, persistence round-trip, exact-match dispatch, threshold suppression, route-only no-alert, missing-flight-number = medium confidence, departure-time drift rejection + within-tolerance acceptance, ExcludeBasic propagation, MatchReason content, BookingURL contents, dedup, force-alert override, JSON-schema shape, safety-notice presence.
- SKILL.md + README.md sections documenting usage, JSON envelope shape, and fare-class safety semantics. `cli-skills/pp-flight-goat/SKILL.md` regenerated via `tools/generate-skills`.

## CLI Shape

```bash
$ flight-goat-pp-cli watch --help
watch tracks flights you have already booked and re-checks their live
prices via the Google Flights backend. When the same flight's fare drops
by more than your threshold you receive an alert (stdout or webhook) with
a safety notice covering refundability, change fees, and credit handling.

Alerts trigger only on EXACT matches (same airline + flight number + date
+ route + cabin). The route-cheapest fare is surfaced for context but is
never used to fire an alert — cheapest-on-route on a different flight
won't help you if your current ticket can't be moved without losing it.

Available Commands:
  add         Register a purchased flight to monitor for price drops
  alert-test  Send a synthetic alert through the watch's notify sink
  check       Re-check live prices and dispatch alerts
  list        List registered watches
  remove      Delete a watch
  show        Show one watch's full state
```

## Design Notes

- **Match-confidence policy.** `high` (exact: airline + flight # + date + route + cabin + optional ±30 min time window) is the only level that alerts. `medium` (same airline + cabin, no flight # echoed) and `low` (route-only) are surfaced as context but never fire. The route-cheapest fare is reported as `route_cheapest_price` so users can see what a different itinerary costs without that being mistaken for "your flight got cheaper."
- **Fare-class safety.** Cabin tier is filtered at search time by Google Flights. Within the economy bucket, `exclude_basic` defaults to true so a $300 basic-economy result never false-matches a $700 main-cabin ticket; users who paid for basic economy opt in with `--include-basic`.
- **Flight-number reuse.** Optional `--departure-time HH:MM` adds a ±30 min sanity check — catches the case where an airline reuses the same flight number for a rescheduled departure later in the day. The matcher demotes from `high` to `low` and sets `match_mismatch_reason` so the alert can explain "your flight # ran but at a different time" instead of going silent.
- **Reasoning surfaced.** Every alert carries `match_reason` — a one-sentence chain of evidence (e.g. `"exact match: same airline DL, flight 668; date 2026-06-21; route SFO→JFK; cabin economy; basic-economy excluded; departure 11:30 within ±30 min of your 11:30"`) so users (and downstream agents) can sanity-check the conclusion.
- **Booking URL.** Each alert includes a Google Flights search URL pre-filled with route + date + cabin + currency. Uses the search-URL form (`?q=...`) rather than the itinerary deep-link (which requires an opaque base64 protobuf) so the link stays durable.
- **State isolation.** Watches live in their own SQLite file (default `~/.local/share/flight-goat-pp-cli/watches.db`, override with `--watch-db` or `$FLIGHT_GOAT_WATCH_DB`) to avoid coupling with the generator-owned `internal/store` migration schedule.
- **Generator path.** This is a hand-written extension; cli-printing-press has no template for stateful long-running monitors today. Recorded in `.printing-press-patches.json` (`watch-command-group`) under `deferred_to_upstream` so future regeneration can absorb it behind a generator hook.

## Match Confidence Truth Table

| Scenario | Confidence | Alert? |
|---|---|---|
| Same airline + flight # + cabin + date + route, departure within ±30 min of user's time (or time unset) | `high` | ✅ if threshold crossed |
| Same airline + flight # + cabin + date + route, but departure drifts > ±30 min | `low` + `match_mismatch_reason` | ❌ |
| Same airline + cabin + date + route, provider didn't echo flight # | `medium` | ❌ (informational) |
| Different airline / different flight # cheaper on the same route | `low` | ❌ (`route_cheapest_price` set for context) |

## Validation

- [x] `python3 .github/scripts/verify-skill/verify_skill.py --dir library/travel/flight-goat/` → all checks passed
- [x] `go build ./...` from the CLI root → clean
- [x] `go vet ./...` from the CLI root → clean
- [x] `go test ./...` from the CLI root → all packages PASS (internal/cli, internal/cliutil, internal/gflights, internal/store, internal/watch)
- [x] `govulncheck ./...` from the CLI root → No vulnerabilities found
- [x] `go run ./tools/generate-skills/main.go` → mirrored 78 skills, including pp-flight-goat
- [x] `git diff --check` → no whitespace issues
- [x] Live end-to-end smoke test with a real Google Flights query (DL 668 SFO→JFK 2026-06-21): exact match resolved, JSON envelope clean, webhook + stdout dispatch verified, safety notice rendered, booking URL pre-filled.

## Gaps / Follow-Up

- The watch DB lives at `~/.local/share/flight-goat-pp-cli/watches.db`. If the generator later grows a stateful-monitor template, the `internal/watch/` package and the `registerWatchCommands` hook should migrate behind it (recorded in `.printing-press-patches.json`).
- Currency-drift handling: today we don't FX-convert if Google's response is tagged in a different currency than the watch. The compare uses the watch's currency directly. If the response currency differs, the comparison would be incorrect — currently rare but worth a follow-up.
- Google Flights doesn't reliably expose fare brand codes (Y, K, M, etc.), so `fare_brand` is a free-text user note for context only; we can't auto-verify brand-vs-brand matches.

### DIFF
--- a/cli-skills/pp-flight-goat/SKILL.md
+++ b/cli-skills/pp-flight-goat/SKILL.md
@@ -63,6 +63,119 @@ The flag is only valid on Google Flights price commands: `flights`, `dates`,
 `compare`, `gf-search`, and `cheapest-longhaul`. Do not add it to AeroAPI or
 Kayak-only commands.
 
+## Watch a Purchased Flight for Price Drops
+
+`watch` monitors a flight the user has *already booked* and alerts when the
+same exact itinerary (airline + flight number + date + route + cabin) reappears
+on Google Flights below `paid - threshold`. Use it when the user says "did my
+flight get cheaper?" or "watch DL 669 for me."
+
+State is persisted in a local SQLite file (default
+`~/.local/share/flight-goat-pp-cli/watches.db`, override with `--watch-db` or
+`$FLIGHT_GOAT_WATCH_DB`). Subcommands: `add`, `list`, `show`, `check`,
+`remove`, `alert-test`.
+
+```bash
+# Register a watch (all required fields)
+flight-goat-pp-cli watch add \
+  --from SFO --to JFK --date 2026-06-21 \
+  --airline DL --flight-number 669 --cabin economy \
+  --paid 428.20 --threshold 50 --currency USD \
+  --notify webhook:https://hooks.example.com/flight-drops --agent
+
+# Stricter watch with departure-time guard, explicit fare brand, and
+# basic-economy excluded so a $300 basic fare won't false-match your
+# $700 main-cabin ticket
+flight-goat-pp-cli watch add \
+  --from SFO --to JFK --date 2026-06-21 --departure-time 07:30 \
+  --airline DL --flight-number 668 --cabin economy --fare-brand "Main Cabin" \
+  --paid 700 --threshold 50 --agent
+
+# Add a watch on a basic-economy ticket you actually paid for (opt-in)
+flight-goat-pp-cli watch add \
+  --from LAX --to ATL --date 2026-07-10 \
+  --airline DL --flight-number 1234 --cabin economy --fare-brand "Basic Economy" \
+  --include-basic --paid 198 --threshold 30 --agent
+
+# List, inspect, re-check
+flight-goat-pp-cli watch list --agent
+flight-goat-pp-cli watch show watch_5a25c89c2966 --agent
+flight-goat-pp-cli watch check watch_5a25c89c2966 --agent     # one watch
+flight-goat-pp-cli watch check --agent                        # all active watches
+
+# Verify the alert path without burning a Google Flights request
+flight-goat-pp-cli watch alert-test watch_5a25c89c2966 --agent
+
+# Force a re-alert after a previous one fired (dedup is on by default)
+flight-goat-pp-cli watch check watch_5a25c89c2966 --force-alert --agent
+
+# Re-alert only when the price drops by at least 25 more
+flight-goat-pp-cli watch check --repeat-delta 25 --agent
+
+flight-goat-pp-cli watch remove watch_5a25c89c2966 --agent
+```
+
+`watch check` returns a stable JSON envelope (also POSTed verbatim to
+`--notify webhook:`):
+
+```json
+{
+  "schema": "flight-goat.watch.check.v1",
+  "watch_id": "watch_…",
+  "origin": "SFO", "destination": "JFK",
+  "departure_date": "2026-06-21", "departure_time": "07:30",
+  "airline": "DL", "flight_number": "668",
+  "cabin": "economy", "fare_brand": "Main Cabin",
+  "original_price": 700, "threshold": 50, "currency": "USD",
+  "booking_url": "https://www.google.com/travel/flights?q=Flights+from+SFO+to+JFK+on+2026-06-21+economy&curr=USD",
+  "found_price": 609, "route_cheapest_price": 280, "delta": 91,
+  "confidence": "high",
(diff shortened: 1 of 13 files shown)


tmchow:
@gitaaronta fix code review feedback

gitaaronta:
Thanks for the review — pushed `dd61dcb8` addressing all three findings on `feat/watch-command`:

**P1 — `departureTimeMatches` midnight wraparound** (`internal/watch/check.go`)
- Added the suggested 12h-then-`24h - diff` wrap. 23:50 watch ↔ 00:10 candidate now correctly returns a 20-min gap (within tolerance) instead of 23h40m.
- Regression test `TestDepartureTimeMatchesWrapsMidnight` covers both directions of the wrap plus the 12h-equator boundary.

**P1 — `ForceAlert` bypassed confidence filter** (`internal/watch/check.go`)
- `case opts.ForceAlert:` now requires `confidence == MatchExact`. When `matched != nil` but confidence is `MatchProbable`, we set `AlertSuppressed: true` with the reason `"force-alert requested but match confidence below high (refusing to alert on a flight the user may not hold)"`.
- Regression test `TestCheckForceAlertSuppressesMatchProbable` asserts dispatcher is never called AND `LastAlertedPrice` stays nil, so the next legitimate high-confidence alert isn't dedup-suppressed.

**P2 — `stdoutDispatcher` silently dropped on nil writer** (`internal/watch/alert.go`)
- `stdoutDispatcher.Dispatch` now falls back to `os.Stdout` when no writer was injected via `SetStdoutWriter`. The CLI path still wires `cobra.Command.OutOrStdout()` so `--deliver` and JSON output capture remain unaffected.
- Regression test `TestStdoutDispatcherFallsBackToOsStdout` redirects `os.Stdout` to a pipe and asserts the alert text (incl. safety notice) actually lands there.

**Validation:**
- `go build ./...` · `go vet ./...` · `go test ./...` (all packages PASS; the watch package now has 14 test functions, up from 11)
- `python3 .github/scripts/verify-skill/verify_skill.py --dir library/travel/flight-goat/` → all checks passed
- `govulncheck ./...` → No vulnerabilities found

The fix commits live alongside the original feature commit on `feat/watch-command`. Ready for re-review.

tmchow:
Fix conflicts

tmchow:
@gitaaronta fix merge conflicts

tmchow:

tmchow:
Fix conflicts

tmchow:

PR #480, opened 2026-05-12
feat(greatclips): add greatclips

mvanhorn:
## greatclips

First CLI for Great Clips Online Check-In. Drives the GreatClips customer webservices (`webservices.greatclips.com`) and the ICS Net Check-In wait/check-in service (`www.stylewaretouch.net`) from one Go binary.

**API:** greatclips | **Category:** other | **Press version:** 4.2.2
**Spec:** internal YAML (browser-sniff sourced)

### Publication Path

New print.

### Novel Commands

| `wait` | Live wait time at any salon | Fetch real-time estimated wait minutes for a specific salon, signed with the ICS Net Check-In HMAC. |
| `checkin` | Party check-in via CLI | Add yourself plus up to 4 additional party members to a salon waitlist with one command. |
| `status` | Active check-in status lookup | Query the live position-in-line and estimated wait for the current account's active check-in. |
| `cancel` | Cancel active check-in | Remove yourself from the salon waitlist without opening the SPA. |
| `customer profile` | Customer profile lookup | Fetch the authenticated customer profile via the Auth0 customer-audience token. |

### What This CLI Does

Talks to two host families that share a single Auth0 customer identity: `webservices.greatclips.com` for salon metadata + profile + hours, and `www.stylewaretouch.net` (the ICS Net Check-In vendor) for wait times, check-in, status, and cancel. The non-trivial piece is `stylewaretouch.net` — every request is HMAC-SHA-256-signed with a 32-byte seed derived from a label-based XOR; this CLI ports that signing scheme byte-identical to the SPA (golden vector test included). Customer-audience JWTs are HS512 tokens minted server-side; the CLI accepts them via `GREATCLIPS_TOKEN` env var plus `GREATCLIPS_PROFILE_ID` for the check-in identity body.

### Manuscripts

Included under `.manuscripts/20260511-095555/` with research brief, absorb manifest, shipcheck, and acceptance evidence.

### Validation Results

| Check | Result |
|-------|--------|
| Manifest | PASS |
| Transcendence | PASS |
| Phase 5 | PASS |
| go mod tidy | PASS |
| go vet | PASS |
| go build | PASS |
| --help | PASS |
| --version | PASS |
| verify-skill | PASS |
| Manuscripts | PASS |

### Gaps

v0.2 of this CLI requires per-session JWT paste (`GREATCLIPS_TOKEN` + `GREATCLIPS_PROFILE_ID`) captured from a logged-in browser DevTools session. Full auto-discovery of the GreatClips token-exchange endpoint (which converts the Auth0 `/cmp`-audience token into the customer-audience HS512 token used for API calls) is scoped as v0.3 future work using `agent-browser` for one-time exchange capture.

macOS-only Chrome cookie extraction (`internal/auth0silent/cookies_darwin.go`); other platforms get a clear error and the paste-token flow still works.


tmchow:
@mvanhorn address code review feedback

tmchow:

mvanhorn:
Thanks @greptileai - all 6 findings addressed in adf4040d:

- P1 `cookies_darwin.go:141` SQLi: added an allowlist guard that returns an error if `host != CIDHost` before the fmt.Sprintf, since sqlite3 CLI doesn't expose parameter binding via -separator. Any future caller passing user input hits the allowlist error instead of splicing into the SQL.
- P2 `cookies_darwin.go:222` 32-byte strip: now compares the first 32 plaintext bytes against `sha256(CIDHost)` before stripping. Older-Chrome cookies (no host-binding prefix) keep their full plaintext; M120+ cookies get the binding tag removed. Threaded the hostKey through decryptCookie so the check is explicit.
- P2 `silent.go:133` state: added `if frag.Get("state") != state { return error }` right before the access_token read. Closes the CSRF round-trip even though the CLI doesn't have a browser redirect surface.
- P2 `icssign_hook.go:83` debug block: deleted. The annotated 'remove before publish' note is now honored; signing input + outgoing HMAC are no longer echoed to stderr under `GREATCLIPS_DEBUG_SIGN=1`. Dropped the now-unused `os` and `fmt` imports.
- P1 `client.go:432` unreachable refresh: replaced the silent `tokenURL := ""; return nil` with an explicit error: `greatclips access token expired and no refresh-token flow is configured; re-run "greatclips auth login" to mint a fresh token`. Removed the rest of the dead function body. Token minting goes through auth0silent.Mint; there's no OAuth2 refresh_token exchange to wire here.
- P1 `promoted_checkin.go:83` silent Atoi drop: now returns `usageErr(fmt.Errorf("GREATCLIPS_PROFILE_ID=%q is not a valid integer: %w", v, err))` so misformatted values fail loudly with a clear message.

All six recorded in `.printing-press-patches.json` with PATCH comments at the call sites. Build + vet + go test (cli + cliutil + icssign) + verify-skill all green.

PR #497, opened 2026-05-12
feat(granola): add granola

DamienStevens:
## granola

A Go CLI for Granola.ai's meeting note-taking app — cache-first, agent-native, with MEMO pipeline runner, attendee timelines, cross-meeting analytics, and three-stream extraction (human notes, AI panels, transcript).

**API:** granola | **Category:** productivity | **Press version:** 4.4.0
**Spec:** `docs.granola.ai/api-reference/openapi.json` (3 endpoints) + reverse-engineered internal API (6 endpoints, via getprobo) + local cache schema

### Publication Path

New print

### CLI Shape

```
Granola CLI — Every Granola feature — plus offline SQLite cross-meeting search, attendee timelines, and a MEMO pipeline runner no oth…

Highlights (not in the official API docs):
  • memo run   Run the preflight → extract pipeline on one meeting or every new meeting since a timestamp, emitting the MEMO three-file artifact and an ndjson run-state ledger.
  • memo queue   List every meeting whose transcript is in the cache but whose MEMO triple is not yet on disk.
  • attendee timeline   Every meeting with a given attendee, ordered oldest→newest, with title, date, folder, and recipe-applied flag per row.
  • attendee brief   Pulls the last N meetings with an attendee and stitches together their real cached notes plus real AI panel summaries — no synthesis.
  • folder stream   ndjson stream of every meeting in a Granola folder (resolved via documentLists + listRules) with notes and a named panel inlined.
  • recipes coverage   Surface meetings that did NOT have a named panel template/recipe applied within a date range.
  • talktime   Per-segment-source talk-time for one meeting — microphone (you) vs system (everyone else) in minutes.
  • talktime   Lifts the per-source talk-time aggregation across N meetings since a date — who-talked-most over time.
  • chat list   List and dump Granola’s AI chat threads anchored to a meeting (entities.chat_thread + entities.chat_message in the cache).
  • duplicates scan   Hash (title, date-bucket, attendee-email-set) across the cache and a meeting-transcripts repo to surface duplicates at scale.
  • tiptap extract   Render documents[id].notes (TipTap JSON: headings, bullet_list, list_item, bold marks, paragraph_break) to canonical markdown instead of falling back to notes_plain.
  • calendar overlay   Left-anti-join meetingsMetadata calendar events with documents.google_calendar_event to find calendared-but-not-recorded meetings.

Agent mode: add --agent to any command for JSON output + non-interactive mode.
Health check: run 'granola-pp-cli doctor' to verify auth and connectivity.
[...]

PR #510, opened 2026-05-12
feat(granola): handle encrypted local cache (cache-v6.json.enc + supabase.json.enc)

mvanhorn:
Follow-up to PR #497 (merged 2026-05-12 by @tmchow). The original granola CLI ships pre-encryption code; this PR adds the encryption-handling layer for Granola desktop 7.205.0+, which writes `cache-v6.json.enc` and `supabase.json.enc` instead of the plaintext files the loaders previously read.

- `internal/granola/safestorage`: new package, two-tier decrypt (Electron safeStorage v10 over `storage.dek`, then AES-256-GCM with the recovered DEK on the `.enc` files). macOS only this round; Linux/Windows stubbed with `ErrUnsupportedPlatform`. Empirical scheme finding documented at `internal/granola/safestorage/testdata/scheme.md`.
- `cache.go` + `workos.go`: prefer the `.enc` sibling when present via a new `ResolveCachePath`; fall back to plaintext for pre-encryption Granola installs. Decrypt failures surface as wrapped errors rather than silently falling back to the stale plaintext stub.
- D6: `RefreshAccessToken` refuses to rotate when source is `supabase.json.enc`. WorkOS uses single-use rotation; rotating the token via the CLI would sign Granola desktop out next time it tries to refresh. Plaintext, env-override, and stored-accounts sources still allow refresh.
- `sync.go`: hydrates documents via `/v2/get-documents` before `SyncFromCache` runs (Granola moved meeting documents out of the local cache at the same time as the encryption rollout; without this, `meetings list` and friends return empty). `openGranolaCache` also backfills `cache.Documents` from the SQLite store so post-sync commands like `show`, `notes-show`, `memo queue` work.
- `doctor`: new Encrypted store check that reads a new sync state file at `~/.local/share/granola-pp-cli/sync_state.json`. Never invokes `safestorage.Decrypt` itself, so running `doctor` never triggers a Keychain prompt.

End-to-end verified on Granola 7.205.0 macOS: `sync` reports 745 meetings + 10,425 transcript segments; `meetings list`, `show`, `notes-show`, `memo queue` all return real data. 206 tests pass.

Customizations recorded in `library/productivity/granola/.printing-press-patches.json` (6 entries) with matching `// PATCH:` markers in the referenced files.


mvanhorn:
Addressed Greptile's P2 polish items in 27df... (force-push wasn't needed; new commit on top):

- **P2 #1** (sync_cache.go / syncstate.go): hydration errors now land in `LastHydrateErrorMsg`, not `LastDecryptErrorMsg`. Doctor surfaces it with a distinct hint.
- **P2 #3** (api_documents.go): pagination now consults `has_more` from the wrapped envelope; saves the one-extra-round-trip on exactly-100-doc edge case. New `InternalClient.GetDocumentsPage` sibling exposes the full envelope.
- **P2 #4** (workos_test.go): renamed the misleading test, added a real encrypted-source test that uses the committed safestorage fixture.

**P2 #2 (Keychain ACL trade-off)**: not changing. The shell-out to `/usr/bin/security` vs. a CGO Security.framework call is a deliberate design choice — see Decision D1 in `docs/plans/2026-05-12-001-feat-granola-encrypted-cache-plan.md`. Trade-offs: CGO adds build complexity (cross-compile, goreleaser, no longer pure Go) for a personal-machine CLI where the ACL-on-/usr/bin/security blast radius is acceptable. Code comments at the `fetchKeychainEntry` definition document this. Worth re-evaluating if this CLI ever ships to multi-user / shared-machine contexts, but not for the current scope.

Tests: 208 passing.

PR #328, opened 2026-05-09
feat(youtube): add youtube

Beomgithb:
## youtube

**Every YouTube Data API v3 endpoint, plus quota-aware planning, plus offline transcripts (no quota), plus an SQLite-backed store so channel and trending analysis cost zero quota after sync.**

**API:** youtube | **Category:** media-and-entertainment | **Press version:** 4.2.0
**Spec:** [apis.guru googleapis.com/youtube v3 (OpenAPI 3.0.0)](https://api.apis.guru/v2/specs/googleapis.com/youtube/v3/openapi.yaml)

### CLI shape

The CLI exposes 93 commands: 76 endpoint mirrors under `youtube <resource>-<method>` covering every documented Data API v3 read/write operation, plus the framework commands (`doctor`, `auth`, `sync`, `agent-context`, `profile`, `feedback`, `analytics`, `workflow`, `import`, `tail`, `which`), plus 8 hand-built transcendence commands and supporting helpers (`transcripts`, `sync-channel`).

### What this CLI does

Existing tools split the YouTube surface — yt-dlp owns downloads, MCP servers own metadata reads, no one lets an operator do all three plus track quota plus query a cached corpus. `youtube-pp-cli` mirrors all 23 Data API resources, shells out to `yt-dlp` for transcripts (no quota cost), and ships a local SQLite store with FTS5 indexes across video metadata, transcripts, and comments.

### Unique features (8 transcendence commands)

| Command | What it does |
|---|---|
| `quota plan <endpoint>...` | Project unit cost vs. local 24h ledger before running a real call |
| `cost ledger` / `cost log` | Read/write the per-API-key quota_log |
| `corpus search "<q>"` | FTS5 union over videos × transcripts × comments — zero quota |
| `digest <channel>` | Channel uploads × transcript-FTS keyword join (Maya's Friday newsletter) |
| `trending snapshot` / `trending diff` | Capture and diff `mostPopular` lists across dates |
| `velocity <channel>` | Δviews/Δlikes/Δcomments per day from sync snapshot history |
| `topic crossover "<kw>"` | trending × videos × transcripts join, ranked by trending position |
| `subscriptions sweep` | OAuth-gated: rebuild the in-app subscription feed locally |

### Auth

- **API key** (read): `YOUTUBE_API_KEY` → attached as `?key=` query param. Read-only path.
[...]

PR #518, opened 2026-05-13
feat(youtube): add youtube (resubmit of #328 with review fixes)

Beomgithb:
Resubmit of #328 with the 4 review items addressed in `08becb66`:

- **`client.go` cacheKey**: sort param keys before hashing so identical `(path, params)` pairs always produce the same cache file. Go map iteration is randomized; the previous version could thrash the cache.
- **`store.go` SQL injection**: added `validateSQLIdent` (`^[A-Za-z_][A-Za-z0-9_]*$`) and gated the two paths that interpolated caller-supplied strings into `fmt.Sprintf`'d queries — `ListIDs` domain table and `ResolveByName` json_extract field. Identifiers can't be parameter-bound; validation rejects anything outside the safe pattern before the SQL is built.
- **`client.go` authHeader**: treat a zero `TokenExpiry` as "expiry unknown → assume expired" so token files written before `ExpiresIn` was persisted still trigger a refresh. The old guard stranded callers on a stale access token.
- **`shellout.go` MCP flag injection**: validate flag names against `^[a-z][a-z0-9-]*$` and reject `-`-prefixed tokens in the raw `args` passthrough so a client cannot smuggle `--config` or `--credentials-file` into the spawned CLI.

Verified locally:
- `go build ./...` ✅
- `go vet ./...` ✅
- `go test ./internal/{client,store,mcp,mcp/cobratree}/...` — store and mcp PASS, no regressions.

Original PR: #328


Beomgithb:
@tmchow Resubmitting #328 with the 4 review items fixed. PTAL when you have a moment.

tmchow:
If you're resubmitting it you should know it should nto be done manually and instead using the publish process via the skill. your agent is misbehaving again.

PR #565, opened 2026-05-14
fix(npm): detect stale PATH-shadowed binary after go install (#470)

### LINKED ISSUE #470: Installer reports stale shadowed binary after go install
## Summary

`printing-press install table-reservation-goat --cli-only` can successfully build the current CLI with `go install`, but then report/leave a stale binary earlier in `PATH` as the installed binary.

This caused `table-reservation-goat-pp-cli` to appear updated while still running an older OpenTable+Tock-only build, even though the repo/registry/release had already landed Resy support.

## What happened

On Ada's Mac mini:

```bash
$ npx -y @mvanhorn/printing-press install table-reservation-goat --cli-only
Installed table-reservation-goat
  binary: /opt/homebrew/bin/table-reservation-goat-pp-cli

$ table-reservation-goat-pp-cli --version
table-reservation-goat-pp-cli 1.0.0

$ table-reservation-goat-pp-cli --help | head
Cross-network reservation CLI for OpenTable and Tock. Search both networks...
```

The repo was correct:

- `registry.json` describes Table Goat as OpenTable, Tock, and Resy
- `library/food-and-dining/table-reservation-goat/README.md` includes Resy auth and Resy addressing
- release `table-reservation-goat-current` exists with `table-reservation-goat-pp-cli-darwin-arm64`

But there were two binaries:

```bash
$ echo $PATH | tr ':' '\n' | grep -E 'homebrew|go/bin'
/opt/homebrew/bin
/opt/homebrew/sbin
# ~/go/bin was not earlier in PATH

$ /opt/homebrew/bin/table-reservation-goat-pp-cli --help | head
Cross-network reservation CLI for OpenTable and Tock. Search both networks...

$ /Users/ada/go/bin/table-reservation-goat-pp-cli --help | head
Cross-network reservation CLI for OpenTable, Tock, and Resy. Search all three networks...
```

So `go install` had produced the correct Resy-capable binary at:

```text
/Users/ada/go/bin/table-reservation-goat-pp-cli
```

…but Printing Press verified/reported the first binary found by `commandOnPath(binary)`, which was the stale one at:

```text
/opt/homebrew/bin/table-reservation-goat-pp-cli
```

Because `/opt/homebrew/bin` came first in PATH, normal invocations kept running the old binary.

## Expected behavior

After `printing-press install ... --cli-only` or `printing-press update`, the command users run from PATH should be the freshly installed binary, or the installer should fail/warn loudly if an older binary earlier in PATH shadows the freshly built one.

## Actual behavior

The installer prints success and reports the stale shadowing binary path, even though the updated binary exists elsewhere.

That makes the install look successful while users continue to run old behavior.

## Likely root cause

In `dist/src/commands/install.js`, install does roughly:

1. `go install <modulePath>@latest`
2. `commandOnPath(binary)`
3. report the first PATH match

But `go install` writes to `GOBIN`, or `$GOPATH/bin` when `GOBIN` is unset. That is not necessarily the same location as the first existing binary in PATH.

So any previously installed binary earlier in PATH can shadow the new `go install` output.

This is not Table Goat-specific; it can affect any Printing Press CLI.

## Possible fixes

Minimum fix:

- Determine the actual Go output path after install (`go env GOBIN`; if empty, `go env GOPATH` + `/bin/<binary>`)
- Compare it to `commandOnPath(binary)`
- If they differ and the PATH match is earlier/stale, warn or fail with a clear remediation

Better UX:

- If a binary already exists earlier in PATH, replace that exact path with the newly built binary, or ask before replacing
- Or install release assets into a managed directory and ensure that managed directory is the reported/expected PATH entry

Suggested warning text:

```text
Installed table-reservation-goat-pp-cli to /Users/ada/go/bin/table-reservation-goat-pp-cli, but /opt/homebrew/bin/table-reservation-goat-pp-cli appears earlier in PATH and will shadow it. Move/remove the old binary or add /Users/ada/go/bin earlier in PATH.
```

## Local workaround used

I manually replaced the stale Homebrew-path binary with the freshly built one:

```bash
cp /opt/homebrew/bin/table-reservation-goat-pp-cli /opt/homebrew/bin/table-reservation-goat-pp-cli.bak-20260511140945
install -m 0755 /Users/ada/go/bin/table-reservation-goat-pp-cli /opt/homebrew/bin/table-reservation-goat-pp-cli
```

After that:

```bash
$ table-reservation-goat-pp-cli goat --help | grep -i network
--network string  Restrict to one network (opentable, tock, resy); default queries all three
```

rob-coco:
## Summary

- `npx -y @mvanhorn/printing-press install …` was reporting the first PATH match as the freshly installed binary. When an older copy lived earlier in PATH (e.g. `/opt/homebrew/bin/<cli>` shadowing a newer `~/go/bin/<cli>`), the installer printed `Installed <cli>` with the stale path and the user kept running stale behavior. Reported in #470 for `table-reservation-goat` but it can affect any Printing Press CLI.

## What Changed

- New `goInstallDir()` helper in `npm/src/go.ts` runs `go env GOBIN GOPATH` and returns the directory `go install` actually writes to (GOBIN if set, otherwise `<GOPATH>/bin`, with Windows backslash handling).
- `install.ts` now compares that path to `which`/`where` after a successful `go install` and:
  - emits a loud `WARNING: installed <cli> at <fresh>, but <stale> appears earlier in PATH and will shadow it` on stderr when they differ,
  - reports the freshly installed path on the `binary:` success line (not the shadow), and prints `shadowed by: <stale>` underneath,
  - exposes both via new `installedPath` and `shadowedBy` JSON fields,
  - if `which`/`where` finds nothing and we know the install dir, names the specific path the user needs to add to PATH (instead of the generic "$(go env GOPATH)/bin" hint).
- `update.ts` is unchanged: it delegates to `installCommand`, so the same detection now applies to `printing-press update <name>` and bulk updates.
- 9 new unit tests covering: shadow detected → warn + correct paths + JSON shape, no warning when paths match, GOBIN unset → falls back to GOPATH/bin, `go env` failure → graceful fallback to PATH match, install dir not on PATH → error names the dir.

## Validation

- [x] `npm run build` (clean)
- [x] `npm test` — 45/45 pass (36 existing + 9 new)
- [x] Existing test contract for "fail when binary not on PATH" preserved; the message now names the exact path

## Gaps / Follow-Up

- A future enhancement could offer to replace the stale shadow binary at that path (with `--force` confirmation), or symlink it. Out of scope for this fix — the warning + JSON metadata give scripts everything they need to decide.

Closes #470.

### DIFF
--- a/npm/CHANGELOG.md
+++ b/npm/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Detect and warn when an older binary earlier in `PATH` shadows the one `go install` just wrote. Previously `install` reported the first PATH hit as success, so a stale `/opt/homebrew/bin/<cli>` (for example) would mask a newer `~/go/bin/<cli>`. The installer now reads `go env GOBIN GOPATH`, compares the actual install path to what `which`/`where` returns, and emits a clear shadow warning when they differ. JSON output adds `installedPath` and `shadowedBy` fields. Fixes #470.
+
 ## 0.1.4
 
 - Drop the `GOPRIVATE='github.com/mvanhorn/*' GOFLAGS=-mod=mod … @main` fallback from the `install` command. The library is fully public, so `go install …@latest` resolves through the public Go module proxy without any private-module configuration. The `@main` retry was only useful when paired with `GOPRIVATE` to bypass the proxy entirely; without it, `@main` issues an identical query subject to the same proxy cache and adds no value.
--- a/npm/src/commands/install.ts
+++ b/npm/src/commands/install.ts
@@ -1,5 +1,5 @@
 import { BUNDLES, isBundle } from "../bundles.js";
-import { detectGo, goInstall, type GoDetection } from "../go.js";
+import { detectGo, goInstall, goInstallDir, type GoDetection, type GoInstallDir } from "../go.js";
 import { commandOnPath, type RunResult } from "../process.js";
 import {
   cliBinaryName,
@@ -25,6 +25,7 @@ interface InstallDeps {
   resolveModulePath: (entryPath: string, registryUrl: string) => Promise<string | null>;
   detectGo: () => Promise<GoDetection>;
   goInstall: (modulePath: string, ref: string, env?: NodeJS.ProcessEnv) => Promise<RunResult>;
+  goInstallDir: () => Promise<GoInstallDir>;
   commandOnPath: (binary: string) => Promise<string | null>;
   installSkill: (skillName: string, agents: string[]) => Promise<RunResult>;
   stdout: (message: string) => void;
@@ -37,7 +38,17 @@ interface InstallSummary {
   binary?: string;
   modulePath?: string;
   skill?: string;
+  /**
+   * The canonical binary path to recommend to the user. Equals `installedPath`
+   * when we can determine it (whether or not PATH agrees), and falls back to
+   * whatever `which`/`where` returned otherwise. Use `shadowedBy` to detect
+   * the shadow case.
+   */
   binaryPath?: string;
+  /** Path `go install` wrote to (derived from `go env GOBIN GOPATH`). */
+  installedPath?: string;
+  /** Set when an older binary earlier in PATH would shadow the freshly installed one. */
+  shadowedBy?: string;
 }
 
 interface InstallOutcome {
@@ -53,6 +64,7 @@ export function createInstallCommand(overrides: Partial<InstallDeps> = {}) {
     resolveModulePath: (entryPath, registryUrl) => fetchGoModulePath(entryPath, registryUrl),
     detectGo: () => detectGo(),
     goInstall: (modulePath, ref, env) => goInstall(modulePath, { ref, env }),
+    goInstallDir: () => goInstallDir(),
     commandOnPath: (binary) => commandOnPath(binary),
     installSkill: (skillName, agents) => installSkill(skillName, { agents }),
     stdout: (message) => console.log(message),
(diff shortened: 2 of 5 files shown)


rob-coco:
Thanks for the review. Addressed:

- **Issue 2 (binaryPath JSDoc)** — reworded the doc to match actual behavior. `binaryPath` is now documented as "the canonical binary path to recommend to the user" — it equals `installedPath` when we can derive it, falls back to the `which`/`where` result otherwise. The shadow case is detected via `shadowedBy`, not by inferring it from `binaryPath`.
- **Issue 3 (goInstallDir per-bundle perf)** — memoized `goInstallDir` for the lifetime of one `installCommandWithDeps` invocation. A `starter-pack` install (4 CLIs) now spawns `go env GOBIN GOPATH` exactly once. Added a regression test that asserts the count.
- **Issue 1 (`bandsintown` printer `"4.5.2"`)** — that was a fork-staleness artifact, not part of this PR's intended changes. My fork's `main` was behind upstream when I branched, so the diff against upstream looked like I was re-adding `bandsintown` to `registry.json` with stale fields. I've rebased the branch onto `upstream/main` and force-pushed; the diff is now npm/* files only. (Reporting the actual bad-printer string in `registry.json` on `main` upstream — if it's still there, that's a separate concern with the regen workflow / merged CLI.)

Force-pushed `0a012a59`; CI should rerun shortly.

rob-coco:
Addressed the trailing-slash edge case in `545affb3`:

- `goInstallDir` now strips trailing `[\\/]+` from both `GOBIN` (used directly) and `GOPATH` (before appending `/bin`), so `GOBIN=/home/user/go/bin/` no longer produces a double-separator path.
- Added two regression tests covering trailing slash on GOBIN and on GOPATH.

48/48 passing.

tmchow:
Hey @rob-coco, this PR has unresolved review feedback from over 6 hours back. Mind taking another pass? If you've hit a blocker or want to hand it off, just leave a note here. Otherwise I'll close for inactivity in about 3 days.

PR #566, opened 2026-05-14
fix(notion): search 400 on empty sort + sync --full 400 on views

### LINKED ISSUE #333: pp-notion: `search` command sends invalid body — missing required `sort.timestamp`
> Spotted by Claude (Anthropic) during a test install on @dubber-icf's machine.

## Summary

`notion-pp-cli search` always fails with HTTP 400 from the Notion API because the request body includes a `sort` object without the required `timestamp` field.

## Reproduction

```
$ notion-pp-cli --version
notion-pp-cli 1.0.0

$ NOTION_TOKEN=ntn_… notion-pp-cli search ""
Error: POST /v1/search returned HTTP 400: {"object":"error","status":400,"code":"validation_error","message":"body failed validation: body.sort.timestamp should be defined, instead was `undefined`."}
```

## Expected

`search` returns a paginated list of pages and databases the integration can see (per [Notion's POST /v1/search docs](https://developers.notion.com/reference/post-search)).

## Actual

The CLI sends a body like `{"sort": {...}}` where the `sort` object is missing the required `timestamp` field. Per Notion's validation, when `sort` is present it must include both `direction` and `timestamp` (currently only `last_edited_time` is supported).

## Likely fix

Either (a) populate `sort.timestamp: "last_edited_time"` by default when the user doesn't pass `--sort`, or (b) omit the `sort` object entirely from the request body when no sort flag is set, letting the API apply its own default.

## Environment

- `notion-pp-cli` 1.0.0 (installed via `go install …@latest` today)
- macOS 25.3.0 (Darwin)
- Go 1.26.3
- Notion Internal Integration token (`ntn_…` format)

### LINKED ISSUE #334: pp-notion: `sync --full` fails on `views` resource — missing required `database_id`/`data_source_id` parameter
> Spotted by Claude (Anthropic) during a test install on @dubber-icf's machine.

## Summary

`notion-pp-cli sync --full` always errors on the `views` resource because the generated request to `GET /v1/views` omits the required `database_id` or `data_source_id` query parameter. Other resources sync successfully; only `views` fails.

## Reproduction

```
$ notion-pp-cli --version
notion-pp-cli 1.0.0

$ NOTION_TOKEN=ntn_… notion-pp-cli sync --full
{"event":"sync_start","resource":"views"}
{"event":"sync_error","resource":"views","error":"GET /v1/views returned HTTP 400: {\"object\":\"error\",\"status\":400,\"code\":\"validation_error\",\"message\":\"At least one of database_id or data_source_id must be provided.\"}"}
{"event":"sync_summary","total_records":33,"resources":6,"success":4,"warned":1,"errored":1}
```

## Expected

`sync --full` enumerates views per database/data source the integration can access, instead of issuing a single unscoped `GET /v1/views`.

## Actual

The sync orchestrator calls `GET /v1/views` with no scope, which the Notion API rejects (the endpoint requires either `database_id` or `data_source_id`).

## Likely fix

After syncing `databases` / `data-sources`, iterate over the discovered IDs and issue one `GET /v1/views?data_source_id=<id>` per data source rather than a single unscoped call.

## Note

Sync exits 0 because of the new `exit_policy_default_changed` semantics, which is reasonable, but `views` will silently never populate until this is fixed.

## Environment

- `notion-pp-cli` 1.0.0 (installed via `go install …@latest` today)
- macOS 25.3.0 (Darwin)
- Go 1.26.3
- Notion Internal Integration token (`ntn_…` format), shared with one database for testing

## Related

- #333 — same install also surfaced an invalid request body in the `search` command

rob-coco:
## Summary

Two Notion CLI bugs reported by @dubber-icf during a fresh install. Both are single-CLI manifestations of API constraints that the generated code didn't model. Hand-edits to `library/productivity/notion/`, recorded in `.printing-press-patches.json` per the repo convention.

### #333 — `search` always returns HTTP 400

`POST /v1/search` was sending `sort: {}` (and empty `filter`, `page_size: nil`, `start_cursor: \"\"`). Notion's API rejects an empty sort object because, when sort is present, it must include both `direction` and `timestamp`. Same constraint applies to `filter` (needs `value` + `property`) and `start_cursor` (must be a real cursor from a prior response).

**Fix:** send only `{\"query\": <query>}` and let the API apply its defaults. Until this CLI surfaces flags for sort/filter/pagination, that's the cleanest body — and matches the issue reporter's option (b).

### #334 — `sync --full` errors on the views resource

`GET /v1/views` requires `database_id` or `data_source_id`; an unscoped call returns 400. Notion exposes no global listing endpoint for databases (`/v1/search` is the only path) or for data sources (they live under `/v1/databases/{id}/data_sources`), so there is no parent-table enumeration today in the local store.

**Fix:** drop `views` from `defaultSyncResources()`. Users who want views populated can run `notion-pp-cli views list --data-source-id <id>` per data source. The dependent `queries` resource already skips cleanly when the `views` parent table is empty (existing \"parent table is empty, sync it first\" code path), so removing views doesn't create a new gap.

Both edits are marked with `// PATCH (...)` source comments. A regression test (`sync_views_test.go`) fails if `views` is re-added to `defaultSyncResources()`.

## Why not fix the generator

Both fits the AGENTS.md \"single-CLI manifestation\" criterion:

- #333 is a Notion-specific API constraint on the search body shape — most search endpoints accept empty optional objects, so a generator change here would either special-case Notion or change a default that other CLIs may rely on.
- #334 is a Notion-specific API constraint plus a Notion-specific listing-endpoint gap. A generator change would need both to detect endpoints that require a scope param and to know how to enumerate the scope, neither of which generalizes from a single API.

## Validation

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` — all green, new regression test passes
- [x] `govulncheck ./...` — \"No vulnerabilities found\"

Closes #333.  
Closes #334.

### DIFF
--- a/library/productivity/notion/.printing-press-patches.json
+++ b/library/productivity/notion/.printing-press-patches.json
@@ -1,6 +1,6 @@
 {
   "schema_version": 1,
-  "applied_at": "2026-05-09",
+  "applied_at": "2026-05-14",
   "base_run_id": "20260508-225745",
   "base_printing_press_version": "4.0.6",
   "patches": [
@@ -22,6 +22,24 @@
         "internal/cli/pages_markdown_update-page.go"
       ],
       "validated_outcome": "pages markdown import <id> --markdown '# Title\\n\\nBody' returns 200 and replaces page content."
+    },
+    {
+      "id": "fix-search-omit-empty-sort-333",
+      "summary": "Send only `query` in the POST /v1/search body; drop the empty sort/filter/page_size/start_cursor fields the API rejects.",
+      "reason": "Reported in #333. Notion's POST /v1/search rejects an empty `sort: {}` because the sort object, when present, must include both `direction` and `timestamp`. The same constraint applies to `filter` (needs `value` + `property`) and `start_cursor` (must be a valid cursor from a prior response). The generated handler unconditionally included all four as empty/nil values, so every `search` call returned HTTP 400 validation_error. Until this CLI surfaces flags for sort/filter/pagination, the cleanest fix is to send only the `query` field and let the API apply its own defaults.",
+      "files": [
+        "internal/cli/search.go"
+      ],
+      "validated_outcome": "notion-pp-cli search \"\" now returns 200 with the API's default ordering; previously returned HTTP 400 'body.sort.timestamp should be defined'."
+    },
+    {
+      "id": "fix-sync-views-requires-scope-334",
+      "summary": "Drop `views` from the default sync resource list because /v1/views cannot be called without a database_id/data_source_id scope.",
+      "reason": "Reported in #334. Notion's GET /v1/views rejects every unscoped call with HTTP 400 'At least one of database_id or data_source_id must be provided.' Notion exposes no listing endpoint for databases or data sources either (databases are discovered via POST /v1/search; data sources via /v1/databases/{id}/data_sources), so there is no parent-table enumeration path in the local store today. `sync --full` therefore always errored on the views resource, and the dependent `queries` resource never had a parent table to iterate. Removing views from defaultSyncResources stops the error; users who want views populated can run `notion-pp-cli views list --data-source-id <id>` per data source.",
+      "files": [
+        "internal/cli/sync.go"
+      ],
+      "validated_outcome": "notion-pp-cli sync --full no longer emits a sync_error for views; the views resource is simply skipped. Per-data-source population via `views list --data-source-id <id>` remains available."
     }
   ]
 }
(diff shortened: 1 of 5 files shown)


rob-coco:
Addressed P2 in `ce8bf4d5` — added a `// PATCH (fix-sync-views-requires-scope-334):` comment to the `syncResourcePath` map next to the `views` entry explaining why it's still mapped (so `--resources views` resolves to the canonical endpoint and surfaces the API's own scope-required error) while being absent from `defaultSyncResources()`.

dubber-icf:
Appreciated. Thanks.

PR #571, opened 2026-05-14
feat(granola): auto-refresh on every CLI invocation

mvanhorn:
## Summary

Auto-refreshes the local SQLite store as the first action of every command, so `meetings list`, `panel get`, etc. no longer return stale data when the caller forgot to run `sync` first. Closes the "agents forget to sync" problem by treating freshness as a CLI-level invariant.

Both auth surfaces refresh independently:
- **Desktop encrypted cache** → `sync` path (cache-v6.json.enc → SQLite). Fires when the encrypted file is present.
- **Public REST API** → `sync-api` path. Fires when `GRANOLA_API_KEY` is set or an access token is in config.

When both are configured, both refresh routines run (cache first, then api). When neither is set, auto-refresh is a silent no-op.

Plan: `docs/plans/2026-05-14-001-feat-cli-auto-refresh-every-invocation-plan.md`

## Key design decisions

- **Hook location:** `PersistentPreRunE` in `internal/cli/root.go`, fired after agent-mode expansion + data-source validation so flag/profile/env precedence resolves correctly first.
- **Skip list:** `sync`, `sync-api`, `auth`, `doctor`, `help`, `version`, `completion`, `agent-context`, `profile`, `feedback`, `which`. These either don't read data or cannot operate before auth is established.
- **Best-effort:** refresh failures print a one-line stderr warning and the requested command always proceeds. A flaky network never breaks read-only data exploration.
- **Provenance line:** `auto-refresh: cache=ok (1.2s, 47 rows)  api=ok (820ms, 12 rows)` on stderr in interactive mode. Suppressed under `--agent` / `--json` / `--compact` / `--quiet` / non-TTY stderr so agent and CI consumers see no chatter.
- **Three-tier opt-out:** `--no-refresh` flag (wins) → profile `no-refresh` value → `GRANOLA_NO_AUTO_REFRESH=1` env.
- **agent-context disclosure:** exposes the full contract (default, flag, env, profile field, surfaces, skip list) so introspecting agents discover it without scraping --help.

## Approach

Five units, four logical commits:

1. **U1:** Extract `runCacheSync(ctx)` from `newSyncCacheCmd.RunE` and add a new focused `runApiSync(ctx, flags)` helper for the auto-refresh path. Existing user-facing commands keep byte-identical output, JSON shape, and exit codes.
2. **U2:** New `internal/cli/autorefresh.go` centralizing skip-list ancestor walking, dual-path detection, dispatch, provenance line formatting, and suppression matrix.
3. **U3:** Wire `runAutoRefresh` into `PersistentPreRunE`. Add `--no-refresh` persistent flag. Extend `agent-context` JSON with the auto_refresh contract.
4. **U4:** Provenance line + quiet-mode suppression (implemented as part of U2's autorefresh.go; covered by U2's tests).
5. **U5:** SKILL.md + README Auto-Refresh sections; replace the stale "Run sync first" troubleshooting advisory; patches manifest entry.

## Testing

- 287 tests pass (`go test ./...` across 13 packages), up from 267 before — 20 new unit tests for skip-list walking, opt-out precedence, plan detection, provenance line format, duration rendering, error shortening, and TTY suppression.
- `go vet ./...` clean.
- `govulncheck ./...` reports no vulnerabilities.
- `python3 .github/scripts/verify-skill/verify_skill.py --dir library/productivity/granola/` passes (flag-names, flag-commands, positional-args, unknown-command).
- Manual smoke test against the live binary:
  - `granola-pp-cli --no-refresh --help` shows the new flag.
  - `granola-pp-cli meetings list --no-refresh --limit 2 --json` returns data without refresh side effects.
  - `granola-pp-cli agent-context --json | jq .auto_refresh` shows the full contract (flag, env, surfaces, skip list).
  - `granola-pp-cli sync` and `granola-pp-cli sync-api` produce identical stdout/stderr/exit codes to pre-change behavior.

## Post-Deploy Monitoring & Validation

This is a CLI behavior change with no remote-service impact, but worth validating:

- **First indicator something is wrong:** users reporting that interactive commands hang for several seconds. Auto-refresh shouldn't add measurable latency for desktop-cache users (~tens of ms), but `GRANOLA_API_KEY` users will see the api-sync latency (~1-3s) on every command. If reports come in, the opt-out flag is the immediate mitigation; the deferred minimum-interval TTL knob is the structural fix.
- **Watch for:** runaway syncs (cache sync running on every command in a tight loop, hitting the Keychain prompt repeatedly). Mitigation: `GRANOLA_NO_AUTO_REFRESH=1`.
- **Doctor stays the source of truth.** Users diagnosing freshness issues should run `granola-pp-cli doctor` — auto-refresh writes the same `SyncState` doctor reads.
- **Validation window:** 1 week post-merge. Watch for issues opened against this repo mentioning "stale data", "slow command", "keychain prompt".
- **Owner:** @mvanhorn (CLI maintainer).
- **Rollback trigger:** if multiple users report keychain-prompt loops or measurable per-command latency >2s on cache-only setups, revert this PR and ship the deferred minimum-interval TTL knob first.

## Scope boundaries (deferred)

- **MCP server auto-refresh** — same idea, different harness; track separately so the CLI ships first.
- **Minimum-interval TTL** (`GRANOLA_AUTO_REFRESH_MIN_INTERVAL=2s`) — add if a high-frequency caller emerges.
- **Desktop AppleScript deep refresh** — opt-in `--refresh=deep` to nudge Granola desktop to pull from server.
- **Concurrency file-lock** — SQLite WAL handles current observed load.

### DIFF
--- a/cli-skills/pp-granola/SKILL.md
+++ b/cli-skills/pp-granola/SKILL.md
@@ -278,6 +278,41 @@ Commands that read from the local store or the API wrap output in a provenance e
 
 Parse `.results` for data and `.meta.source` to know whether it's live or local. A human-readable `N results (live)` summary is printed to stderr only when stdout is a terminal — piped/agent consumers get pure JSON on stdout.
 
+## Auto-Refresh
+
+Every command auto-refreshes the local store as its first action. You do **not** need to run `granola-pp-cli sync` before `meetings list`, `panel get`, or any other read — the CLI handles that for you on every invocation.
+
+**Two auth surfaces refresh independently:**
+
+| Surface | What runs | When it fires |
+|---------|-----------|---------------|
+| Desktop encrypted cache | `sync` (cache → SQLite) | When `~/Library/Application Support/Granola/cache-v6.json.enc` (or pre-encryption `cache-v6.json`) is present |
+| Public REST API | `sync-api` (public-api.granola.ai → SQLite) | When `GRANOLA_API_KEY` is set or an access token is saved in the config file |
+
+When both are available, both refresh routines fire (cache first, then api). When neither is configured, auto-refresh is a silent no-op and your underlying command produces its own auth error.
+
+**Freshness ceiling.** Auto-refresh reads from Granola desktop's encrypted cache file; it does **not** poke the desktop app to refresh from Granola servers. The freshness ceiling is whatever Granola desktop has already pulled. If a meeting just ended and the desktop hasn't synced from servers yet, no CLI-side refresh will surface it. For latest-second-fresh data, open Granola desktop briefly before invoking the CLI.
+
+**Provenance line.** When stderr is a TTY and you are not in `--agent` / `--json` / `--compact` / `--quiet` mode, a one-liner like `auto-refresh: cache=ok (1.2s, 47 rows)  api=ok (820ms, 12 rows)` lands on stderr after the refresh. Agent and JSON consumers see no chatter on stdout.
+
+**Failures are non-fatal.** A refresh that fails prints `cache=failed: <short reason>` on stderr and the command proceeds against whatever data is already in the store. Run `granola-pp-cli doctor` to investigate persistent refresh failures.
+
+**Opt out** (precedence: flag wins over env):
+
+```bash
+# Single command:
+granola-pp-cli meetings list --no-refresh
+
+# For a shell session / CI job:
+export GRANOLA_NO_AUTO_REFRESH=1
+
+# Saved per-profile via the existing profile mechanism:
+granola-pp-cli profile save fast --no-refresh
+granola-pp-cli --profile fast meetings list
+```
+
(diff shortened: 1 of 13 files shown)


tmchow:
Hey @mvanhorn, this PR has unresolved review feedback and a low review score from over 6 hours back. Heads-up so it doesn't drift. No auto-close on maintainer PRs.

mvanhorn:
Thanks @greptileai - both addressed in 9b5209bb:

- P1 `sync_autorefresh.go:109` warned rows dropped: moved `totalSynced += res.Count` into the `res.Warn != nil` branch as well. A warning is partial-success, not a no-op; the previous code zeroed out the provenance row count whenever every resource warned, even when real rows landed in the store.
- P2 `autorefresh.go:160` misleading doc: rewrote the autoRefreshOptedOut comment to match the code. A bool flag can't distinguish 'explicitly set to false' from 'default', so the env var actually runs whenever `flags.noRefresh` is false; the doc now also notes that env beats a profile-set `--no-refresh=false`.

PR #586, opened 2026-05-15
fix(openart): surface experimental flag and document nano-banana family

mvanhorn:
## openart

Stacked on PR #554. Three small patches caught while using the CLI live: agents and humans had no way to tell which image models need `--accept-experimental` until the submit was rejected, and the docs never mentioned that `nano-banana-pro` exists.

### What changed

- `internal/cli/models_novel.go` — Add `experimental` to `modelRow()` next to `recommended`, so `models list` and `models show` JSON expose the catalog's Experimental flag. Without this, the field was set correctly on every catalog entry but silently dropped from output.
- `internal/cli/models_novel_test.go` — New test file with three table-driven cases: verified model (nano-banana) returns experimental=false, experimental model (nano-banana-pro) returns experimental=true, every catalog entry exposes the key regardless of value.
- `internal/cli/image.go` — Sharpen `image gen --help` Long+Example block to point at `models list --family image` for discovery and show both Nano Banana variants in examples.
- `SKILL.md` + `README.md` — Enumerate the Nano Banana family (regular 50cr verified, Pro 120cr experimental), state explicitly that there is no `nano-banana-2` (Pro is the upgrade path), show both example commands side by side.
- `.printing-press-patches.json` — Catalog the customization (id `modelrow-surface-experimental`).
- `cli-skills/pp-openart/SKILL.md` — Regenerated mirror via `go run ./tools/generate-skills/main.go`.

### Validation

- `go test ./...` — 193 passed in 12 packages (3 new modelRow tests)
- `go vet ./...` — clean
- `python3 .github/scripts/verify-skill/verify_skill.py --dir library/ai/openart/` — clean
- `openart-pp-cli models show nano-banana-pro --agent` returns `experimental: true`
- `openart-pp-cli models show nano-banana --agent` returns `experimental: false`
- `openart-pp-cli image gen --help` shows both Nano Banana variants in examples

### DIFF
--- a/cli-skills/pp-openart/SKILL.md
+++ b/cli-skills/pp-openart/SKILL.md
@@ -42,10 +42,14 @@ These capabilities aren't available in any other tool for this API.
   openart-pp-cli video gen --prompt "..." --model byte-plus-seedance-2 --duration 10 --count 2 --wait --download ./out/
   ```
 
-- **`image gen`** — Submit + poll + download images in one command. Nano Banana is verified; other image models (GPT Image 2, FLUX 2 Pro, Seedream, Imagen 4, Qwen Image) require `--accept-experimental` because their submit shapes are inferred from the JS bundle but not individually verified live.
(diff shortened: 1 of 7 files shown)


mvanhorn:
Folding into #554 instead since that PR is still open.

PR #643, opened 2026-05-17
feat(instacart): named address profiles + per-call --profile flag

omarshahine:
## Summary

Add a named-address profile store to instacart so users with multiple delivery addresses (home, work, vacation house, ...) can switch between them without re-running `config set-address` every time.

The single-address shape that today's `config set-coords` / `set-address` ships with means switching addresses is destructive — every switch overwrites the previous one. For users with several active addresses on their Instacart account (and the `CurrentUserAddresses` GraphQL query that fix-instacart-location-config-546 already added returns them all) this is just a missing layer of indirection.

## What this changes

- New subcommand tree: `instacart config profiles list | show | add | use | rm | import`
- New persistent root flag: `--profile <name>` — applies a named profile for a single call without changing the active profile (e.g., `instacart --profile work add safeway "cold brew"`)
- `internal/config` gains a `Profile` struct, a `Profiles` map, and an `ActiveProfile` key. Helpers do not auto-save; callers persist via `Save()`.
- `profiles import` reuses the existing `currentUserAddressesQuery` + `client.Mutation()` path (introduced for tryAutoPopulateLocation) to seed one profile per saved Instacart address, slugifying the street line for the name.

## Backwards compatibility

When `profiles` and `active_profile` are absent from `config.json`, behavior is identical to today — the top-level `postal_code` / `address_id` / `latitude` / `longitude` keys remain authoritative. Activating a profile copies its values onto those same top-level keys so every downstream `gql.NewClient(...)` consumer continues to work unchanged. Round-trip tests pin that legacy configs do not gain the new keys.

## Implementation notes

- All hand-edits carry `// PATCH (instacart-address-profiles): ...` comments and are catalogued in `.printing-press-patches.json` (id `instacart-address-profiles`).
- `newAppContext` reads the `--profile` flag after loading config and calls `ApplyProfile` in-memory; no persistence side-effect.
- `fetchAddressByID` and `fetchUserAddresses` live in `internal/cli/profiles.go` so the existing `internal/cli/config.go` surface doesn't widen further.
- `slugifyName` caps at 40 chars to match `ValidProfileName`'s regex; collisions in `import` get a `-2`, `-3` suffix.

## Validation Results

| Check | Status |
|---|---|
| `go build ./...` | PASS |
| `go vet ./...` | PASS |
| `go test ./...` | PASS — 5 new tests in `internal/config/profiles_test.go`, 9 new tests in `internal/cli/profiles_test.go` |
| `govulncheck ./...` | PASS — no vulnerabilities affecting reachable code |
| `.github/scripts/verify-skill/verify_skill.py --dir library/commerce/instacart/` | PASS |
| Live smoke (auth → `config profiles add home --id 73256642 --use` → `add safeway "cold brew" --dry-run`) | PASS — `resolved_via: live`, retailers shop reachable |

## Test plan

```bash
cd library/commerce/instacart
go test ./...
python3 ../../../.github/scripts/verify-skill/verify_skill.py --dir .
go install ./cmd/instacart-pp-cli

# functional smoke (after `instacart auth login` or `auth import-file`)
instacart config profiles import
instacart config profiles list
instacart config profiles use <name>
instacart --profile <other-name> retailers list
```

### DIFF
--- a/cli-skills/pp-instacart/SKILL.md
+++ b/cli-skills/pp-instacart/SKILL.md
@@ -199,6 +199,22 @@ The post-`auth login` step auto-populates `address_id`, `postal_code`, `latitude
 
 `instacart doctor` surfaces a `location: fail` check whenever this is missing, so an agent driving the CLI can detect the broken state before invoking a real command.
 
+### Multiple addresses (named profiles)
+
+`config profiles` is a named-address store on top of the single active-location config. Use it when the user has more than one delivery address (home, work, vacation house) and wants to switch without re-running `config set-address` each time.
+
+- `instacart config profiles list` — show saved profiles; the active one is marked with `*`.
+- `instacart config profiles add <name> --id <address_id> [--label "..."] [--use]` — save a profile by Instacart address ID (uses `GetAddressById` to fill coords). Pass `--use` to also activate it.
+- `instacart config profiles add <name> --lat <N> --lon <N> [--postal <zip>] [--label "..."]` — save a profile by raw coordinates, no network call.
+- `instacart config profiles use <name>` — switch the active profile (copies its location onto the top-level config keys so every downstream call uses it).
+- `instacart config profiles show <name>` — print one profile.
+- `instacart config profiles rm <name>` — delete a profile. If it was active, the active profile is cleared and the existing top-level config still applies.
+- `instacart config profiles import [--prefix <p>] [--overwrite] [--use <name>]` — fetch every saved address from the user's Instacart account (via `CurrentUserAddresses`) and save each as a profile, slugifying the street address for the name.
+
+Per-call override: pass `--profile <name>` to any command that needs location (e.g. `instacart --profile work add safeway "cold brew"`). This applies the named profile for that single call without changing the active profile.
+
+When no profiles are defined, the CLI behaves exactly as before — `config set-coords` / `set-address` / `show` continue to drive the top-level location keys directly.
+
 ## Agent Mode
 
 The CLI is agent-native by default. Pass `--json` on any command for machine-readable output. `--dry-run` previews `add` without firing the mutation and surfaces which resolver (`history`, `live`, or `item-id`) would have fired.
--- a/library/commerce/instacart/.printing-press-patches.json
+++ b/library/commerce/instacart/.printing-press-patches.json
@@ -26,6 +26,21 @@
         "internal/cli/doctor.go",
         "internal/cli/root.go"
       ]
+    },
+    {
+      "id": "instacart-address-profiles",
+      "summary": "Add a named-address profile store under `instacart config profiles` (list / show / add / use / rm / import) plus a `--profile <name>` per-call persistent flag so users with multiple delivery addresses can switch without re-running `config set-address` each time.",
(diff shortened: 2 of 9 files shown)


omarshahine:
Ran a Codex review pass on the original commit; all three P2 findings were valid and now fixed in `f93dc77c`:

1. **`profiles import` idempotency** — previously the suffix loop ran before the existing-name skip check could fire, so a re-run without `--overwrite` invented `<name>-2` instead of reporting the entry as skipped. New `resolveImportName` distinguishes "same address, different name" (skip / update) from "same name, different address" (suffix). Test: `TestResolveImportNameIdempotentReimport` + `TestResolveImportNameCollidesWithDifferentAddress`.

2. **`--prefix` overflowing the 40-char `ValidProfileName` cap** — `slugifyName` only truncated the street portion, so a long prefix + long street could abort the entire import at `SetProfile`. Import now truncates the prefix+slug combination together, and the suffix loop keeps variants under the cap too. Test: `TestResolveImportNameLongPrefixStillFits`.

3. **Stale `ZoneID` on profile switch** — `ApplyProfile` previously only copied `ZoneID` when non-empty, so switching from a profile with a custom zone to one without it left the old zone in place. Now copies verbatim; `EffectiveZoneID` already handles the empty case with a fallback to "38". Test: `TestApplyProfileClearsStaleZoneID`.

Preflight still green: `go build`, `go vet`, `go test ./...` (3 new tests added), `govulncheck`, `verify_skill.py`.

omarshahine:
Pushed `8a2c1285` addressing all three open items:

**CI fix — `Guard against hand-edits to cli-skills mirror`** was failing because commit `827f0c1d` hand-regenerated `cli-skills/pp-instacart/SKILL.md`. Restored that file to upstream/main state; the post-merge `generate-skills` bot will mirror the library SKILL.md change.

**Greptile P2 #1 — `fetchAddressByID` redundantly reloads config + session.** Refactored to accept `sess *auth.Session, cfg *config.Config` as params, matching `fetchUserAddresses`. Caller in `newConfigProfilesAddCmd.RunE` now passes its already-loaded values. Two redundant disk reads gone; helper shape is symmetric.

**Greptile P2 #2 — package-level `fetchUserAddressesFn` is parallel-test-fragile.** Replaced with a typed `userAddressFetcher` plus a context-key lookup (`userAddressFetcherFrom`). Tests inject via `context.WithValue` through a new `runCmdCtx` helper — no global mutation, parallel-safe. Both import tests use the new pattern.

Preflight green: `go build`, `go vet`, `go test ./...`, `govulncheck`, `verify_skill.py`.

omarshahine:
**Force-pushed history rewrite.** The Verify gate (Guard-against-hand-edits-to-cli-skills-mirror) scans every commit that *touches* `cli-skills/pp-*/SKILL.md` regardless of net diff. Two of the fix commits (`827f0c1d` adding the mirror section, `8a2c1285` reverting it) both showed up as offenders even though the branch net-state matches main on that file.

Squashed the 6 fix commits into one `a236aaec`. The squashed commit doesn't touch `cli-skills/` at all (the toggles canceled). Branch is now:

- `8f859d10` — feat: named address profiles + per-call `--profile` flag (unchanged)
- `a236aaec` — fix: review fixes for profiles patch (squashed)

Full review history is preserved in the PR comment thread; the squashed commit message catalogs each fix with attribution to the review round that surfaced it.

Preflight still green locally: `go build`, `go vet`, `go test ./...`, `govulncheck`, `verify_skill.py`.

PR #645, opened 2026-05-17
fix(redfin): stop bogus homes auto-refresh; filter price/type client-side

jwmoss:
## Problem

`redfin-pp-cli homes` had three issues that made filtered search workflows unreliable:

1. `homes` was registered as an auto-refreshable read path, so every call tried to generic-sync `/stingray/api/gis` without required Stingray params like `al`, `region_id`, and `region_type`. That produced noisy `sync_start` / `sync_error` / stale-cache warnings before the real search ran.
2. Stingray accepts range filters such as `min_price`, `max_price`, `min_beds`, `min_baths`, `min_sqft`, `max_sqft`, `min_year_built`, `max_year_built`, and `min_lot_size`, but observed responses do not consistently honor them server-side. The website filters those rows client-side; the CLI was returning them.
3. Stingray also accepts `uipt` for property type, but observed responses can still include rows tagged with other `uiPropertyType` values. The CLI could not enforce `--type` because the gis parser discarded `uiPropertyType`.

## Fix

- Remove `homes` from the auto-refresh registry. `homes` is a per-call search endpoint, not a list-all resource that can be safely refreshed without the user's region and filters.
- Add `filterListings` after the Stingray response parse to enforce price, beds, baths, sqft, year-built, lot-size, and property-type filters locally.
- Parse `lotSize` and `uiPropertyType` from `/api/gis` rows and surface them on `Listing` so `--lot-min` and `--type` can be enforced.
- Tighten the zero-field behavior: price and lot bounds now require known values when those filters are set; sparse non-price fields are still preserved where zero means unknown.
- Fix the generated `RetryAfter` helper so future Unix epoch seconds/milliseconds are parsed before numeric values are treated as delta seconds. This clears the existing full-package test failure.
- Regenerate the `cli-skills/pp-redfin` mirror in a bot-authored commit for the SKILL freshness-contract update.

## Validation

| Check | Result |
|---|---|
| `go test ./internal/cli/... ./internal/redfin/...` | PASS |
| `go test ./...` | PASS |
| `go vet ./...` | PASS |
| `python3 .github/scripts/verify-skill/verify_skill.py --dir library/other/redfin` | PASS |
| `go run ./tools/generate-skills/main.go` | PASS, generated mirror updated |
| `git diff --check` | PASS |

Live smoke, price/type filter:

```bash
go run ./cmd/redfin-pp-cli homes --region-slug "city/30818/TX/Austin" \
  --price-min 700000 --price-max 900000 --beds-min 3 --baths-min 2.5 \
  --type house --sort price-asc --limit 350 --json --no-cache --rate-limit 1 \
  | jq '{count: length, missing_price: (map(select(.price == null or .price == 0)) | length), non_house: (map(select(.ui_property_type != 1 and .ui_property_type != null)) | length), under_price: (map(select(.price != null and .price != 0 and .price < 700000)) | length), over_price: (map(select(.price != null and .price > 900000)) | length)}'
```

Result:

```json
{
  "count": 30,
  "missing_price": 0,
  "non_house": 0,
  "under_price": 0,
  "over_price": 0
}
```

Live smoke, lot-size filter:

```bash
go run ./cmd/redfin-pp-cli homes --region-slug "city/30818/TX/Austin" \
  --lot-min 10000 --limit 350 --json --no-cache --rate-limit 1 \
  | jq '{count: length, missing_lot: (map(select(.lot_size == null or .lot_size == 0)) | length), under_lot: (map(select(.lot_size != null and .lot_size != 0 and .lot_size < 10000)) | length)}'
```

Result:

```json
{
  "count": 93,
  "missing_lot": 0,
  "under_lot": 0
}
```

## Notes

This is a published-library patch per `AGENTS.md`: fix the shipped CLI first, record the local customizations in `.printing-press-patches.json`, and leave generator-shaped follow-up work for `cli-printing-press`.


tmchow:

tmchow:
@jwmoss fix CI

jwmoss:
> @jwmoss fix CI

Done

PR #663, opened 2026-05-18
fix(registry): fall back to manifest description, name offending entry on parse error

### LINKED ISSUE #661: Catalog entry `lawhub` has empty description, breaks `printing-press list`
## Repro

```
$ npx -y @mvanhorn/printing-press list
registry entry missing string field: description
```

The command exits with no entries listed. The error is repeatable on a fresh `npx` invocation.

## Cause

`registry.json` contains one entry whose `description` field is the empty string:

```
$ curl -s https://raw.githubusercontent.com/mvanhorn/printing-press-library/main/registry.json \
    | python3 -c "import json,sys; d=json.load(sys.stdin); \
        bad=[e for e in d['entries'] if not e.get('description')]; \
        print(len(bad), bad)"
1 [{'name': 'lawhub', 'category': 'legal', 'description': '', ... 'path': 'library/education/lawhub', ...}]
```

The `lawhub` entry under `library/education/lawhub` has `description: ""`. Every other entry (128 of 129 as of `schema_version: 2` at HEAD) has a non-empty description.

The CLI's parser treats `description` as a required non-empty string field and aborts on the first violation — so a single bad row stops the entire `list` output. Users see "registry entry missing string field: description" with no clue which entry caused it or how many other entries are intact.

## Fix

One-line fix is to populate `lawhub`'s description with a real summary of what the CLI does (similar shape to peer entries — see `linear`'s description for a model).

If empty descriptions should remain valid (e.g., during a work-in-progress catalog PR), the CLI's parser could either skip the offending entry and continue OR include the entry name in the error message so the failing row is identifiable from the CLI output alone.

I'd lean fix-the-entry, since the empty string is more likely an oversight than intent. The "name the offending entry in the error" change is worth keeping in mind regardless — surfacing which row failed makes future occurrences self-diagnosing.

## Surrounding context

Discovered while debugging an Aileron install path that fetches the catalog to drive `aileron pp add <name>`. Aileron's installer reads the JSON directly and doesn't trip on this (every other entry parses), but the `printing-press list` UX is the natural way for a user to browse the catalog before installing — having it return zero entries because of one stale row is a sharp paper cut.

ALRubinger:
## Summary

- Extend the registry generator's description sourcing chain to fall back to `.printing-press.json`'s `description` field when both the existing `registry.json` value and the `.goreleaser.yaml` brews description are empty. This fixes the `lawhub` entry that ships with `description: ""` because lawhub is the only CLI in the catalog without a `brews:` block in its `.goreleaser.yaml`, and its manifest carries a perfectly good description that the generator simply wasn't reading.
- Surface the offending entry name in the npm parser's "missing string field" errors. With 129 entries in the live registry, an anonymous `registry entry missing string field: description` was a process-of-elimination diagnosis; `registry entry 'lawhub' missing string field: description` is one line of output.

Fixes #661.

## Why two changes

The reporter recommended fix-the-entry as the primary path and called out the error-naming improvement as defense in depth ("worth keeping in mind regardless — surfacing which row failed makes future occurrences self-diagnosing"). Both are small, isolated, and address different layers of the same paper cut.

## What does NOT change

- `registry.json` and `cli-skills/**` are bot-regenerated post-merge; not in the PR diff (the `Fail on changes to generated artifacts` step in `verify-library-conventions.yml` would hard-fail otherwise). The `generate-registry.yml` workflow will pick up the new generator logic on the merge commit and populate `lawhub`'s description automatically.
- No source-of-truth edits to lawhub's manifest — its description was already there.

## Test plan

- [x] `go test ./tools/generate-registry/...` — new `TestRegistryDescriptionManifestFallback` covers the third-tier fallback, brews-over-manifest precedence, and all-empty case.
- [x] `npm test` — new cases pin both the named-entry error format and the anonymous-fallback when `name` itself is missing. 50/50 pass.
- [x] Ran `GO111MODULE=off go run ./tools/generate-registry/main.go --print` locally and confirmed lawhub's description now populates from its manifest, no other entries change.


tmchow:

tmchow:
Hey @ALRubinger, this PR has merge conflicts. Mind taking another pass? If you've hit a blocker or want to hand it off, just leave a note. Otherwise I'll close it for inactivity in about 2 days, happy to reopen anytime.

tmchow:

Issue #400, opened 2026-05-10
install <name> always fails skill step on Windows: 'Skill install failed' (subprocess npx loses TTY)


roiizchak:
## Summary

`@mvanhorn/printing-press install <name>` consistently reports `Skill install failed for pp-<name>. The binary remains installed at <path>.` on Windows PowerShell. The Go binary installs successfully every time; only the skill step fails. Running the same `npx -y skills@latest add ...` command directly from the same shell **succeeds**.

## Repro

```powershell
PS> npx -y "@mvanhorn/printing-press" install notion
Skill install failed for pp-notion. The binary remains installed at C:\Users\royko\go\bin\notion-pp-cli.exe.
```

Then running the underlying skill install command directly:

```powershell
PS> npx -y skills@latest add mvanhorn/printing-press-library/cli-skills/pp-notion -g -y
# ...interactive UI runs, summary panel, "Installed 1 skill" — exit 0
```

The skill installs correctly when invoked directly, confirming the binary, registry, and skill source are all valid.

## Root Cause (Hypothesis)

`dist/src/commands/install.js:85` calls `installSkill()`, which in `dist/src/skill.js:14-16` does:

```js
return runner("npx", skillsAddArgs(skillName, options));
```

`execFileRunner` spawns `npx` as a piped subprocess (no TTY). The `skills` CLI (v1.5.6) is interactive — clack-style prompts, spinners, summary panels. Without a TTY on stdin/stdout, it appears to exit non-zero, even though the skill files are written. The wrapper sees the non-zero exit and reports `Skill install failed`.

Notably, the binary remains installed because `goInstall` ran first and succeeded.

## Environment

- OS: Windows 11 Pro 26200, PowerShell 7
- `@mvanhorn/printing-press`: 0.1.3
- `printing-press` Go binary: 4.2.2 (Go 1.26.3)
- `skills` CLI: 1.5.6
- Node: v24.15.0

## Workaround

Two-step install:

```powershell
npx -y "@mvanhorn/printing-press" install <name> --cli-only
npx -y skills@latest add mvanhorn/printing-press-library/cli-skills/pp-<name> -g -y
```

The second command runs interactively in the user's shell, gets a TTY, and succeeds.

## Suggested Fixes

Pick whichever fits the wrapper's design:

1. **Pass `--no-interactive` (or equivalent)** to the `skills add` invocation if the CLI supports it — keeps subprocess execution but stops it requiring a TTY.
[...]

PR #664, opened 2026-05-18
fix(npm): skill install/uninstall fails on Windows (execFile npx → ENOENT)

magoo242:
## Summary

On Windows, `printing-press install <name>` installs the Go binary but **always fails the skill step** ("Skill install failed"); `uninstall <name>` breaks the same way. This fixes both. Refs #400 — thanks @roiizchak for the clear report and repro.

## Root cause (empirically verified — Windows 11, Node v24.14.1)

`npm/src/skill.ts` calls `runner("npx", …)` → `npm/src/process.ts` `execFileRunner` → Node `execFile("npx", …)` **with no `shell` option**. On Windows `npx` ships as `npx.cmd`/`npx.ps1` — there is no `npx.exe`. `execFile` without a shell uses `CreateProcess`, which only auto-appends `.exe`, never `.cmd`. So **npx is never spawned → `ENOENT` (errno -4058) → exit 127 → "Skill install failed."** The `skills` CLI never runs. This isn't a TTY/stdio issue (the original #400 hypothesis was a reasonable first guess) — the subprocess never starts.

**The go-vs-npx asymmetry:** `npm/src/go.ts` uses the *same* `execFileRunner` via `runner("go", …)` and *works*, purely because `go.exe` exists and `CreateProcess` auto-appends `.exe`. Same runner, opposite outcome, solely `.exe` vs `.cmd`.

| Isolated repro (Windows 11, Node v24.14.1) | Result |
|---|---|
| `execFile("npx", ["--version"])` bare | `spawn npx ENOENT` (errno -4058) |
| `execFile("npx", ["--version"], { shell: true })` | ✅ `11.11.0` |
| `execFile("npx.cmd", ["--version"])` bare | `spawn EINVAL` — "just use `npx.cmd`" is **not** valid on Node ≥20 Windows |
| `execFile("go", ["version"])` bare | ✅ (control: `.exe` resolves) |

## Scope

Two call sites, both in `skill.ts` (`installSkill`, `removeSkill`) → fixes **install and uninstall**. `go.ts` unaffected and stays unaffected.

## The fix

- `npm/src/process.ts`: opt-in `shell?: boolean` on `RunOptions`, passed through in `execFileRunner`, **default `false`** so every existing caller (notably `go.ts`) is byte-for-byte unchanged.
- `npm/src/skill.ts`: opt in at both npx sites with `{ shell: process.platform === "win32" }` (Windows only; POSIX unchanged).

+4/−2 across 2 files; `go.ts` untouched.

## Verification (Windows 11, Node v24.14.1)

End-to-end through the real `bin/printing-press.mjs` → `cli` → `installSkill`/`removeSkill` → patched `execFileRunner`:

- `install <cli> --skill-only` → **exit 0**, skill installed (confirmed in agent skill list) — the exact step that fails today.
- `uninstall <cli> --yes` → **exit 0**, skill removed (`removeSkill` path).
- Control: patched npx (`shell:true`) → exit 0; `go` path (default options, unchanged) → exit 0.
- `npm test`: 44 pass / 4 fail; the 4 failures pre-exist on `main` (unrelated `install`-command tests asserting POSIX paths that fail on any Windows checkout). **Zero** patch-induced regressions — verified by re-running the suite on `main` with the patch reverted (identical 4 failures).

*(End-to-end used a local registry fixture because a live default-registry install is currently blocked by an unrelated malformed `registry.json` entry — being addressed separately in #663. Registry resolution is orthogonal to the npx-spawn path under test.)*

## Honest caveat (your call)

`shell: true` triggers Node `DEP0190` (args concatenated, not escaped). Safe here: args are fixed flags + a library slug path, no untrusted shell metacharacters. The heavier-but-fully-robust alternative is a `cross-spawn` dependency. I chose the zero-dependency `shell` opt-in for a minimal diff — happy to switch to `cross-spawn` if you'd prefer.

One open question: this fix won't reach `npx @mvanhorn/printing-press` users until `npm/package.json`'s `version` is bumped (per AGENTS.md). Glad to add a patch bump here, or leave version control to you.


tmchow:

tmchow:
@magoo242 address code review feedback

tmchow:
Hey @magoo242, this PR has unresolved review feedback. Mind taking another pass? If you've hit a blocker or want to hand it off, just leave a note. Otherwise I'll close it for inactivity in about 2 days, happy to reopen anytime.

magoo242:
Addressed the review feedback by switching to `cross-spawn` — Greptile's preferred remedy, and the option I'd flagged in the original PR body. Pushed as `16a931ff`.

**What changed**
- `npm/src/process.ts`: replaced `execFile` with `cross-spawn.spawn`, collecting stdout/stderr from the child streams. No more `shell: true` anywhere — cross-spawn handles `.cmd`/`.bat` resolution on Windows itself. ENOENT continues to map to exit 127.
- `npm/src/skill.ts`: reverted the per-platform `{ shell: process.platform === "win32" }` opt-in at both `installSkill` and `removeSkill` — no longer needed.
- `RunOptions.shell?: boolean` removed.
- `npm/package.json`: `cross-spawn ^7.0.6` runtime dep, `@types/cross-spawn ^6.0.6` dev dep.

The `Runner` signature is unchanged, so `go.ts` and POSIX behavior are byte-identical.

**Verification on Windows 11 / Node v24.14.1** (same end-to-end harness as the original PR):
- `install --skill-only` → exit 0; skill genuinely installs (confirmed present in the agent skill list afterward).
- `uninstall --yes` → exit 0; skill removed (`removeSkill` path).
- Parent-process stderr captured during a real `npx` invocation contains **no DEP0190** (vs. visible on every run with the prior approach).
- Control: `execFileRunner("go", ["version"])` → exit 0 with no shell, confirming the go-install path is unaffected.
- `npm test`: 44 pass / 4 fail — the same 4 pre-existing Windows POSIX-path failures in `install.test.js` that are present on `main` and are untouched by this change. No patch-induced regressions.

**Side benefit:** because args remain an array and are never shell-concatenated, the latent argument-injection surface that came with `shell: true` is also closed.

Happy to address anything else. The cross-spawn alternative also resolves the open `shell: true`-vs-`cross-spawn` question from the PR body in cross-spawn's favor — if there are no concerns with the dep, I think this is mergeable.

tmchow:
Hey @magoo242, this PR has unresolved review feedback. Mind taking another pass? If you've hit a blocker or want to hand it off, just leave a note. Otherwise I'll close it for inactivity in about 2 days, happy to reopen anytime.

tmchow:
Closing this for now after a couple quiet days. Feel free to reopen anytime you want to keep going.

(This is an automated triage message. If you think it was closed in error, just reopen it or leave a note here and a maintainer will take a look.)

magoo242:
For anyone landing here: opened #864 as a fresh PR since the API wouldn't reopen this one after the bot-close. Same branch, same fix (now caught up to current main, including the package rename). Continuing the discussion there.

PR #665, opened 2026-05-18
feat(ci): add supply-chain hardening (scan workflow + Greptile rules)

tmchow:
## Summary

Adds layered PR-time defenses against the supply-chain attack shapes observed in four real-world incidents (March–May 2026): Axios, TanStack mini-Shai-Hulud, node-ipc, and BufferZoneCorp. Cross-repo mirror lands in [mvanhorn/cli-printing-press](https://github.com/mvanhorn/cli-printing-press) as mvanhorn/cli-printing-press/pull/1619

## What this PR does

Two layers, applied to every PR touching `.github/workflows/**`, `library/**/go.mod`, or `npm/package.json`:

1. **`verify-supply-chain.yml`** — deterministic Python scan at `.github/scripts/verify-supply-chain/scan.py` (with `signals.py` pattern catalog and `scan_test.py` covering 41 unit + integration tests). Hard-fails on block-tier signals:
   - `pull_request_target` workflow that checks out PR head ref (TanStack OIDC theft)
   - `id-token: write` outside `npm-publish.yml` allowlist
   - New `replace` directive in `library/**/go.mod` pointing at a remote URL (BufferZoneCorp)
   - `GOPROXY` / `GOFLAGS` / `GONOSUMCHECK` overrides in workflow env
   - `postinstall` / `preinstall` / `prepare` added to `npm/package.json` (Axios)
   - `module` directive drift on an existing CLI's `go.mod` (catalog variant of BufferZoneCorp)
   - Local-path `replace` directives emit **advisory** notices only — `library/food-and-dining/ordertogo/go.mod` has three legitimate ones; the diff-aware scan only flags newly-added ones.

2. **Seven new `greptile.json` rules** covering the same signals plus a judgment-only rule for Go-source credential-path reads in `library/**/internal/**/*.go` and `library/**/cmd/**/*.go` (catches the node-ipc shape when a CLI's purpose doesn't explain the read).

Plan: [docs/plans/2026-05-17-001-feat-supply-chain-hardening-plan.html](docs/plans/2026-05-17-001-feat-supply-chain-hardening-plan.html)
Rationale & primary sources: [docs/solutions/security/2026-05-supply-chain-hardening.md](docs/solutions/security/2026-05-supply-chain-hardening.md)
AGENTS.md addition under "Supply-chain hardening" (after the existing Greptile section).

## Current state audit

Performed against fresh `origin/main` as of 2026-05-17:

- ✅ Only `pull_request_target` workflow in repo (`greptile-policy-gate.yml`) is safe — API calls only, no checkout of PR head, no `id-token`.
- ✅ `id-token: write` grants only in `npm-publish.yml` (the OIDC Trusted Publishing allowlist).
- ✅ No `GOPROXY` / `GOFLAGS` / `GONOSUMCHECK` overrides in any workflow.
- ✅ Existing `replace` directives in `library/food-and-dining/ordertogo/go.mod` are all local vendoring (`./third_party/...`) — these are advisory, not blocking, and the scan is diff-aware so they don't trip on unrelated PRs.
- ✅ Every spot-checked `library/**/go.mod` uses canonical `module github.com/mvanhorn/printing-press-library/library/<cat>/<slug>`.

All new rules are forward-looking gates; no existing code needs to change for them to clear.

## Rollout posture

This workflow runs **informationally** on landing — see plan KD7. After a one-week green window with no false-positive miscalibration on real PRs, promote to a required check via branch protection.

## Test plan

- [x] 41 unit + integration tests pass: `cd .github/scripts/verify-supply-chain && python3 -m unittest scan_test`
- [x] Scan against `origin/main` produces zero findings (no regression on existing tree)
- [x] All four motivating attack shapes are covered by test fixtures (TanStack, Axios, BufferZoneCorp remote+local, node-ipc-class credential reads)
- [ ] Sandbox PR with each attack shape produces the expected red workflow + Greptile P0 — to be exercised in the one-week informational window before promoting to required check.

## Linked work

- Cross-repo mirror: PR B at [mvanhorn/cli-printing-press#1619](https://github.com/mvanhorn/cli-printing-press/pull/1619).
- Primary sources cited in the solutions doc: Datadog ([Axios](https://securitylabs.datadoghq.com/articles/axios-npm-supply-chain-compromise/)), Socket.dev ([TanStack](https://socket.dev/blog/tanstack-npm-packages-compromised-mini-shai-hulud-supply-chain-attack)), Microsoft ([Axios mitigation](https://www.microsoft.com/en-us/security/blog/2026/04/01/mitigating-the-axios-npm-supply-chain-compromise/)), The Hacker News ([BufferZoneCorp](https://thehackernews.com/2026/05/poisoned-ruby-gems-and-go-modules.html), [TanStack at OpenAI](https://thehackernews.com/2026/05/tanstack-supply-chain-attack-hits-two.html)), Security Arsenal ([BufferZoneCorp deep dive](https://securityarsenal.com/blog/bufferzonecorp-supply-chain-attack-poisoned-ruby-and-go-modules-targeting-ci-pipelines)).

### DIFF
--- a/.github/scripts/verify-supply-chain/scan.py
+++ b/.github/scripts/verify-supply-chain/scan.py
@@ -48,8 +48,14 @@ def git_show(ref: str, path: str) -> str | None:
     return result.stdout if result.returncode == 0 else None
 
 
-def changed_files(base_ref: str, head_ref: str) -> list[tuple[str, str]]:
-    """Return [(status, path)] for files touched between base and head."""
+def changed_files(base_ref: str, head_ref: str) -> list[tuple[str, str, str | None]]:
+    """Return [(status, path, old_path)] for files touched between base and head.
+
+    old_path is non-None only for renames and copies (status R/C); the signal
+    layer uses it to fetch the file's pre-rename content from base, so that
+    e.g. R6 module-path drift can detect renames whose new module string also
+    starts with the canonical prefix.
+    """
     result = run_git(["diff", "--name-status", "-z", f"{base_ref}...{head_ref}"])
     if result.returncode != 0:
         print(result.stderr, file=sys.stderr)
@@ -59,7 +65,7 @@ def changed_files(base_ref: str, head_ref: str) -> list[tuple[str, str]]:
     if fields and fields[-1] == "":
         fields.pop()
 
-    entries: list[tuple[str, str]] = []
+    entries: list[tuple[str, str, str | None]] = []
     i = 0
     while i < len(fields):
         status = fields[i]
@@ -67,15 +73,14 @@ def changed_files(base_ref: str, head_ref: str) -> list[tuple[str, str]]:
         if not status:
             continue
         if status.startswith(("R", "C")):
-            # rename or copy: status, old_path, new_path
-            _ = fields[i]  # old path
+            old_path = fields[i]
             new_path = fields[i + 1]
             i += 2
-            entries.append((status[0], new_path))
+            entries.append((status[0], new_path, old_path))
         else:
             path = fields[i]
             i += 1
-            entries.append((status, path))
+            entries.append((status, path, None))
     return entries
 
 
@@ -151,8 +156,14 @@ def emit_annotation(f: signals.Finding) -> None:
 # ---------------------------------------------------------------------------
 
 
-def build_change(base_ref: str, head_ref: str, status: str, path: str) -> signals.FileChange:
-    base_content = None if status == "A" else git_show(base_ref, path)
+def build_change(
+    base_ref: str, head_ref: str, status: str, path: str, old_path: str | None
+) -> signals.FileChange:
+    # For renames/copies, the file's prior version lives at old_path on base.
+    # Treating it as a true rename (rather than "new file at new_path") lets
+    # signals like R6 compare base_module against head_module across the rename.
+    base_lookup_path = old_path if old_path else path
+    base_content = None if status == "A" else git_show(base_ref, base_lookup_path)
     head_content = None if status == "D" else git_show(head_ref, path)
     diff_added = added_lines(base_ref, head_ref, path) if head_content is not None else []
     return signals.FileChange(
@@ -165,10 +176,10 @@ def build_change(base_ref: str, head_ref: str, status: str, path: str) -> signal
 
 def scan(base_ref: str, head_ref: str, strict: bool) -> list[signals.Finding]:
     findings: list[signals.Finding] = []
-    for status, path in changed_files(base_ref, head_ref):
+    for status, path, old_path in changed_files(base_ref, head_ref):
         if not is_scoped(path):
             continue
-        change = build_change(base_ref, head_ref, status, path)
+        change = build_change(base_ref, head_ref, status, path, old_path)
         for finding in signals.run_signals(change):
             if strict and finding.severity == "advise":
                 finding = signals.Finding(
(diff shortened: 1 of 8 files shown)


tmchow:
Both findings addressed in [commit cf99c…](https://github.com/mvanhorn/printing-press-library/commit/HEAD) (will update with hash after Greptile re-runs):

**1. go.mod block-form `replace ( ... )` evasion** — fixed. `signal_gomod_replace` now parses block ranges from head_content via `_replace_block_ranges()` and matches body lines inside any block against `_REPLACE_BLOCK_BODY` (the keyword-less body form). Same remote-vs-local tiering applies. Four new tests cover: remote target inside block, local target inside block, three mixed entries in one block, and a guard that `=>` arrows outside any block don't false-positive on comments. 45 tests pass (was 41).

**2. Scanner self-bypass via head checkout** — fixed. Workflow now checks out the **base** branch from the **base** repo, then fetches the PR head SHA as a detached ref and passes it to `scan.py` via `--head-ref`. The scan.py executing the gate is the trusted base version; a PR can't weaken it to mask its own payload. This was forward-looking (currently informational) but addresses the concern before required-check promotion.

Thanks for both — the block-form one was a genuine evasion path that would have shipped silently. Greptile should re-review on the next push.

tmchow:
Addressed both findings from the second review pass:

**1. R1 flow-sequence trigger form** — fixed. Added `_PR_TARGET_TRIGGER_FLOW` to match `on: [pull_request_target, push]` and the reverse ordering. Unified via `_has_pr_target_trigger()`. Two new tests cover both orderings.

**2. R1/R2/R4 full-content scans** — fixed via a shared `_lines_to_scan()` helper:
   - new files (base absent) → all head lines are "new" and scanned (so a fresh workflow with an attack shape still trips)
   - existing files → only `added_lines` (so a pre-existing pattern unrelated to the PR can't be re-flagged)

R3 (gomod_replace), R5 (npm lifecycle), R6 (module-path) were already diff-aware via different paths. Five new tests cover the diff-aware behavior: pre-existing patterns unchanged → no fire; freshly-added patterns → still fire.

53 tests pass (was 45). Same fixes mirrored to [mvanhorn/cli-printing-press#1619](https://github.com/mvanhorn/cli-printing-press/pull/1619).

Greptile should re-review on the new HEAD.

tmchow:
@greptileai review

tmchow:
Both findings addressed:

**1. R6 within-canonical-prefix rename gap** — fixed. The blind spot was that a rename like `library/payments/kalshi` → `library/payments/kalshi-evil` (both under the canonical prefix) passed R6 silently. Now `scan.py`'s `changed_files()` also returns the pre-rename `old_path`, and `build_change()` fetches `base_content` from `old_path` for renames/copies. R6 sees the diff across the rename and flags ANY module-directive change on an existing CLI — outside-canonical drift keeps the `module_path_drift_on_existing_cli` signal, within-canonical rename gets the new `module_path_rename_on_existing_cli`. Greptile R6 rule prose updated to match. Two new tests (unit + end-to-end via `git mv`).

**2. `go.sum` + `npm/package-lock.json` in path filter** — fixed. Both removed from the trigger list. `is_scoped()` in `scan.py` routes only `.yml/.yaml`, `library/**/go.mod`, and `npm/package.json`, so the lockfiles only ever produced no-op runs on `go mod tidy` / `npm install` commits. No coverage lost.

55 tests pass (was 53).

tmchow:
@greptileai review

PR #666, opened 2026-05-18
fix(ci): bump greptile policy gate timeout 15min → 25min

tmchow:
Small follow-up to #665. Bumps the Greptile policy gate's polling deadline because the 15-minute budget left zero headroom when Greptile re-reviews queue behind rate limits (#665 hit this).

## Change

- Inner polling: `deadline=$((SECONDS + 900))` → `deadline=$((SECONDS + 1500))` (15 → 25 min)
- Job timeout: `timeout-minutes: 17` → `timeout-minutes: 27` (preserving the 2-min buffer for checkout/setup)

## What was dropped

The original version of this PR also added a step to skip the gate when the head commit was co-authored by `greptile-apps[bot]` (intent: avoid the gate timing out on "Fix in Codex/Claude" deeplink commits, where Greptile rarely re-reviews its own suggestions). Greptile correctly flagged that as P1: any contributor — including fork PRs on `pull_request_target` — can write `Co-authored-by: greptile-apps[bot]` in a commit message and bypass the gate. Forgeable bypass in a security gate is worse than the problem it solved. Dropped.

For the original problem (Greptile not re-reviewing deeplink commits), the existing escape hatches stay: admin force-merge, or `@greptileai review` to manually trigger.

## Test plan

- [x] YAML parses
- [ ] In-the-wild observation over the next week — does the 25-min budget absorb Greptile's queue variance? If still tight, revisit.

No behavioral change beyond the timeout values.

### DIFF
--- a/.github/workflows/greptile-policy-gate.yml
+++ b/.github/workflows/greptile-policy-gate.yml
@@ -61,29 +61,7 @@ jobs:
           echo "number=$(jq -r '.number' <<<"${pr_json}")" >> "${GITHUB_OUTPUT}"
           echo "head_sha=$(jq -r '.head.sha' <<<"${pr_json}")" >> "${GITHUB_OUTPUT}"
 
-      - name: Detect Greptile-authored fix commit
-        id: bot-check
-        # "Fix in Codex/Claude/Cursor/Conductor" deeplinks in Greptile reviews
-        # produce a commit co-authored by greptile-apps[bot]. The gate would
-        # otherwise poll for a Greptile review of that commit specifically,
-        # but Greptile rarely re-reviews its own suggestions in time, so the
-        # gate times out on a commit Greptile already implicitly endorsed.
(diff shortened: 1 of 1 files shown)


tmchow:
You're right — the bot-bypass was a P1 forgeable hole. Any contributor could write \`Co-authored-by: greptile-apps[bot]\` in their commit message (including fork PRs on \`pull_request_target\`) and skip the gate. Dropped it entirely.

Kept only the timeout bump (deadline 900→1500, job timeout 17→27). The remaining fix is the safe one: Greptile re-reviews queue past the prior 15-min budget when a PR accumulates rapid pushes; 25 min absorbs the variance without making real failures noticeably slower.

The original PR #665 issue — Greptile not re-reviewing the "Fix in Codex" deeplink commit — remains a known gap, but the right response is to force-merge or \`@greptileai review\` manually (the existing escape hatches), not a forgeable bypass.

PR #673, opened 2026-05-18
fix(suno): detect expired __session JWT and refuse stale tokens

mvanhorn:
## Summary

Chrome's `__session` cookie carries a Clerk JWT with a ~1h lifetime. Chrome keeps the cookie row long after the JWT inside it expires, so a long-idle Suno tab leaves a stale value on disk.

The previous flow silently saved the stale value:
- `auth login --chrome` accepted the cookie verbatim
- doctor's read-side validation hit `/api/billing/eligible-discounts` which is lenient and returned 200 even on expired tokens
- every write endpoint (`/api/billing/info`, `/api/generate/v2-web/`) returned HTTP 401 with no hint that pointed back at the cookie

Repro on a fresh checkout: leave a Suno tab idle for >1h, then run `suno-pp-cli generate create ...` → silent 401 loop, even after re-running `auth login --chrome` because Chrome's stored cookie is the same stale JWT until you actually load suno.com again.

## Fix

- New `config.JWTExpiry(token) (time.Time, bool)` decodes the Bearer JWT payload without verifying signature (server enforces signature; we only want to avoid sending tokens we know are expired)
- `auth login --chrome` and `auth refresh` refuse to save an expired token and tell the user to open https://suno.com in Chrome so the cookie rotates
- `doctor` now reports `auth_token: valid for 33m30s` or `auth_token: expired 1h0m0s ago` + actionable hint
- `--fail-on=error` trips on `expired ...` so scheduled agents catch the silent staleness before issuing real requests

## Files

- `internal/config/jwt.go` — new (JWT payload decoder, Config.AuthHeaderExpiry)
- `internal/config/jwt_test.go` — new (6 cases: valid expiry, Bearer prefix, opaque token, missing exp, expired token, nil-safe)
- `internal/cli/auth.go` — refuse expired token on login + refresh, with hint
- `internal/cli/doctor.go` — surface auth_token age + hint; trip --fail-on=error on expiry
- `.printing-press-patches.json` — patch entry

## Test plan

- [x] `go test ./internal/config -v` — 7/7 pass (1 existing config + 6 new JWT)
- [x] `go test ./... -count=1` — 175/175 pass across 11 packages
- [x] `go vet ./...` — clean
- [x] `python3 .github/scripts/verify-skill/verify_skill.py --dir library/media-and-entertainment/suno/` — passes
- [x] Live doctor on fresh token: `"auth_token": "valid for 33m30s"`
- [x] Synthetic expired token: `"auth_token": "expired 1h0m0s ago"`, hint surfaces, `--fail-on=error` exits 1
- [x] Human-readable doctor renders `FAIL Auth Token: expired Xs ago`

### DIFF
--- a/library/media-and-entertainment/suno/.printing-press-patches.json
+++ b/library/media-and-entertainment/suno/.printing-press-patches.json
@@ -1,9 +1,21 @@
 {
   "schema_version": 1,
-  "applied_at": "2026-05-15",
+  "applied_at": "2026-05-18",
   "base_run_id": "20260514-184203",
   "base_printing_press_version": "4.6.1",
   "patches": [
+    {
+      "id": "suno-jwt-staleness-check",
+      "summary": "Decode the __session JWT on login/refresh/doctor and refuse stale tokens with an actionable hint.",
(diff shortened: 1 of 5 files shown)


mvanhorn:
Thanks for landing this, @tmchow - catching the stale Suno JWT before doctor accepts it should save the next hour of head-scratching.

PR #683, opened 2026-05-18
fix(pcgs): unify facts-cert shape; null unpriced; surface year mismatch

vinnyp:
## Summary

Bundles three response-layer fixes for `pcgs-pp-cli` surfaced during a real agentic enrichment pass (31 coins from `pasceri-coins`, 2026-05-18). All three share the same coin-response-transformation layer, so they land together: unify `coin facts-cert` onto the flat `{cert_no, data, _keep}` shape that `coin batch` already emits per JSONL line; rewrite `PriceGuideValue: 0` to `null` so unpriced modern slabs don't silently undercount in sums; and surface PCGS's occasional Name-vs-Year disagreement as a top-level `year_mismatch` object.

## Findings

| ID | Category | Type | Rationale |
|---|---|---|---|
| F1 | output-shape | bug | `coin facts-cert --agent` and `coin batch --agent` used different envelopes (`{meta, results}` vs `{cert_no, data, _keep}`), forcing agents to write two parsers. Unifies onto the batch shape (AGENTS.md P1, locked-in decision). |
| F2 | price-guide-ambiguity | bug | PCGS returns `PriceGuideValue: 0` for unpriced modern slabs (David Hall FDI, brand-new releases — 4-of-31 certs on the surfacing pass). `0` is indistinguishable from "genuinely zero". Rewrites to `null` per AGENTS.md P3 option 1. |
| F3 | year-name-mismatch | feature | PCGS occasionally returns Name="2022-S …" with Year=2021 (2-of-31 certs on the surfacing pass). Silent disagreement forces the agent to guess which field is canonical. Adds top-level `year_mismatch={name_year, year_field}` per AGENTS.md P4. |

## Changes

```
 library/other/pcgs/.printing-press-patches.json            |  25 +++
 library/other/pcgs/SKILL.md                                |  47 +++++-
 library/other/pcgs/internal/cli/coin_batch.go              |  20 ++-
 library/other/pcgs/internal/cli/coin_facts-cert.go         |  60 +++++--
 library/other/pcgs/internal/cli/coin_facts-grade.go        |   4 +
 library/other/pcgs/internal/cli/coin_response_transform.go | 167 +++++++++++++
 6 files changed, 322 insertions(+), 21 deletions(-)
```

[...]

PR #686, opened 2026-05-19
feat(pcgs): omit Images stub field; add coin batch --csv --columns

vinnyp:
## Summary

Final cleanup pass on the AGENTS.md priority list for `pcgs-pp-cli`. Bundles two independent findings from the 2026-05-18 PCGS inventory dogfood, both surfaced in `/Users/vinnypasceri/Projects/pp-pcgs/AGENTS.md`:

- **P6 (Images noise)** — Strips the noisy `Images: [{}, {}]` stub array from every CoinFactsModel response (`coin facts-cert`, `coin facts-grade`, `coin facts-barcode`, `coin batch`). Keeps the `HasObverseImage` / `HasReverseImage` / `HasTrueViewImage` / `ImageReady` booleans so callers can still tell images exist. Image URLs continue to be fetched via `coin images <cert>` (one call, one quota tick).
- **P7 (`--csv` batch export)** — `coin batch` grows `--csv` (flatten the default JSONL stream into one CSV with a header row) paired with `--columns` (comma-separated dotted paths over the `{cert_no, data, _keep}` envelope). Missing fields emit empty cells, NOT `"null"`. New helper file `coin_batch_csv.go` isolates the dotted-path lookup and value-to-cell shaping.

Stacked on top of #683 (unified flat shape) because P7's column paths depend on the `{cert_no, data, _keep}` envelope introduced there. Once #683 merges, this PR's diff against `main` collapses to just the P6 + P7 changes.

P1-P5 already shipped via #682 and #683. This is the last item on the AGENTS.md scope.

## Why P6 fix is safe

The envelope classifier (`classifyPCGSEnvelope` in `pcgs_envelope.go`) runs inside `resolveRead` BEFORE `applyCoinResponseTransforms`. It consumes the `Images=[] + all-Has*Image-false + ImageReady=false` heuristic on the `GetImagesByCertNo` endpoint to flag bogus certs. Stripping `Images` at the response-transform layer happens after the classifier has already done its work — bogus-cert detection on the images endpoint is unaffected.

Also closes a small gap from #683: `coin facts-barcode` returns the same CoinFactsModel shape as `facts-cert` / `facts-grade` but was missing the `applyCoinResponseTransforms` call. Added here so all three coin-fact entry points share one canonical output (PriceGuideValue=0→null, year_mismatch injection, Images stripped).

## Why the user picked option 2 for P6

From AGENTS.md (not re-litigated here): option 1 (auto-fetch URLs) doubles quota per call, violating VISION.md's 1,000/day budget; option 3 (null) leaves the same "where are my URLs?" confusion in a different shape. Option 2 (omit entirely + document the separate `coin images <cert>` route) is the only one that keeps the quota contract intact AND removes ambiguity.

## Verification

```
go build ./...           # PASS
go vet ./...             # PASS
go test ./...            # PASS (all packages)
printing-press publish validate --dir library/other/pcgs --json   # exit 0
```

Live smoke against `/Users/vinnypasceri/Projects/pp-pcgs/pcgs-coin-list.csv`:

P6 acceptance:
```bash
pcgs-pp-cli coin facts-cert 50483263 --agent | jq '.data | has("Images")'   # false
pcgs-pp-cli coin facts-cert 50483263 --agent | jq '.data.HasTrueViewImage'   # true
pcgs-pp-cli coin images 50483263 --agent | jq type                            # "object"  (URLs still flow here)
```

P7 acceptance:
```bash
pcgs-pp-cli coin batch --file pcgs-coin-list.csv \
  --csv --columns "_keep.box,_keep.slot,cert_no,data.Name,data.Year,data.Grade,data.PriceGuideValue" \
  > out.csv
```

Produces a valid CSV with one header row + one row per cert; names containing commas are properly quoted; David Hall null `PriceGuideValue` (from P3 transform) renders as an empty cell — not `"null"`.

Auto-header path (no `--columns`) confirms P6 + P7 interaction: the `Images` column is absent from the auto-discovered header set, exactly as the spec requires.

## Test plan

- [x] `go build ./... && go vet ./... && go test ./...` from `library/other/pcgs/`
- [x] `printing-press publish validate --dir library/other/pcgs --json` exits 0
- [x] `coin facts-cert <cert> --agent | jq '.data | has("Images")'` returns `false` on a cert with images
- [x] `coin facts-cert <cert> --agent | jq '.data.HasTrueViewImage'` still returns the expected boolean
- [x] `coin images <cert> --agent` continues to return image data (unaffected by P6)
- [x] `coin batch --file <csv> --csv --columns "..."` writes a valid CSV with one header + one row per cert
- [x] CSV missing fields render as empty cells, NOT the literal `"null"` or `"<nil>"`
- [x] CSV values containing commas (e.g. coin names with `, DMPL`) are properly quoted
- [x] `coin batch --list-certs --json` and `coin batch --dry-run --json` are unchanged

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>


vinnyp:
Stacked on #683 — contains its commit (aa31978) plus the P6+P7 changes (41ff4c7). Either merge #683 first then this, or merge this and close #683 (its diff would become empty). Either order is safe; the code lands in main in both cases.

vinnyp:
Pushed 5fd3ca8f to address the three Greptile findings:

- **P1 (coin_batch.go:137)** — Replaced `defer csvW.Flush()` with an explicit Flush + `csvW.Error()` check at `runBatch`'s success path. Partial-file truncation now surfaces as a wrapped `csv flush: <err>` return error instead of silently dropping buffered rows.
- **P2 (coin_batch_csv.go:128)** — Removed dead `json.Number` and `json.RawMessage` branches in `csvCellFromValue`. `envelopeToAnyMap` re-decodes top-level `json.RawMessage` values via `json.Unmarshal` into `any`, which only produces standard Go types.
- **P2 (coin_batch_csv.go:170)** — Fixed `autoColumnsFromEnvelope` doc to say "alphabetically sorted" so it matches the `sortedMapKeys` implementation. Users relying on the auto-header to mirror their input CSV's column order should pass `--columns` explicitly.

`go build`, `go vet`, and `go test ./...` all pass locally.

tmchow:
Hey @vinnyp, this PR has unresolved review feedback. Mind taking another pass? If you've hit a blocker or want to hand it off, just leave a note. Otherwise I'll close it for inactivity in about 2 days, happy to reopen anytime.

vinnyp:
Pushed 67467758 addressing the auto-header-from-error-row finding:

- `emit()` now buffers pre-discovery rows (envelopes without `data`) until a row carrying `data` arrives, then derives the schema from THAT row, writes the header, and flushes the buffered rows. Subsequent valid rows get the correct `data.*` columns instead of empty cells.
- `csvFlush()` includes an all-error-batch fallback: if every input row failed and we never saw `data`, the schema is derived from the first buffered row so the CSV carries `cert_no + _keep.* + error` instead of being silently empty.
- `autoColumnsFromEnvelope()` now appends an `error` column at the tail when present so the fallback CSV actually includes the error message.
- Header/row writing extracted into named closures so the new fallback path doesn't duplicate cell-shaping logic.

Passing `--columns` explicitly continues to write the header on the first row unconditionally — that path was unaffected.

`go build`, `go vet`, `go test ./...` all pass locally.

vinnyp:
Pushed 1bf7ed07 addressing the `--columns` silent-no-op finding:

- Added early `usageErr` at the top of `runBatch`: when `columnsSpec != "" && !flags.csv`, return `"--columns requires --csv"` before any store/quota/API work. The caller sees the misconfiguration immediately instead of receiving a full JSONL stream from what they expected to be a filtered CSV.
- Even malformed specs like `--columns ,` now error out via `parseColumnSpec` (which runs inside the `if flags.csv` branch) for users who do pair `--columns` with `--csv`.

`go build`, `go vet`, `go test ./...` all pass locally.

vinnyp:
Pushed d5c74821 addressing the checkpoint-vs-pending-row finding:

The prior deferred-pending-rows fix had a real drop in `--csv --resumable` auto-discovery: buffering a row in `pending` returned `nil` from emit, then the caller wrote the checkpoint anyway. A crash between checkpoint write and pending flush would leave a cert marked complete with no row in the CSV — next run silently skips it.

**Fix:** `emit()` now takes a second arg, the cert number to checkpoint after a successful row write (empty for parse-error rows or non-resumable runs). The checkpoint is written *inside* emit, either:
- immediately, after `writeRow` for an unbuffered row, OR
- in lockstep with the pending flush (`flushPending` writes row + checkpoint per entry), which happens when the schema settles via the first `data` row, or in the `csvFlush` all-error fallback.

The JSONL closure honours the same signature so the call site stays uniform across formats. Caller no longer calls `appendCheckpoint` directly.

`go build`, `go vet`, `go test -count=1 ./...` all pass locally.

vinnyp:
Pushed 95d11855 — `error` column now always emitted in auto-discovery mode.

Previous fix's conditional `if envelope["error"] present` branch never fired in mixed batches: runBatch derives the schema from the first envelope containing `data` (a success row by definition), so the trigger envelope never had an error key. Later failure rows in the same batch would emit against a schema with no error column → all-empty row indistinguishable from missing-data, error text silently dropped.

Now the `error` column is appended unconditionally at the tail. All-success batches take on a single trailing empty column — small cost for ensuring no error text ever drops.

`go build`, `go vet`, `go test -count=1 ./...` all pass locally.

vinnyp:
Rebased onto main (commit 4ece6f07; 7 commits replayed cleanly with one trivial conflict in `.printing-press-patches.json` — kept #687's `skill-cookbook-numista-catalogue-enrichment` entry alongside this PR's two entries). All test plan boxes ticked, ran live against `pcgs-coin-list.csv` (56 certs) with PCGS quota at 6/1000:

| # | Test | Result |
|---|---|---|
| 1 | `go build/vet/test ./...` from library/other/pcgs/ | PASS |
| 2 | `printing-press publish validate --json` | exit 0 |
| 3 | `facts-cert --agent \| jq '.data \| has("Images")'` | `false` |
| 4 | `facts-cert --agent \| jq '.data.HasTrueViewImage'` | `true` |
| 5 | `coin images --agent \| jq type` | `"object"` (unaffected) |
| 6 | `coin batch --csv --columns ...` | 1 header + 56 rows |
| 7 | Missing fields render as empty cells | PASS (grepped for `null`/`<nil>`: zero hits) |
| 8 | Comma-containing names quoted | PASS (3 examples verified; csv parser walks all 56 rows cleanly) |
| 9 | `--list-certs --json` + `--dry-run --json` | 56 entries; `{planned:56, cache_hits:50, live_calls:6, fits_today:true, quota_after:988}` |

CI was already green on the pre-rebase head; force-pushed the rebased branch and waiting for it to re-run.

vinnyp:
Re-rebased onto current main after #682 merged (head: af9d6c6e). Same trivial `.printing-press-patches.json` conflict — kept the quota-stderr entry that #682 just landed alongside this PR's entries. `go build`, `go vet`, `go test ./...` all pass locally. CI re-running.

tmchow:
@vinnyp fix conflicts

vinnyp:
@tmchow fixed

PR #691, opened 2026-05-19
tc-ceo8.14: add workflow verify manifests

cathrynlavery:
Adds workflow_verify.yaml manifests for library/theclose and library/follow-up-boss so dogfood validates end-to-end agent journeys with explicit command bindings.

Verification:
- workflow-verify PASS for follow-up-boss
- workflow-verify PASS for theclose
- dogfood PASS for both with workflow_completeness mapped/passed


cathrynlavery:
Closing this PR because these printed CLIs are The Close/FUB project-specific artifacts and should not be published as visible entries in the public printing-press-library repo. The generic Printing Press fixes continue in mvanhorn/cli-printing-press#1648.

PR #705, opened 2026-05-20
fix(linear): verify credentials in doctor with a real viewer probe

tmchow:
## Summary

`linear-pp-cli doctor` used to report credentials as `WARN … (not verified — set auth.verify_path in spec …)` — a hint aimed at whoever authors the API spec, not the person running the CLI, and with no step they could act on. The check now actually verifies the configured token by running Linear's `viewer` query (the same one `me` uses) and reports one of three honest states: `OK verified (viewer: …)`, `FAIL invalid credentials …`, or `INFO present, probe inconclusive` (with a `Run linear-pp-cli me` hint) when the probe can't reach a verdict.

Two correctness guards are worth calling out because the diff alone doesn't explain them. A 5xx response is never keyword-scanned for auth words, so a vendor error page that happens to mention "token" or "credential" can't masquerade as a credential rejection — that would shift the doctor's wrongness from "WARN with a bad hint" to "FAIL with the wrong cause." And the inconclusive INFO message is deliberately kept free of the substrings `--fail-on=error` keys on (`error`, `invalid`, `unreachable`, `missing`), so an honest "couldn't tell" never silently trips a non-zero exit in CI.

This is a library-side patch recorded in `.printing-press-patches.json` (id `doctor-credential-verify`). The misleading WARN originates in the generator's doctor template, so the durable fix belongs upstream once it grows an `auth.verify_query` field for GraphQL specs; this patch is scoped to the published Linear CLI in the meantime.

## Test plan

- `go build ./...`, `go vet ./...`, `go test ./...` — 176 pass
- `verify_skill.py --dir library/project-management/linear/` — clean
- Manual: valid token → `OK Credentials: verified (viewer: …)`; bad token → `FAIL Credentials: ERROR invalid credentials …` and `doctor --fail-on error` exits 1

---

Claude Code

### DIFF
--- a/library/project-management/linear/.printing-press-patches.json
+++ b/library/project-management/linear/.printing-press-patches.json
@@ -3,5 +3,15 @@
   "applied_at": "2026-05-19",
   "base_run_id": "20260518-232543",
   "base_printing_press_version": "4.9.0",
-  "patches": []
+  "patches": [
+    {
+      "id": "doctor-credential-verify",
+      "summary": "Doctor now runs a real GraphQL `viewer` probe to verify credentials instead of emitting a WARN that asks the user to set a generator-side spec field.",
+      "reason": "The generated doctor template prints `WARN Credentials: present (not verified — set auth.verify_path in spec for an API acceptance check)`. The hint is wrong-audience (users don't own the spec) and offers no actionable next step. Linear has a universal cheap auth probe (`{ viewer { id displayName } }`), so we run it and classify into verified/invalid/probe-failed states. Probe failures fall back to INFO with a `Run linear-pp-cli me` hint.",
+      "files": [
+        "internal/cli/doctor.go"
+      ],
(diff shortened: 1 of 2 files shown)


tmchow:
Addressed the empty-viewer finding in cf9266c's follow-up commit. The inconclusive `default` branch now distinguishes the two paths on stderr:

- `probeErr != nil` → `doctor: credential probe inconclusive: <sanitized error>` (transport / GraphQL error)
- `probeErr == nil` but empty `viewer.ID` → `doctor: credential probe returned an empty viewer; possible API shape change`

The user-facing `INFO` report line is unchanged (still free of the `--fail-on=error` keyword set). Build, vet, and the 36 cli tests pass.

tmchow:
Fixed the P1 (verified line embedding user data into the fail-on keyword scan). Two changes:

1. The verified status is now a clean `"verified"` token; the account name moved to a separate display-only `viewer` field, rendered as a `viewer: <name>` sub-line under `Credentials`. So a displayName like `invalid-test-bot` can no longer skew the indicator or `--fail-on`.
2. `doctorExitForFailOn` now scans only the status-bearing keys (`config`, `auth`, `env_vars`, `api`, `credentials`) instead of every report value — which also closes the latent variant where a `config_path`/`base_url` containing `error` would trip a false failure.

Verified: good token → `OK Credentials: verified` + `viewer: …`, `--fail-on=error` exits 0; bad token → `FAIL … invalid credentials`, `--fail-on=error` exits 1. Build, vet, 36 cli tests, and verify_skill all pass.

The earlier P2 (silent inconclusive on empty-viewer) was addressed in the prior commit with a distinct stderr note.

PR #723, opened 2026-05-20
[codex] Update WHOOP skill local sleep lookup

ffu3xu4tw:
## Summary

- Adds a fast local SQLite lookup path for WHOOP sleep-summary prompts.
- Documents the current cache shape where scored sleep records can live in the `activity` table even when `sleep` is empty.
- Updates Direct Use guidance so agents try the local cache before calling the live sleep collection endpoint.

## Validation

- Ran the documented SQLite query against a local WHOOP cache and confirmed it returns the latest scored overnight sleep.
- Ran `python3 .github/scripts/verify-skill/verify_skill.py --dir library/devices/whoop`; it exited 0 with one existing likely false positive for the documented `auth set-token` example.
- Ran `git diff --check`.


tmchow:

tmchow:

tmchow:

tmchow:
Closing this out. Quiet since the August follow-up on the Greptile policy-gate failure. Fresh PR welcome once that is cleared.

PR #285, opened 2026-05-08
fix(ci): add MCPB release security gates

tmchow:
## Summary

MCPB release artifacts now have pre-signing CI gates for provenance, dependency risk, bundle shape, and newly committed binary payloads. This makes future macOS signing safer: artifacts can be signed only after GitHub-native security checks and a custom MCPB policy verify they were built from source.

## What Changed

This adds GitHub-native guardrails for dependency review, Dependabot updates, CodeQL scanning, reachable Go vulnerability checks, and release artifact attestations. The MCPB-specific policy now rejects newly committed bundle/native binary payloads under `library/`, and the manifest verifier pins generated MCPB manifests to the expected binary entrypoint, empty args, and exact `${user_config.*}` env references.

The stricter verifier also caught and fixes the Seats Aero MCP manifest env mapping, so its bundled server now receives `SEATS_AERO_API_KEY` from the declared user config.

## Test Plan

- `python3 .github/scripts/verify-manifest/verify_manifest.py`
- `python3 .github/scripts/verify-mcpb-policy/verify_mcpb_policy.py`
- `python3 -m py_compile .github/scripts/verify-manifest/verify_manifest.py .github/scripts/verify-mcpb-policy/verify_mcpb_policy.py`
- `ruby -e 'require "yaml"; ARGV.each { |f| YAML.load_file(f); puts "ok #{f}" }' .github/workflows/*.yml .github/dependabot.yml`
- `actionlint .github/workflows/*.yml`

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
Codex

PR #729, opened 2026-05-21
fix(ci): harden MCPB manifest policy

tmchow:
## Summary

This replaces the stale draft in #285 with the MCPB-specific pieces that are still useful on current `main`.

It tightens `manifest.json` validation so generated binary MCPB manifests must keep empty `mcp_config.args` and map each env var to its lower-case `user_config` key. It also fixes the three current manifests exposed by that stricter check: Conduyt CRM, PostHog, and Seats Aero.

The PR also adds a PR-time library convention guard that blocks newly added or modified `.mcpb` and native binary payloads under `library/`. Release assets should be built from source in GitHub Actions before publishing, not committed into the catalog tree.

## Deliberately left out from #285

- `govulncheck`: superseded by the current scoped workflow on `main`.
- dependency review: still requires repo Dependency Graph enablement before it can pass.
- CodeQL / Dependabot: useful as separate repo-admin hardening work, but not part of this MCPB-focused repair.
- artifact attestations: the old draft granted `id-token: write` outside the current allowlist, conflicting with the newer supply-chain policy.

## Validation

- `python3 .github/scripts/verify-manifest/verify_manifest.py`
- `python3 .github/scripts/verify-binary-payload/verify_binary_payload.py --base-ref origin/main`
- `python3 -m py_compile .github/scripts/verify-manifest/verify_manifest.py .github/scripts/verify-binary-payload/verify_binary_payload.py`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/verify-library-conventions.yml"); puts "yaml ok"'`
- `actionlint .github/workflows/verify-library-conventions.yml`
- `git diff --check`

### DIFF
--- a/.github/scripts/verify-binary-payload/verify_binary_payload.py
+++ b/.github/scripts/verify-binary-payload/verify_binary_payload.py
@@ -26,7 +26,7 @@ MAGIC_PREFIXES = {
     b"\x7fELF": "ELF executable/shared object",
     b"MZ": "Windows PE executable",
     b"PK\x03\x04": "ZIP/MCPB archive",
-    b"\xca\xfe\xba\xbe": "Mach-O universal binary",
+    b"\xca\xfe\xba\xbe": "Mach-O universal binary / Java class file",
     b"\xfe\xed\xfa\xce": "Mach-O executable",
     b"\xfe\xed\xfa\xcf": "Mach-O executable",
     b"\xce\xfa\xed\xfe": "Mach-O executable",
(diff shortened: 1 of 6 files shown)


tmchow:
Greptile note on inverse `user_config -> mcp_config.env` validation: I tested that stricter symmetric check against the full catalog and it currently fails existing manifest shapes such as `cf-domain` and `namecheap`, where `user_config` includes values that are not injected one-to-one into `mcp_config.env`. I am intentionally deferring that broader generator/catalog audit out of this PR; this PR validates the env entries that are actually injected and fixes the concrete broken MCPB env mappings it exposes.

PR #731, opened 2026-05-21
fix(setlist-fm): repair auth set-token, doctor env-vars, and signup hint

mvanhorn:
Three related auth bugs in the setlist-fm CLI shipped via #724 made the documented setup flow fail silently. A fresh user ran `auth set-token`, the CLI reported the token saved, and every subsequent API call returned HTTP 403.

## Bugs fixed

**1. `auth set-token` wrote to the wrong config field.** The token landed in `access_token` (a dead OAuth field). The HTTP layer reads from `fm_api_key`, so the saved token was never used. Re-routed via a new `Config.SaveAPIKey` method.

**2. Dropped the unused OAuth-shaped fields from the config schema.** Setlist.fm's API has no OAuth flow — only the static `x-api-key` header. The generator-emitted `auth_header`, `access_token`, `refresh_token`, `token_expiry`, `client_id`, `client_secret` fields were dead storage. Removed alongside the dead `refreshAccessToken` path in the HTTP client.

**3. `doctor` reported `FAIL Env Vars` when config-based auth was valid.** The env var is one of multiple valid auth sources, not a hard requirement. When config provides auth, the env-vars verdict is now `OK config provides auth (env var optional)`.

**4. `doctor` hint only mentioned the env var path.** Now leads with `setlist-fm-pp-cli auth set-token <key>` (the more durable path, and the one that actually works after this fix) and explicitly says the API key is free.

## Backward compatibility

Legacy configs containing the removed OAuth fields load without error (`go-toml/v2` ignores unknown keys); the dead rows are dropped on the next save. No migration tooling needed.

## Patch tracking

Recorded as the `auth-flow-fix` patch in `.printing-press-patches.json` so Printing Press tooling re-applies the change after future regenerations. Upstream generator template fix tracked separately.

## Verification

- `go test ./...` — 109 passed in 11 packages
- `go vet ./...` — clean
- `go build ./cmd/setlist-fm-pp-cli ./cmd/setlist-fm-pp-mcp` — both build
- `python3 .github/scripts/verify-skill/verify_skill.py --dir library/media-and-entertainment/setlist-fm/` — all checks pass
- Manual smoke: with `fm_api_key` set in config and no env var, `setlist-fm-pp-cli doctor` shows all five health checks as `OK` (was previously `FAIL Env Vars`); `auth set-token TESTKEY123` writes `fm_api_key = 'TESTKEY123'` and no legacy fields are emitted

Closes the issues from `setlist-fm-pp-cli-bugs.md`. Related: #724.

### DIFF
--- a/docs/plans/2026-05-20-002-fix-setlist-fm-auth-and-doctor-plan.md
+++ b/docs/plans/2026-05-20-002-fix-setlist-fm-auth-and-doctor-plan.md
@@ -1,0 +1,235 @@
+---
+title: "fix(setlist-fm): repair auth set-token, doctor env-vars, and signup hint"
+type: fix
+status: active
+created: 2026-05-20
+depth: standard
+target_repo: printing-press-library
+origin:
+  bug_report: /Users/mvanhorn/Downloads/setlist-fm-pp-cli-bugs.md
+  related_pr: https://github.com/mvanhorn/printing-press-library/pull/724
+---
+
+# fix(setlist-fm): repair auth set-token, doctor env-vars, and signup hint
+
+## Summary
+
+The setlist-fm CLI (shipped via PR #724) has a broken auth flow. Three related bugs make the documented setup path fail silently:
+
+1. `auth set-token` writes the API key to a dead OAuth field (`access_token`); the HTTP layer reads from `fm_api_key`, so the saved token is never used and every request returns HTTP 403.
+2. `doctor` always emits `FAIL Env Vars` when the env var is unset, even when `fm_api_key` is correctly configured in `config.toml`.
+3. The `doctor` hint only mentions the env var path; it does not mention `auth set-token` (the more durable config path) and does not say the key is free.
+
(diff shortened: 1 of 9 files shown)


mvanhorn:
Both findings addressed in b2c05de8:

- **P1 logout still-set check**: `newAuthLogoutCmd` now chains `SETLISTFM_API_KEY` (higher priority in `config.Load`) ahead of `SETLIST_FM_API_KEY`. Four scenarios covered in the new `internal/cli/auth_test.go` (only-higher-priority, only-lower-priority, both-set, neither-set).
- **P2 doctor tests hitting api.setlist.fm**: `writeConfigForStubAPI` now stands up an `httptest.NewServer` registered with `t.Cleanup`. All existing assertions pass unchanged; `TestDoctorAPIReachableThroughStub` confirms the stub is actually wired in. Doctor test suite drops from network-bound to ~190ms.

mvanhorn:
Thanks for the merge, @tmchow. Good to have the setlist-fm auth set-token and doctor env-var fixes in.

PR #807, opened 2026-05-24
fix(fireflies): tolerate paid_required GraphQL errors

shanegardner405-arch:
## Summary

The Fireflies GraphQL API returns per-field `paid_required` errors (in `extensions.code`) alongside partial data for fields like `video_url` that are gated behind paid plans. The previous client treated any non-empty `errors` array as a hard failure, dropping otherwise-valid transcript data on the floor for non-business-tier accounts.

This PR makes the GraphQL client tolerant of plan-gated errors and drops `video_url` from the transcript query so the round-trip isn't wasted on a guaranteed plan-gated field.

## Changes

- `internal/gql/client.go` — parse `extensions.code`, classify each error as plan-gated vs. genuine, return partial `data` when every error is plan-gated. Non-plan-gated errors still fail the request.
- `internal/cli/transcripts_ff.go` — remove `video_url` from the Transcript query (paid-tier field).

## Verification

- ` + "`go build ./...`" + ` clean
- ` + "`go vet ./...`" + ` clean
- Dogfooded against Shane's Fireflies account (non-business tier): ` + "`fireflies transcripts get <id>`" + ` now returns transcript data instead of failing with ` + "`GraphQL error: paid_required`" + `.

## Commits

1. ` + "`fix(fireflies): tolerate paid_required GraphQL errors`" + ` — client.go tolerance logic
2. ` + "`fix(fireflies): drop video_url from transcript query`" + ` — remove the field that triggers the gate

### DIFF
--- a/library/productivity/fireflies/internal/cli/transcripts_ff.go
+++ b/library/productivity/fireflies/internal/cli/transcripts_ff.go
@@ -32,7 +32,6 @@ query Transcript($id: String!) {
     privacy
     transcript_url
     audio_url
-    video_url
     is_live
     meeting_link
     calendar_type
--- a/library/productivity/fireflies/internal/gql/client.go
+++ b/library/productivity/fireflies/internal/gql/client.go
@@ -54,7 +54,22 @@ type response struct {
 }
 
 type gqlError struct {
-	Message string `json:"message"`
+	Message    string         `json:"message"`
+	Path       []any          `json:"path,omitempty"`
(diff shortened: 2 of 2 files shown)


tmchow:
@shanegardner405-arch please address code review feedback

tmchow:
Closing this for now after a couple quiet days. Feel free to reopen anytime you want to keep going.

(This is an automated triage message. If you think it was closed in error, just reopen it or leave a note here and a maintainer will take a look.)

PR #678, opened 2026-05-18
feat(icloud): add icloud-pp-cli

matysanchez:
## icloud

Query iCloud Photos data stored locally on macOS — storage analysis, heaviest-file ranking, and safe deletion via AppleScript — all without launching Photos.app or making any network calls. Reads `Photos.sqlite` in read-only mode directly; `photos delete` moves items to Recently Deleted via `osascript`, preserving the 30-day recovery window.

**API:** iCloud (local SQLite) | **Category:** media-and-entertainment | **Press version:** 4.8.0

**Spec:** local-sqlite (no network spec; source is `~/Pictures/Photos Library.photoslibrary/database/Photos.sqlite`)

### CLI Shape

```
icloud-pp-cli gives AI agents and power users direct access to iCloud data
stored locally on macOS — Photos library storage analysis today, with Contacts
and more coming.

All reads are local. No network calls, no iCloud API token required.

Pipe any command for automatic JSON output.

Usage:
  icloud-pp-cli [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
  doctor      Check system requirements and Photos library access
  help        Help about any command
  photos      Query and manage your Photos library

Flags:
      --agent            Agent-friendly mode: --json --compact --no-color
      --compact          Minimal fields only
  -h, --help             help for icloud-pp-cli
      --json             Output as JSON
      --library string   Path to Photos.sqlite (default: ~/Pictures/Photos Library.photoslibrary/database/Photos.sqlite)
      --no-color         Disable colored output
  -v, --version          version for icloud-pp-cli
```

### What This CLI Does

- **`photos stats`** — total item count and total library size in one query
- **`photos storage`** — storage breakdown by media type (photo/video) and by year
- **`photos top`** — top N heaviest files across all media types, filterable by `--type photo|video`
- **`photos videos`** — largest videos sorted by size, filterable by `--year` / `--month`
- **`photos delete`** — dry-run preview by default; `--confirm` moves items to Recently Deleted via `osascript`
- **`doctor`** — pre-flight check: macOS guard, Photos.app presence, library path, read access, schema, asset count

All read commands open `Photos.sqlite` with `mode=ro&_query_only=1`. `--json` on any command emits newline-delimited JSON for piping to `jq` or an AI agent.

### Manuscripts

- [Research](library/media-and-entertainment/icloud/.manuscripts/20260517-000000/research/icloud-research.md)
[...]

PR #811, opened 2026-05-24
feat(productivity): add chrome-history CLI

vinnyp:
## Summary

Add `chrome-history` — a local-first Chrome browsing-history CLI. It reads Chrome's local `History` SQLite database, snapshots it, builds an offline FTS5 full-text index, and answers questions about your browsing. Read-only, zero network, fully on-device. Ships a human CLI surface **and** 19 read-only MCP tools (`chrome-history-pp-mcp`).

What sets it apart from a plain history dump is the **transcendence layer** — commands that reconstruct, analyze, and profile browsing rather than just list it:

| Command | Name | What it does |
|---------|------|--------------|
| `journeys` | Chrome Journeys clusters | Surfaces Chrome's own on-device topic clusters from the local history DB |
| `topic` | Topic gather | Merges Journeys clusters with full-text page matches to assemble everything browsed about a topic |
| `timeline` | Session reconstruction | Rebuilds ordered browsing sessions for a date/window, splitting on an idle-gap threshold |
| `rabbitholes` | Drift detection | Finds sessions that started on a productive (typed) page and drifted into distracting ones, with time-of-day patterns |
| `dwell` | Time-on-site estimate | Estimates per-domain dwell time by inferring sessions from inter-visit gaps (Chrome's `visit_duration` is 0 on ~89% of rows) |
| `graph` | Navigation graph | Builds a page-node / `from_visit → to` edge graph over a window (JSON or Graphviz `dot`) |
| `profile` | Behavioral self-profile | Summarizes peak hours, busiest weekday, top domains/search terms, and a productive/neutral/distracting split |
| `heatmap` | Activity heatmap | Weekday-by-hour visit-count grid (ASCII or JSON) |
| `devices` | Cross-device origins | Partitions history into this-device vs other synced devices via `visit_source` + `originator_cache_guid`; every read accepts `--device all\|this\|synced\|device-N` |

Topic categorization is intentionally split: the CLI provides clean data + coarse hints (`domains` buckets, `journeys` clusters), and agents do the real clustering by reading titles — documented in the SKILL/README.

## Published CLI Metadata

**API:** `chrome-history` (local SQLite) | **Category:** `productivity` | **Press version:** `4.13.0`
**Spec:** local-sqlite (no network spec; source is `~/Library/Application Support/Google/Chrome/Default/History`)
**Required env:** none

## What Changed

- New CLI under `library/productivity/chrome-history/`.

## CLI Shape

```bash
$ chrome-history-pp-cli --help
[...]

PR #814, opened 2026-05-24
feat(productivity): add safari-history CLI

vinnyp:
## Summary

Add `safari-history` — a local-first Safari browsing-history CLI. It reads Safari's local `History.db`, snapshots it, builds an offline FTS5 full-text index, and answers questions about your browsing. Read-only, zero network, fully on-device. Ships a human CLI surface **and** 19 MCP tools (`safari-history-pp-mcp`; all read-only except the `sync` write tool).

It's the Safari sibling of `chrome-history` and shares the same source-adapter architecture — Safari plugs into the identical `Source` interface, which is what made it a clean second implementation. Safari stores less than Chrome, so some commands are honestly capability-gated rather than faked:

| Command | Name | What it does |
|---------|------|--------------|
| `timeline` | Session reconstruction | Rebuilds ordered browsing sessions for a date/window, splitting on an idle-gap threshold |
| `dwell` | Time-on-site estimate | Estimates per-domain dwell time by inferring sessions from inter-visit gaps |
| `graph` | Navigation graph | Page-node / referrer-edge graph over a window (JSON or Graphviz `dot`) |
| `profile` | Behavioral self-profile | Peak hours, busiest weekday, top domains, productive/neutral/distracting split |
| `heatmap` | Activity heatmap | Weekday-by-hour visit-count grid (ASCII or JSON) |
| `devices` | Cross-device origins | Partitions history into this-device vs synced (Safari's `origin` is binary — no per-device identity, unlike Chrome) |

**Capability-gated (Safari's `History.db` doesn't store the underlying data):** `journeys`, `searches`, `downloads`, and `rabbitholes` (needs navigation transition types Safari doesn't record) return an honest "not available for Safari" rather than fabricating or silently-empty data.

## Published CLI Metadata

**API:** `safari-history` (local SQLite) | **Category:** `productivity` | **Press version:** `4.13.0`
**Spec:** local-sqlite (no network spec; source is `~/Library/Safari/History.db`, requires macOS Full Disk Access)
**Required env:** none

## What Changed

- New CLI under `library/productivity/safari-history/`.

## CLI Shape

```bash
$ safari-history-pp-cli --help
Local-first Safari history CLI

Usage:
  safari-history-pp-cli [command]

Available Commands:
  agent-context Emit the machine-readable command tree for agents and tooling
  devices       List visit-origin buckets (Safari reports this/synced, no per-device split)
  doctor        Check safari db and snapshot health
  domains       Rank the most-visited registrable domains over --since, with productivity category
  downloads     (not available for Safari — History.db omits downloads)
  dwell         Estimate time-on-site per domain over --since, capping per-visit dwell at --gap
  graph         Build a navigation graph over --since (JSON or Graphviz dot; edges sparse on Safari)
  heatmap       Render a weekday-by-hour activity heatmap of visit counts over --since
  journeys      (not available for Safari — no journeys tables)
  list          List recent visits, filtered by --since/--until, --domain, --min-visits
  mcp           Start MCP stdio server
  profile       Summarize your browsing self over --since: peak hours, busiest weekday, top domains
  rabbitholes   (not available for Safari — no navigation transition types)
  report        Summarize browsing activity over --since: per-day/per-hour counts, top domains, split
  search        Full-text (FTS5) search over URLs and titles, ranked by relevance
  searches      (not available for Safari — History.db omits search terms)
  sql           Run a read-only SELECT query against the snapshot's history tables
  sync          Snapshot Safari history and build FTS
  timeline      Reconstruct ordered browsing sessions for a date or --since/--until window
  topic         Gather everything you browsed about a topic via full-text page matches
  visited       Check whether a URL or domain was visited, with first/last seen and referrers

Flags:
      --browser string   history source (default "safari")
      --compact          compact output
      --csv              CSV output
      --device string    visit origin filter: all|this|synced|device-N (default "all")
  -h, --help             help for safari-history-pp-cli
      --json             JSON output
      --limit int        row limit (default 20)
      --profile string   Safari profile (ignored; Safari uses default history DB) (default "Default")
      --quiet            suppress output
      --select string    select dotted paths
  -v, --version          version for safari-history-pp-cli
```

## What This CLI Does

`safari-history-pp-cli` turns Safari's local history DB into a queryable, agent-friendly surface without anything leaving the machine. Every read command supports `--json`/`--select`/`--csv`/`--compact`, and the read surface is mirrored as MCP tools (`safari-history-pp-cli mcp`, or the standalone `safari-history-pp-mcp` binary).

- **Find:** `search` (FTS5), `visited`, `list`, `topic`
- **Aggregate:** `domains`, `report`, `heatmap`, `profile`, `dwell`
- **Reconstruct:** `timeline`, `graph`
- **Cross-device:** `devices` + `--device this|synced` filter (Safari `origin` is binary)
- **Ops:** `sync`, `doctor`, `sql` (read-only SELECT), `agent-context`

### Novel Commands

| `timeline` | Session Timeline Reconstruction | Reconstructs session-level browsing timelines from visit timestamps and an idle-gap split. |
| `dwell` | Derived Dwell Estimate | Estimates per-domain time-on-site from inter-visit gaps. |
| `graph` | Navigation Graph | Builds a page-node / referrer-edge navigation graph (JSON or Graphviz dot). |

## Manuscripts

- [Research Brief](library/productivity/safari-history/.manuscripts/20260523-231334/research/safari-history-research.md)
- [Phase 5 acceptance proof](library/productivity/safari-history/.manuscripts/20260523-231334/proofs/phase5-acceptance.json)

## Validation

| Check | Result |
|---|---|
| Manifest (`.printing-press.json`) | PASS |
| Transcendence | PASS |
| Phase 5 acceptance | PASS (18/18 commands, hand-attested — see note) |
| `go mod tidy` | PASS |
| `go vet ./...` | PASS |
| `go build ./...` | PASS |
| `go test ./...` | PASS |
| `--help` | PASS |
| `--version` (`safari-history-pp-cli 0.1.0`) | PASS |
| verify-skill | PASS |
| `govulncheck ./...` | PASS (0 reachable) |

### Publication Path

New print.

## Gaps / Follow-Up

- **macOS only.** Safari does not exist off macOS, and reading `History.db` needs Full Disk Access. Not portable to Linux/Windows by nature.
- **Phase 5 hand-attested.** The automated `dogfood --live` runner scopes `HOME`/`XDG_CACHE_HOME` away from the local source DB and cannot exercise a local-SQLite CLI, so all 18 commands were run against the real local database and verified (exit 0, valid output). Follows the `icloud-pp-cli` local-source convention (#678); superseded once dogfood gains local-source support.
- **Hand-built local-SQLite CLI.** No OpenAPI spec — the source is Safari's local DB. Built in Printing Press agent-native shape on the same `Source` adapter interface as `chrome-history`; the three hand-authored divergences are documented in `.printing-press-patches.json`. `journeys`/`searches`/`downloads` are capability-gated off because Safari's `History.db` doesn't store those datasets.


vinnyp:
Addressed the review findings:

1. **MCP binary deadlock (critical)** — `cmd/safari-history-pp-mcp` now dispatches subcommand args to the CLI root, so the in-process `runSelf` (which shells back into `os.Executable()`) returns real tool data instead of starting a second server. Verified: `domains` via the standalone MCP binary now returns real rows.
2. **`rabbitholes` always-empty** — capability-gated on `Transitions`. Safari records no transition types, so it now reports "not available for Safari" instead of silently returning zero sessions. Removed from `novel_features` (added `graph`).
3. **`RowCount` table name** — added an allowlist for the interpolated table name.
4. **`sync` cleanup** — remove the temp snapshot if the final rename fails.

(Findings 1, 3, 4 also applied to the chrome-history PR #811.) `go build`/`vet`/`test`/`govulncheck` green.

vinnyp:
Addressed both findings:

1. **WAL-safe snapshot** — `copySnapshot` now prefers a read-only `VACUUM INTO` (reads through Safari's WAL so visits not yet checkpointed into History.db are captured; the live DB is opened read-only and never modified), falling back to `cp` if Safari holds an exclusive lock so sync never breaks.
2. **FTS relevance rank** — `search` now surfaces real `bm25()` scores and orders by them (was hardcoded `rank: 0`, ordered by recency despite "ranked by relevance"). bm25 lives in a `MATERIALIZED` FTS-only CTE because SQLite rejects the ranking functions when the matched table is inside a JOIN. Verified against the real DB: `search` now returns descending-relevance rows with populated `rank`.

`go build`/`vet`/`test` green.

vinnyp:
Fixed: the `sql` read-only guard was rejecting valid SELECTs that contained a blocked word inside a string literal (`WHERE url LIKE '%create%'`, `'%delete%'`) or as the `REPLACE()` function. Now: string-literal contents are blanked before the keyword scan, `replace` is dropped from the blocklist (it's a read-only function; a REPLACE *statement* is still caught by the SELECT/WITH prefix check), and read-only is enforced authoritatively via `PRAGMA query_only` on the RunSelect connection. Verified: `LIKE '%delete%'` and `REPLACE(...)` now return rows; `DELETE …` and a DML-in-subquery are still rejected. Added regression cases. (Same fix applied to chrome #811.)

vinnyp:
Carried over the same `report --compact` projection fix from chrome #811 (report was reusing profile's fields, dropping per_day/hour_of_day). report and profile now have distinct compact projections.

vinnyp:
Carried over the same `productivity_split` accuracy fix from chrome #811 (was computed from top-N domains only; now from all domains in the window).

vinnyp:
Single-day `timeline <date>` scoping fix + the same seed-a-snapshot integration harness (`internal/source/safari/integration_test.go` + `internal/cli/timeline_window_test.go`) that asserts on domain ranking, profile counts, FTS bm25 rank, and the timeline window. (Safari's DomainStats already aggregated all rows, so no ranking fix needed there.)

PR #825, opened 2026-05-24
fix(digg): rank-merge first-non-zero-wins and peakRank sentinel normalization

mvanhorn:
## The bug

`digg-pp-cli top --limit 20 --json` returned 7 distinct stories all stamped `currentRank: 1`, in an order that did not match `digg.com/ai`. The Karpathy joins-Anthropic story sat at rank 1 for 4+ days; the Cursor Composer story leaked `peakRank: 9999`.

## Root cause

`internal/diggparse/parser.go` ExtractClusters walks every JSON object containing `"clusterId":"` in the decoded RSC stream and merges occurrences by clusterId via mergeClusters. The /ai page legitimately emits the same cluster in multiple sections — main leaderboard `items[]`, plus one or more of `featuredStories`, `weeklyTop`, `sevenDaysStories`, pinned. Each section carries its own `rank` for the cluster.

mergeClusters previously used last-wins-when-nonzero for every field. Whichever section appeared last in the stream stamped the cluster's `currentRank`. Featured / weekly / pinned sections tend to assign low ranks like 1 to a small curated set, so they displaced the live leaderboard ranks of clusters that appeared in both.

`peakRank: 9999` is Digg's internal placeholder for newly-detected clusters with no tracked all-time peak; it should be normalized at the CLI's output boundary.

## The fix

Gate the four rank-bearing branches in mergeClusters (CurrentRank, PeakRank, PreviousRank, Delta) on `accumulator == 0` so first non-zero wins. The main /ai leaderboard array is emitted first in the RSC stream, so first-wins captures the canonical leaderboard rank for any cluster present in both sections. Zero accumulator still accepts a later non-zero, preserving sparse-then-full enrichment that legitimately fills missing fields from later occurrences.

Normalize `PeakRank == 9999` to 0 in ExtractClusters' output loop. The struct's `json:peakRank,omitempty` tag drops zero values, so the sentinel disappears from CLI JSON output without a schema change.

## Tests

`internal/diggparse/parser_rank_test.go` adds 5 scenarios:

- `TestMergeClustersFirstWinsForRankFields` — same clusterId twice with differing ranks; asserts first-wins for CurrentRank, PeakRank, PreviousRank, Delta.
- `TestMergeClustersStillGapFillsZeroAccumulator` — sparse first occurrence, full second; asserts later non-zero still fills the gap.
- `TestMergeClustersDistinctClustersPreserveOwnRanks` — two distinct clusters with rank 1 and rank 2; asserts no collapse.
- `TestExtractClustersNormalizesPeakRankSentinel` — peakRank=9999 in; PeakRank=0 out and json.Marshal omits the field.
- `TestExtractClustersPreservesLegitimatePeakRank` and `TestExtractClustersOmitsAbsentPeakRank` — non-sentinel values pass through unchanged; absent field stays absent.

61 of 61 diggparse tests pass. go vet clean.

## Out of scope

- Envelope-aware RSC extraction (parsing the RSC tree to scope cluster extraction to the main feed array only). Held in reserve if cross-section pollution recurs in fields the merge gate does not cover.
- Pre-existing govulncheck finding in `internal/client/digg_html.go` (golang.org/x/net@v0.43.0). Unrelated to this fix; per AGENTS.md scoped fixes do not double as dep-freshness campaigns.
- Two adjacent UX bugs in the report (`authors top --agent` returns empty objects; `author <name>` matches contributors only) deferred to follow-up.

Patches recorded in `.printing-press-patches.json`.


mvanhorn:
Addressed the P1 in 021cbe02. The fix moves the `PeakRank == 9999` normalization from the output loop into the parse loop, right next to the existing `Rank → CurrentRank` normalization. Now the merge guard sees the sentinel as 0 and a later occurrence's legitimate PeakRank fills the gap correctly.

Added `TestExtractClustersSentinelDoesNotBlockLaterPeakRank` to pin the interaction: first occurrence has `peakRank: 9999`, second has `peakRank: 42`, output must be 42. 62/62 diggparse tests pass.

Patches manifest entry updated to reflect the parse-time placement.

mvanhorn:
Glad the digg rank-merge fix is in, thanks @tmchow; first-non-zero-wins fixes the duplicated cluster ranks.

PR #826, opened 2026-05-24
feat(sweep-learn): add sweep-learn-install tool for retrofitting learn loop

mvanhorn:
## Summary

Phase 2 of plan [`cli-printing-press#2085`](https://github.com/mvanhorn/cli-printing-press/pull/2085) (generator-wide self-learning CLI loop). Builds `tools/sweep-learn-install/` to retrofit the learn package + commands + schema migration into existing library CLIs. Sibling to `tools/sweep-canonical/` with the same GOPATH-mode + idempotency contract.

## What ships

- AST surgeon for `internal/cli/root.go` (rejects legacy `var rootCmd` shape with clear manual-review error)
- Anchor-based migrations-slice rewriter for `internal/store/store.go` + `StoreSchemaVersion` bump
- Byte-for-byte parity renderer for `internal/learn/*.go` (27 templates embedded via `go:embed` from the cli-printing-press generator)
- SKILL.md Automatic Learning section patcher (anchor-bracketed, idempotent)
- `go mod tidy` helper for CLIs without `modernc.org/sqlite`
- `printing_press_version` manifest stamper
- Snapshot-restore on per-CLI failure
- `-readme-only`, `-dry-run`, `SWEEP_LIBRARY_ROOT` flags
- 28 tests pass; sibling `tools/sweep-canonical/` tests still pass (no regression)

## Known gaps for U14 (pilot sweep)

1. **Existing library CLIs lack the `// CLI Printing Press: learn migrations` anchor in their `store.go`** — the anchor was introduced by U6 in the cli-printing-press PR, so any CLI generated before that PR ships will be skipped with `SKIPPED (anchor not found)`. U14 needs either a one-time bootstrap step that adds the anchor, or this tool needs to be enhanced to handle the no-anchor case via AST search.
2. **The sweep does not emit `internal/cli/teach.go` or `internal/cli/learn_init.go`** because those are spec-driven (TickerPatterns, Stopwords, EntityLookupSeeds) and the per-CLI Learn config is per-CLI knowledge. U14 needs to hand-author a `learn:` block per pilot CLI spec OR this tool needs a stub-emit mode for empty defaults.
3. **Legacy `var rootCmd` shape** (agent-capture, instacart) is refused with `manual review required` per AGENTS.md — those CLIs need a separate retrofit path.

These gaps are intentional in this PR; resolving them is part of U14's pilot sweep work, where per-CLI authorship + tool refinement can be done together.

## Verification

- `cd tools/sweep-learn-install && GO111MODULE=off go test .` — 28 pass
- `cd tools/sweep-canonical && GO111MODULE=off go test .` — still green
- Byte-for-byte parity validated against `cli-printing-press/testdata/golden/expected/generate-learn-loop-api/learn-loop-example/internal/learn/`

## Test plan

- [ ] CI green
- [ ] Sweep dry-run against a representative CLI shows expected skip + planned modifications
- [ ] Decide which option (bootstrap pre-pass, tool enhancement, or per-CLI hand-craft) covers gaps 1 and 2 for U14


tools/sweep-learn-install/learn_files.go
```diff
@@ -0,0 +1,267 @@
+// learn_files.go renders the internal/learn package files byte-for-
+// byte identical to what cli-printing-press's generator emits when a
+// spec opts into the self-learning loop.
+//
+// Parity strategy:
+//
+//   - The generator's learn templates are embedded into this binary
```
tmchow, line 7: This parity strategy currently points at the unaccepted #2085 template shape. Because #2085 is changes-requested, this tool needs an explicit sync point to the accepted generator commit/release and should be refreshed after the generator fixes land. Without that, the library sweep can preserve stale schema/init/CLI-help behavior even if the generator PR is corrected.

tools/sweep-learn-install/root_ast.go
```diff
@@ -0,0 +1,453 @@
...
+//
+// Idempotent: skipped when learnHookSkipList already exists.
+func injectLearnHookSkipList(src string) (string, bool) {
+	if strings.Contains(src, "learnHookSkipList") {
+		return src, false
+	}
+	// Append at file end so we don't disturb any existing top-level
+	// declarations. The block carries its own doc comment so a
+	// downstream reader knows what it's for without grepping.
+	insertion := "\n// learnHookSkipList enumerates framework command names that any\n" +
```
tmchow, line 608: This is the core approach mismatch with the generator review. The sweep injects a skip list/helper but no actual learn-init hook; if #2085 keeps the hook, swept CLIs will not match fresh generator output, and if #2085 moves to a lazy/no-ambient init strategy, this block should change or disappear. Please wait for the generator init contract to settle before landing the retrofit tool.

tools/sweep-learn-install/templates/cli/teach.go.tmpl
```diff
@@ -0,0 +1,1030 @@
...
+	var strategy string
+	var entityKind string
+	var dbPath string
+
+	cmd := &cobra.Command{
+		Use:   "teach-pattern",
+		Short: "Install a manual generalization pattern (query_template, resource_template, entity_kind)",
+		Long: `Adds one row to search_patterns. The recall path uses patterns to
+substitute live query entities into a resource template, so a pattern
+with query_template="odds {entity} wins world cup" and
```
tmchow: Same issue as the generator PR: the teach-pattern help text leaks a domain-specific betting/sports example into every retrofitted CLI. The accepted template should use neutral placeholder resources, and the purity check should cover this CLI surface too so future generator/sweep copies cannot drift back.

tools/sweep-learn-install/main.go
```diff
@@ -0,0 +1,558 @@
...
+
+	storeAfter, storeChanged, err := patchStoreMigrations(string(storeData), ctx)
+	if err != nil {
+		return plan, fmt.Errorf("patch store.go: %w", err)
+	}
+	plan.StoreAfter = storeAfter
+	plan.StoreChanged = storeChanged
+
+	skillPath := filepath.Join(ctx.CLIDir, "SKILL.md")
+	if skillData, err := os.ReadFile(skillPath); err == nil {
```
tmchow: This still silently ignores SKILL.md read errors. Missing SKILL.md can be non-fatal if that is an intentional supported case, but permission/read/parse errors should surface in the plan so the sweep does not report a clean retrofit while skipping the agent-facing skill instructions.

tools/sweep-learn-install/store_migration.go
```diff
@@ -0,0 +1,497 @@
...
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"regexp"
+	"strings"
+)
+
+const (
+	learnMigrationAnchor = "// CLI Printing Press: learn migrations"
+	learnSchemaVersion   = 3
```
tmchow, line 38: This hard-coded schema version is a copied generator contract. For a learn-enabled retrofit it is reasonable to bump the store schema, but the final value and migration shape need to come from the accepted #2085 generator contract, not the current disputed template snapshot.

tmchow (changes requested):
I’m requesting changes for sequencing and contract reasons. The sweep-tool direction is useful, but this PR embeds and replays the generator learn templates/schema while cli-printing-press#2085 is currently changes-requested for schema gating, store coupling, learn-init side effects, and domain-neutrality. Please treat this as blocked on the accepted #2085 contract, then refresh the embedded templates/tooling and rerun the pilot. Otherwise the library sweep can lock in a generator shape we have explicitly asked to change.

mvanhorn:
Pushed three commits addressing the Greptile P1 + tmchow threads. cli-printing-press PR #2085 has merged to main at `23b6239c`, which clears a chunk of the now-stale tmchow concerns.

### Per-thread status

| Thread | Status | Commit / reasoning |
|---|---|---|
| `root_ast.go:435` Greptile P1 silent partial-patch | Fixed | `e0a3eb54` — `patchRootAST` now refuses with a clear error when the `PersistentFlags()` anchor is absent AND the `BoolVar(&flags.noLearn, "no-learn"` binding isn't already present. No more silent emission of a non-functional `--no-learn` flag. |
| `main.go:466` Greptile P1 + P2 + tmchow SKILL.md silent read errors | Fixed | `e0a3eb54` — now distinguishes `os.IsNotExist` (skip silently, since SKILL.md is genuinely optional) from other errors (return wrapped error). |
| `main.go:473` Greptile P1 go.mod silent read errors | Fixed | `e0a3eb54` — same shape as the SKILL.md fix; `os.IsNotExist` skips, other errors abort. |
| `store_migration.go:397` Greptile P1 `ensureStoreSchemaVersion` inserts const before package | Already fixed by `2080be2b` (this PR) | The current implementation uses `parser.ParseFile` to derive byte offsets from `f.Name.End()`, not a string scan. Greptile is reviewing an earlier shape of the code. `TestEnsureStoreSchemaVersion_InsertsAfterPackageDecl_WhenAbsent` regression-pins the exact `Bug A` Greptile describes (package doc comment before `package`). |
| `store_migration.go:273` Greptile P1 `ensureStoreOpenAcceptsPath` assumes OpenAt exists | Fixed | `a92f554b` — added a `func OpenAt(` substring guard alongside the existing `config.Dir()` body-shape check. Zero published library `store.go` files trigger the new refusal path today, so this is defense-in-depth for retrofit robustness. |
| `teach.go.tmpl:84` Greptile P1 `usageErr` undeclared | Fixed | `7c088138` — added `learnUsageErr` inline helper following the existing `learnDryRunOK` / `learnPrintJSON` / `learnParentNoSubcommandRunE` pattern, rewrote all 9 call sites. Added `usageErr(` to `bannedCallSitePatterns` and `learnUsageErr` to the positive assertions so future regressions surface immediately. 3 published library CLIs (airframe, monarch-money, slack) ship without `func usageErr` — confirmed by `grep` — and 2 of those 3 lack `cliError` entirely, which is why the wrapper returns the error verbatim rather than constructing a `*cliError`. |
| `teach.go.tmpl:901` tmchow domain-neutral example | Fixed | `7c088138` — replaced `odds {entity} wins world cup` / `WORLDCUP-{entity:country}` with the canonical `items in {entity}` / `GROUP-{entity:category}` / entity-kind `category` text mirroring the merged generator template at `~/cli-printing-press/internal/generator/templates/teach.go.tmpl`. |
| `root_ast.go:601` tmchow skip-list w/o init hook | Reply: design as-shipped | The skip-list is intentional forward-looking framework, documented in the existing source comment at `learnHookSkipList`. v1 invokes teach/recall explicitly via cobra command paths, with no `PersistentPreRunE` recall hook today; the skip-list ships so a later auto-recall hook can consult a stable set without re-deriving it per PR. #2085's accepted shape keeps the same scaffolding posture (no ambient `PreRunE` learn-init), so the sweep emission matches fresh generator output. The doc comment in the source covers the design rationale. |
| `store_migration.go:38` tmchow hardcoded schema version | Reply: matches merged contract | `learnSchemaVersion = 3` in the sweep matches `const StoreSchemaVersion = {{if .Learn.Enabled}}3{{else}}2{{end}}` in the merged upstream template at `~/cli-printing-press/internal/generator/templates/store.go.tmpl` after #2085. The "copied generator contract" coupling is acceptable given the byte-for-byte parity contract this tool maintains; if the upstream value changes again, the parity test `TestRenderLearnPackage_ByteForByteParity` is the canary. |
| `learn_files.go:7` tmchow blocked by #2085 | Reply: unblocked | #2085 merged to main at `23b6239c`. Thread can resolve. |
| `learn_files_test.go:106` Greptile P2 parity covers only 4 of 27 | Not addressed in this batch | Out of scope of the P1 sweep. The 4 files chosen (recall/seed_test/teach/teach_log) are the load-bearing logic surface; the remaining 23 are mostly entity / lookup / pattern test scaffolding. Happy to expand if the parity contract should be tighter. |

### Verification

- `cd tools/sweep-learn-install && GO111MODULE=off go test .` — 57 tests pass (unchanged from baseline)
- `cd tools/sweep-canonical && GO111MODULE=off go test .` — 27 tests pass (no regression)
- `go vet ./tools/sweep-learn-install/...` — clean
- `gofmt -l tools/sweep-learn-install/` — clean

mvanhorn:
Closing in favor of the restructured plan at \`docs/plans/2026-05-25-001-feat-generic-recall-rerank-middleware-plan.md\` (in \`mvanhorn/cli-printing-press\`).

**Why:** This sweep tool implementation hardcodes a divergent \`search_learnings\` schema (\`store_migration.go:54\`) that doesn't match prediction-goat's canonical schema. It also doesn't extract the \`internal/store/learnings.go\` Apply engine — meaning even pilots that received the sweep can't run prediction-goat's rerank pattern. See the closing comment on #827 for the full divergence breakdown.

U7 in the restructured plan rewrites this tool to emit canonical templates (either by importing cli-printing-press as a library or embedding prediction-goat sources). The 5-pilot rollout (#827's job) then resumes via U8 + U20 on the canonical foundation.

PR #846, opened 2026-05-25
fix(tesla): regional Fleet support + route reads via Fleet when owner-api unavailable

aomartinao:
## Problem

On a 2021+ / non-NA vehicle (tested against an EU Model Y), `tesla products`, `ready`, and `vehicle_data` return HTTP 404/401 even after a successful `fleet-login`, and `fleet-register` fails with HTTP 421/412 against the wrong region.

## Root cause

- The read client uses the owner-api `base_url` + the **top-level owner-api bearer** — `config.AuthHeader()` never consults the `[fleet]` block — so reads can't use the Fleet token.
- The Fleet audience/base is **hardcoded to North America** (`fleetAPIAudience`).

Owner-api reads are gone for 2021+ cars, so reads have no working path; and partner registration + token audience target the wrong region for non-NA accounts.

## Fix

- **`fleetAPIBase()`** resolves the regional Fleet root: `TESLA_FLEET_API_URL` env → persisted `[fleet].api_base` → North America default. Used by `fleet-register` / `fleet-login` / partner-token, and **persisted** at register/login time so it's sticky.
- **`newClient()`** points the read client at the regional Fleet API with the Fleet bearer when no *usable* owner-api token exists (absent / expired / stale), with a Fleet-specific 401 auto-refresh (`makeTeslaFleetRefreshCallback`). The owner-api path is unchanged for pre-2021 cars. `TESLA_PP_FORCE_FLEET_READS=0|1` overrides the heuristic.

Files: `internal/config/config.go`, `internal/cli/tesla_fleet_auth.go`, `internal/cli/root.go`, `internal/cli/tesla_auth_autorefresh.go`. Recorded in `.printing-press-patches.json` (`tesla-regional-fleet-reads`).

## Validation

- New unit tests: `fleetAPIBase`, `teslaShouldUseFleetForReads`, `SaveFleetAPIBase`.
- `go build ./...`, `go vet ./...`, `go test ./...`, `govulncheck ./...` all green; SKILL.md verifier passes.
- Verified **live against an EU vehicle**: after `fleet-register` persisted `api_base=eu`, `tesla ready` and `tesla products` return real data with **zero env vars**, where the unpatched CLI returned HTTP 404/421/412.

## Notes

`config.go` is a generated file; this edit is recorded as a library patch per `AGENTS.md`. The same gap likely belongs upstream in the generator's Tesla template (region selection + Fleet-aware read routing) — happy to follow up there if useful.


aomartinao:
Thanks @greptileai — both addressed in the latest commit:

1. **fleet-login atomicity:** `cfg.Fleet.APIBase` is now set in memory before the single `SaveFleetTokens` write, so tokens + `api_base` persist atomically (no two-save window that could drop the region on interruption). Mirrors `fleet-register`.
2. **loopback scheme:** `normalizeFleetBase` now also requires http/https for loopback, so unsupported schemes are rejected at validation rather than failing opaquely at request time.

Build/vet/test/govulncheck/verify-skill all green.

PR #856, opened 2026-05-26
docs(instacart): align agent docs with shipped flags

skylight96:
## Summary
- align `pp-instacart` SKILL examples with the installed `instacart-pp-cli` binary name
- replace stale `--agent`/`--select` guidance with current `--json` + jq/Python filtering guidance
- update the Instacart local AGENTS guide to avoid non-shipped helpers (`agent-context`, `which`, `--no-input`)
- record the published-library docs customization in `.printing-press-patches.json`

## Test Plan
- `python3 .github/scripts/verify-skill/verify_skill.py --dir library/commerce/instacart`
- `python3 .github/scripts/verify-skill/verify_skill.py --dir library/commerce/instacart --strict`
- `go test ./...` from `library/commerce/instacart`
- `go run ./cmd/instacart-pp-cli doctor --json`
- confirmed stale probes still fail as documented: `--agent`, `--select`, `agent-context`, `which`

### DIFF
--- a/library/commerce/instacart/.printing-press-patches.json
+++ b/library/commerce/instacart/.printing-press-patches.json
@@ -1,6 +1,6 @@
 {
   "schema_version": 1,
-  "applied_at": "2026-05-14",
+  "applied_at": "2026-05-25",
   "base_run_id": "2026-04-11T06:12:25Z",
   "base_printing_press_version": "1.2.1",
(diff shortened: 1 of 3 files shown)


skylight96:
Closing this; the intended fix was local-only, not an upstream PR.

PR #905, opened 2026-05-28
feat(superhuman): add lookup and userdata write commands

mvanhorn:
Adds two commands to the Superhuman CLI, both reachable with the existing cookie-derived auth (no new auth setup):

- `lookup <email>` — live contact enrichment via /v2/profile (name, bio, location, timezone, avatar, social links, Twitter handle), with optional `--photo` download from /contact/<email>/photo. Read-only.
- `userdata write <path> <json>` — low-level escape hatch over /v3/userdata.write for paths the typed commands don't cover. Dry-run by default; `--apply` to fire; path must start with `users/`.

Also fixes a pre-existing `go vet` failure on main: two stale `ListThreadsByParticipants` test calls missing the `accountEmail` parameter.

Not included: a `share` command (the sharing write shape wasn't captured in the source traffic — deferred until a real share action can be observed) and anything depending on /messages/send (audience-gated to the iOS app token).

### Validation
- `go build` / `go vet` / `go test ./...` — green
- `govulncheck ./...` — no vulnerabilities
- `verify_skill.py` — all checks passed
- `lookup` verified live against both endpoints; `userdata write` guards covered by tests


mvanhorn:
Thanks @tmchow for merging the superhuman lookup and userdata write commands.

mvanhorn:
Appreciate you landing the superhuman lookup and userdata write commands @tmchow, that unblocks the send pipeline.

PR #908, opened 2026-05-28
feat(contact-goat): add local LinkedIn network source

jdplaymaker:
## Summary
- adds a local LinkedIn Network source for Contact GOAT backed by James/Holger CSV exports on the Mac Mini
- wires local results into coverage, prospect, and warm-intro via CLI JSON shell-out, preserving Happenstance as explicit fallback
- documents local-first usage and records the customization in .printing-press-patches.json

## Verification
- go test ./... from library/sales-and-crm/contact-goat
- python3 .github/scripts/verify-skill/verify_skill.py --dir library/sales-and-crm/contact-goat --json
- contact-goat-pp-cli coverage <known-company> --source local --limit 3 --agent
- contact-goat-pp-cli prospect partnerships --source local --limit 3 --agent
- ssh macmini LinkedIn Network unittest + stats check

### DIFF
--- a/library/sales-and-crm/contact-goat/.printing-press-patches.json
+++ b/library/sales-and-crm/contact-goat/.printing-press-patches.json
@@ -3,5 +3,21 @@
   "applied_at": "2026-05-08",
   "base_run_id": "2026-04-19T21:05:44.741872Z",
   "base_printing_press_version": "1.3.2",
-  "patches": []
+  "patches": [
+    {
+      "id": "2026-05-28-local-linkedin-network-source",
+      "kind": "code",
+      "summary": "Add a local LinkedIn Network source backed by James/Holger CSV exports on the Mac Mini. The source shells to the linkedin-network CLI JSON output, maps rows into flagshipPerson, and wires local-first results into coverage, prospect, and warm-intro without directly reading SQLite or silently spending Happenstance credits on local-source failure.",
+      "files": [
+        "internal/cli/local_linkedin_network.go",
+        "internal/cli/local_linkedin_network_test.go",
+        "internal/cli/coverage.go",
+        "internal/cli/prospect.go",
+        "internal/cli/warm_intro.go",
+        "internal/cli/flagship_helpers.go",
+        "internal/cli/source_selection.go",
+        "SKILL.md"
+      ]
+    }
+  ]
 }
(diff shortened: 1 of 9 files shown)


tmchow:
Thanks for the work here, @jdplaymaker, but I don't think this lands as written. A couple of things:

**Catalog fit.** `contact-goat` is in the public catalog and gets installed by people with no relationship to your team. This PR:

- Hardcodes a default path to one specific user's home directory on one specific Mac Mini (`/Users/gonkv2/Documents/Agentic Engineering/Playmaker/LinkedIn-Network`) and an SSH host alias literally named `macmini`. The env-var overrides help, but the default makes the code only usable by you.
- Depends on an external `linkedin-network` CLI that isn't in this repo or anywhere in the catalog, reading CSV exports ("James and Holger's first-degree LinkedIn exports") whose format is nowhere documented. Reviewers (and Greptile) can't tell what data shape we're actually consuming — only the JSON envelope your subprocess happens to emit.
- Adds `local` to the default `--source both` set on `coverage`, `prospect`, and `warm-intro`, so every vanilla invocation by every user now tries to `ssh macmini` and prints `warning: local LinkedIn Network unavailable: …` on stderr. That's a behavior change to a published CLI to support a workflow only you can run.

**Real correctness issues** (both flagged by Greptile, both legit):

1. The docs and the `scoreForRelationship` table contradict each other — `linkedin_export_1deg_both` scores 9.0 vs `happenstance_friend` at 10.0, so Happenstance friends always outrank local both-owner contacts even though the prose says the opposite.
2. In `prospect`, the local fetch fires on every invocation regardless of `--source`, unlike `coverage` which gates it on `sources[SourceFlagLocal]`.

**What I'd suggest:**

- Keep this as a personal fork — it's clearly useful for the Playmaker team and you can pin it to whatever Mac Mini you want there.
- If you want this upstream, the shape would need to be a generic "shell out to a `linkedin-network` provider" source that:
  1. Is **off by default** (excluded from `--source both`; only enabled when the user passes `--source local` or sets one of the env vars).
  2. Has no personal-infrastructure default path baked in — let `CONTACT_GOAT_LINKEDIN_NETWORK_CMD` be required, fail clean when unset rather than warning on every run.
  3. Documents the JSON envelope contract in SKILL.md so anyone else can write a compatible local provider.
  4. Fixes the two correctness items above.

Closing for now — happy to look at a revised version if you want to take another swing at the generic-provider shape.

PR #925, opened 2026-05-28
fix(wavespeed): preserve completed run output before downloads

cathrynlavery:
## Summary
Fixes three `wavespeed-pp-cli run --wait --download` issues observed after a successful paid generation:

- prints the completed prediction payload/output URLs before attempting downloads
- treats post-success download failures as stderr warnings instead of fatal errors
- avoids sending WaveSpeed API credentials to public CDN download hosts
- accepts `run <model-id> ... --download <path>` with the positional model form
- skips prediction management URLs in `urls`/`links` containers when collecting downloadable outputs

## Validation
- `make`
- `go test ./...`
- `go run github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.8 run ./...`

Note: I did not run a live paid WaveSpeed generation because `WAVESPEED_API_KEY` was not present in the environment. The regression tests use httptest servers to verify the API host still receives auth and the CDN host does not.

### DIFF
--- a/library/ai/wavespeed/.printing-press-patches.json
+++ b/library/ai/wavespeed/.printing-press-patches.json
@@ -1,6 +1,6 @@
 {
   "schema_version": 1,
-  "applied_at": "2026-05-27",
+  "applied_at": "2026-05-28",
   "base_run_id": "20260527-214905",
   "base_printing_press_version": "4.18.1",
   "patches": [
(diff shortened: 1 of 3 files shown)


cathrynlavery:
Pushed 92aa1766a to address the remaining Greptile P1: the run command now plans downloads once for the JSON envelope and reuses that exact ordered slice for the actual transfers, avoiding divergent URL→path mappings from a second map iteration.

Validation: `git diff --check`; `go test ./internal/cli`.

@greptile-apps please re-review.

PR #936, opened 2026-05-30
fix(npm): tailor the not-on-PATH fix to the user's platform and shell

tmchow:
## Summary

`npx -y @mvanhorn/printing-press-library install <cli>` runs `go install`, which drops the binary in `$(go env GOPATH)/bin` (usually `~/go/bin`) — a directory Go never adds to `PATH` itself. When it's missing, the installer printed **one Unix-flavored hint regardless of OS or shell**:

```
WARNING: installed espn-pp-cli at …/go/bin/espn-pp-cli, but its directory is not on PATH.
Add it to PATH (e.g. $(go env GOPATH)/bin, usually $HOME/go/bin) and retry…
```

That `$(go env GOPATH)/bin` is broken shell syntax on Windows (a meaningful share of installs) and imprecise for fish. The installer now detects the platform and shell and prints the exact, copy-pasteable line to fix it.

## Before → after

macOS zsh now gets a line you can paste as-is:

```
Add it to PATH (zsh):
    echo 'export PATH="$HOME/go/bin:$PATH"' >> ~/.zshrc && source ~/.zshrc
```

Windows now gets the persistent, registry-backed command instead of `$()` syntax that does nothing in PowerShell or cmd:

```
PowerShell (persistent — recommended), then open a new terminal:
    [Environment]::SetEnvironmentVariable("Path", [Environment]::GetEnvironmentVariable("Path","User") + ";C:\Users\you\go\bin", "User")
```

## What it emits per environment

| Platform / shell | Instruction |
|---|---|
| macOS zsh | `export … >> ~/.zshrc` |
| macOS bash | `~/.bash_profile` (login shells don't read `.bashrc`) |
| Linux bash | `~/.bashrc` |
| fish (any OS) | `fish_add_path` |
| Windows PowerShell / cmd | persistent `[Environment]::SetEnvironmentVariable(… "User")` + GUI fallback |
| Windows Git Bash | POSIX-translated path appended to `~/.bashrc` |
| unknown shell | generic `export` line with file guidance |

## Design decisions

- **No silent dotfile mutation.** `npx -y` is consent to run the wrapper, not to rewrite `~/.zshrc`. The fix stays a printed instruction the user pastes — never an automatic edit. Behavior is otherwise unchanged: warn, exit non-zero, don't touch the shell config.
- **Windows persistent `PATH` is registry-based, not an rc-append**, so it gets the PowerShell `SetEnvironmentVariable(… "User")` form plus a GUI fallback — and deliberately avoids `setx`, which truncates at 1024 chars and corrupts a merged User+System `PATH`.
- **Parent-shell detection on Windows is unreliable** from a Node grandchild of npx, so rather than guess PowerShell vs cmd we print the correct PowerShell form (callable from both) plus the GUI path. Git Bash is the one case we can detect (it sets `SHELL`).

The matrix lives in a new pure `pathFixInstructions()` helper, so every `(platform × shell)` combination is unit-tested rather than hand-verified.

## Tests

12 new tests — the full `pathfix` matrix (zsh, macOS vs Linux bash, fish, Windows PowerShell, Git Bash, custom GOPATH, missing bin dir) plus two `install` integration tests asserting the warning is platform-correct end-to-end. Full suite: **73 passing**. Includes the `npm/package.json` `0.1.10 → 0.1.11` bump that `verify-npm-version-bump.yml` requires for `npm/**` changes.

---

Claude Code

### DIFF
--- a/npm/CHANGELOG.md
+++ b/npm/CHANGELOG.md
@@ -4,6 +4,14 @@
 
 - When `go install` writes a CLI to a directory that isn't on `PATH`, print the exact, copy-pasteable fix for the detected platform and shell instead of a single Unix-flavored hint. macOS zsh gets a `~/.zshrc` line, macOS bash gets `~/.bash_profile` (login shells don't read `.bashrc`), Linux bash gets `~/.bashrc`, fish gets `fish_add_path`, Windows gets the persistent PowerShell `[Environment]::SetEnvironmentVariable(... "User")` command plus a GUI fallback (and never the truncating `setx` footgun), and Git Bash gets a POSIX-translated path. The previous message printed `$(go env GOPATH)/bin` shell syntax that was wrong on Windows and imprecise on fish.
 
+## 0.1.10
+
+- Fix `install` / `uninstall` failing at the skill step on Windows with `ENOENT`. The process runner invoked `npx` through `node:child_process.execFile`, which does not resolve the `npx.cmd` shim Windows requires; it now uses `cross-spawn`, so the `npx skills …` install/remove step runs on Windows the same as on macOS and Linux. (#864)
+
+## 0.1.9
+
+- Catalog hints now match how the tool was invoked. The `list` and `search` usage lines and each entry's `install:` hint previously hardcoded the long `npx -y @mvanhorn/printing-press-library …` prefix; they now detect an `npx` run versus a globally-installed `printing-press-library` binary and print the shorter `printing-press-library …` form when that's how you're running it. (#749)
+
 ## 0.1.8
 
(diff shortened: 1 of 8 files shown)


tmchow:
Both Greptile findings addressed:

- **Issue 1 — Windows test leaked the host `SHELL`.** Real bug: on Linux CI (`SHELL=/bin/bash`), `detectShell("win32","/bin/bash")` returns `"gitbash"`, so the `SetEnvironmentVariable` assertion would fail. The suggested `shell: undefined` pin was accepted in d29472fd4. Verified green locally under `SHELL=/bin/bash npm test` (73 passing), which reproduces the CI condition.
- **Issue 2 — missing 0.1.9 / 0.1.10 changelog entries.** Backfilled with real entries reconstructed from history rather than placeholders: 0.1.9 = #749 (invocation-aware catalog hints, npx vs global binary), 0.1.10 = #864 (`cross-spawn` fix for `npx` `ENOENT` on Windows skill steps). Changelog is now a complete `0.1.0 → 0.1.11` history.

PR #944, opened 2026-05-30
feat(npm): add reinstall command as an alias for update

tmchow:
## Summary

`reinstall` is now a command on the installer. Before this, the only ways to force-refresh an installed CLI were to know that `install <name>` happens to overwrite in place, or that `update` is mechanically a reinstall — neither is the verb you reach for when a binary or skill has gone stale. `reinstall <name>` rebuilds one CLI's binary from the latest catalog code and re-adds its skill; `reinstall` with no name does the same for every Printing Press CLI already on your `PATH`.

It routes to the existing `update` handler instead of duplicating logic, because the operation is genuinely identical — both run `go install …@latest` and re-add the skill. As part of that, the shared arg-parser's `Unknown update option` error becomes the command-agnostic `Unknown option`, so it reads correctly whether the user typed `update` or `reinstall`.

| Command | Effect |
|---|---|
| `reinstall <name>` | Rebuild one CLI's binary + re-add its skill (to latest) |
| `reinstall` | Same, for every Printing Press CLI found on `PATH` |

## Verification

`npm test` passes 75/75 (includes two new tests: alias routes to the update handler rather than the unknown-command path, and `--help` lists the command). Built-binary smoke test:

```
$ printing-press-library --help
  …
  update [name]             Refresh one installed CLI, or all installed CLIs
  reinstall [name]          Reinstall one CLI (or all installed); alias for update
  …

$ printing-press-library reinstall --bogus
Unknown option: --bogus        # routed to update parser, not "Unknown command"; exit 1
```

## Release note

Bumps the npm package to `0.1.12` (required for `npm/src/**` changes). On merge to `main` this auto-tags and publishes `0.1.12` via the OIDC pipeline — no CLI/catalog/skill changes are involved.

---

Claude Code

### DIFF
--- a/npm/CHANGELOG.md
+++ b/npm/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 0.1.12
 
-- Add a `reinstall` command as an alias for `update`. `reinstall <name>` rebuilds one CLI's binary from the latest catalog code and re-adds its skill; `reinstall` with no name does the same for every Printing Press CLI already on `PATH`. The mechanics are identical to `update` (both run `go install …@latest` and re-add the skill) — this just exposes the verb users reach for when a binary or skill needs a clean refresh.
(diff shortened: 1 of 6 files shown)


tmchow:
Addressed the P2 (verb/message mismatch) in fd1d2be78: the shared "nothing found on PATH" message is now verb-neutral — `No Printing Press CLIs found on PATH to refresh.` — so it reads correctly whether invoked as `update` or `reinstall`. No test asserted the old string; 75/75 still green.

PR #996, opened 2026-06-01
fix(verify-attribution): tolerate leading BOM/comment when reading SKILL.md author

cathrynlavery:
## Summary

Hardens the attribution guard so it reads the SKILL.md `author:` correctly even when the file begins with a UTF-8 BOM or a leading HTML comment. **This is a prerequisite for a follow-up fix** (#995) that removes such leading bytes from two published `SKILL.md` files — without this change first on `main`, that follow-up cannot pass CI.

## The problem

`skill_author()` parses frontmatter with a strict `\A---` match. If a SKILL.md starts with a BOM or an HTML comment, the match fails and the author reads as `None`. So when a PR removes those leading bytes, the guard sees `None -> "Real Name"` and reports a **phantom attribution change** — even though the author value never changed. Paired with the (legitimate) surface change to the same file, that trips the attribution-vs-surface gate and blocks the fix.

This is the same root cause as a real install bug: `skills add` also requires `---` at the very top, so those files fail to install today. The guard shouldn't be blind to leading noise either.

## Why this must be a separate, first PR

`verify-library-conventions.yml` deliberately overlays the **base branch** copy of `.github/scripts/verify-attribution/` before running it (so a PR can't ship its own modified verifier). That means a guard fix only takes effect once it is on `main`. Bundling it with the SKILL.md change leaves CI running the old guard against the changed files, which always fails. Hence: land the guard fix here first, then the SKILL.md fix in #995 goes green.

## Change

- `strip_leading_noise()` removes a leading BOM and any leading HTML-comment blocks before frontmatter parsing; `skill_author()` uses it. Surface-change detection is unchanged.
- Added unit tests covering BOM-prefixed, comment-prefixed, plain, and no-frontmatter inputs.

## Validation

- `python3 -m unittest verify_attribution_test` → 7 passed (4 new)
- No `library/**` or manifest changes in this PR, so the guard itself reports nothing to gate.

### DIFF
--- a/.github/scripts/verify-attribution/verify_attribution.py
+++ b/.github/scripts/verify-attribution/verify_attribution.py
@@ -150,6 +150,10 @@ def skill_author(text: str | None) -> str | None:
 def normalize_skill_author(text: str | None) -> str | None:
     if text is None:
         return None
+    # Strip the same leading noise as skill_author so that adding/removing a BOM
+    # or leading comment is not itself read as a surface change, and so an
+    # attribution-only correction on such a file doesn't trip the surface gate.
+    text = strip_leading_noise(text)
     fm = FRONTMATTER_RE.match(text)
     if not fm:
         return text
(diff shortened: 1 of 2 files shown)


cathrynlavery:
Addressed both findings in 8604388:

1. **`normalize_skill_author` noise-stripping** — now calls `strip_leading_noise` too, so adding/removing a BOM or leading comment is not itself read as a surface change, and an attribution-only correction on such a file no longer trips the surface gate. (For the two files in #995 this means the change now registers as *neither* an attribution nor a surface change.)
2. **Test coverage** — added cases for BOM+comment combined, multiple consecutive leading comments, and two `normalize_skill_author` cases (attribution-only correction on a BOM file, and BOM-removal alone). Suite is now 11 tests, all passing.

Greptile will re-review on the new commit.

PR #1022, opened 2026-06-03
fix(granola): transcript auth in headless environments

jeffreydebolt:
## Summary

Fixes Granola transcript/live API access for headless or agent environments where the desktop encrypted store can block on macOS Keychain and the local cache may incorrectly make transcripts look unavailable.

This came up using `granola-pp-cli` from an OpenClaw agent session: the CLI could find the meeting records, but local cache reported no transcript data. The transcripts did exist via Granola's live API, but the live path was blocked by a combination of Keychain access and stale refresh/client behavior.

## What changed

- Add current Granola desktop headers to internal API calls:
  - `X-Client-Version: 7.299.0`
  - `X-Granola-Platform: darwin`
  - updated `User-Agent`
- Refresh expired tokens through Granola's current refresh endpoint:
  - `POST https://api.granola.ai/v1/refresh-access-token`
  - instead of the legacy direct WorkOS client-id flow, which returned `invalid_client` in this environment
- If `supabase.json.enc` decrypt fails because Keychain is unavailable, fall back to plaintext `supabase.json` when present
  - other encrypted-store errors still surface rather than silently falling back
- Add a 15s timeout around the macOS `/usr/bin/security find-generic-password` call so headless runs fail clearly instead of hanging indefinitely
- Update refresh tests for the Granola refresh request shape

## Why

In headless/agent workflows, the current behavior can create false negatives or indefinite hangs:

1. Local cache says `transcript_available = 0` / no segments.
2. Live transcript fetch tries to decrypt Granola safe storage.
3. macOS Keychain prompt can block indefinitely without a human at the desktop.
4. Existing refresh fallback can fail with `invalid_client` against current Granola sessions.

The live transcript endpoint works once the current refresh path and headers are used.

## Verification

From `library/productivity/granola`:

```sh
GRANOLA_SUPPORT_DIR=$(mktemp -d) go test ./internal/granola ./internal/granola/safestorage ./internal/cli
```

Passed.

Live smoke test against a document that previously failed from the CLI:

```sh
go run ./cmd/granola-pp-cli --no-refresh --data-source live transcript get 5d26f463-928a-4940-9f59-c7770b9709e8 --format json --json
```

Result:

- `source: live`
- `segments: 1083`
- first segment text: `Hey.`

### DIFF
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -108,7 +108,7 @@ jobs:
       - uses: actions/setup-go@v6
         if: steps.modules.outputs.count != '0'
         with:
-          go-version: '1.26.3'
+          go-version: '1.26.4'
           # This workflow may scan several unrelated standalone modules. There
           # is no single dependency file that produces a useful setup-go cache.
           cache: false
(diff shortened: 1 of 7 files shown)


jeffreydebolt:
Addressed the Greptile review items in follow-up commits:\n\n- Removed the Authorization header from the refresh request; verified live refresh still works without it.\n- Updated the stale WorkOSClientID comment.\n- Preserved plaintext fallback parse errors instead of hiding them behind the encrypted-store Keychain error.\n- Moved the Keychain timeout check under the command error branch to avoid discarding valid output in the deadline race.\n- Tightened the D6 safeguard: if supabase.json.enc exists but Keychain is unavailable, plaintext supabase.json fallback is treated as desktop-owned and refresh is refused to avoid consuming Granola Desktop's single-use refresh token. Explicit env override remains the opt-in refresh path for headless agents.\n\nVerification after the changes:\n\n`GRANOLA_SUPPORT_DIR=$(mktemp -d) go test ./internal/granola ./internal/granola/safestorage ./internal/cli`\n\nLive smoke with explicit env override still fetched the prior failing transcript: source=live, segments=1083.

jeffreydebolt:
Thanks!

Issue #1125, opened 2026-06-09
Write-side Amazon CLI (cart + checkout) — is a Node/Playwright checkout helper acceptable in a library entry?


sophiealula:
I have a working write-side Amazon buyer CLI I'd like to contribute: `amazon-pp-cli` — add-to-cart, cart inspection, and full checkout, complementing the read-side `amazon-orders` already in commerce.

Source (currently standalone): https://github.com/sophiealula/pp-amazon-skill

Quick shape:
- Originally printed with cli-printing-press (May 2026), then hand-evolved through a few weeks of real daily use. `go build` / `go vet` / tests all green.
- **Repurchase-only by design**: `add` refuses anything that doesn't match the user's purchase history (local SQLite mirror, FTS match with strict/loose quality reporting). That's the safety feature — an agent can't impulse-buy.
- Confirmation gates everywhere: `--dry-run` on every mutating command, checkout requires `--yes`, loose matches require `--allow-loose`, CAPTCHA/sign-in gates exit 9 with a deeplink for manual handoff.
- Multi-account via `--profile`, cookie auth (same approach as amazon-orders), MCP server binary included.

Before I do the packaging work, three questions:

1. **The main one:** Amazon's checkout endpoint rejects static-POST automation, so cart writes and checkout run through a bundled Playwright helper (`checkout-helper/amazon-checkout.mjs`, needs Node 22 + Chromium). Every other library entry I've looked at is pure Go. Is a Node helper acceptable as part of an entry, or is that a hard no for the library?

2. **Slug:** `amazon` is unclaimed, but given `amazon-orders` / `amazon-seller` / `amazon-ads` exist, would you prefer something function-scoped like `amazon-cart`?

3. **Flow:** since it was originally pressed but has drifted from press output (no current `.printing-press.json`), is `/printing-press import` the right path back into library format, or `publish` after regenerating the manifest?

Happy to adjust whatever's needed — the repurchase-only + confirmation-gate design came out of real incidents and I think it's a good fit for what the library is going for.

sophiealula:
Went ahead and opened #1126 with the packaged entry so there's something concrete to look at — the Node-helper and slug questions from this issue are restated there. Validation and packaging done with printing-press 4.9.0, all checks green.

PR #1126, opened 2026-06-09
feat(amazon): add amazon-pp-cli — history-first repurchases with gated checkout

sophiealula:
Closes the question side of #1125 with a concrete PR.

## What this is

The write-side complement to `amazon-orders`: add-to-cart, cart inspection, and full checkout against amazon.com. The defining property is the **repurchase-only gate** — `add` refuses to write to the cart unless the query matches the user's own purchase history (local SQLite FTS mirror), so an agent holding this CLI cannot impulse-buy. Loose matches require an explicit `--allow-loose`, checkout exits 10 without `--yes`, and every mutating command supports `--dry-run`.

Provenance: originally printed with cli-printing-press on 2026-05-19 (run `20260519-122542` — manuscripts ride along under `.manuscripts/`), then hand-evolved through three weeks of real-account daily use. The hand-evolution is documented in `.printing-press-patches.json`.

## Validation

`printing-press publish validate` (4.9.0): all 11 checks pass — manifest, transcendence, phase5, go mod tidy, govulncheck, go vet, go build, --help, --version, verify-skill, manuscripts. Staged via `publish package`, which also ran the vendor-secret and PII audits clean. `go` directive is 1.26.4 to clear govulncheck's stdlib findings.

## The one structural question (from #1125)

Cart writes and checkout run through a bundled Playwright helper (`checkout-helper/amazon-checkout.mjs`, Node 22 + Chromium) because Amazon's checkout endpoints reject static-POST automation. Everything else — history, search, auth, doctor — is pure Go. If a Node component is a hard no for the library, happy to discuss splitting the checkout surface out; the read/add surfaces stand alone.

Also flexible on the slug: `amazon` matches the binary (`amazon-pp-cli`), but if you'd rather have `amazon-cart` next to `amazon-orders`/`amazon-seller`/`amazon-ads`, I'll rename via `publish rename`.

Per CONTRIBUTING, `registry.json` and `cli-skills/` are untouched.


sophiealula:
All three review findings addressed in a92125d:

- **client.go signin classification**: now requires the final request URL to be `/ap/signin` (redirect case) or an inline `action="/ap/signin"` form — a bare `/ap/signin` substring in nav JS no longer fires `ErrAuthExpired`.
- **parse.go SPC tokens**: hidden inputs matched per-tag with separate attribute regexes, so attribute order doesn't matter; regression test added covering name-first and value-first orders.
- **checkout.go**: dead `client.DryRun = false` removed.

`go build` / `go vet` / tests green.

tmchow:
@sophiealula please address this Greptile comment

> Safe to merge with one follow-up: the browser-helper path resolution is missing its documented dev-mode candidate, which will trip up contributors building from source without the env-var override.
> 
> The core safety invariants — repurchase-only gate, --yes confirmation, per-profile isolation, most-frequent tiebreak — are all correctly implemented. The previously-flagged auth-classification operator-precedence bug and the SPC token attribute-ordering bug are both fixed in this version. The one active defect is in browserHelperPath: the docstring promises a third search location for dev builds that was never added to the candidates slice, so any contributor running a locally-compiled binary without AMAZON_PP_BROWSER_HELPER set hits an opaque failure for every browser-backed command.

tmchow:
Thanks for the contribution! This PR doesn't look like it came through the Printing Press publish process. Its description is missing the standard details that process produces, which we rely on to review and accept library submissions. We're going to close this one. Please re-submit with a fresh PR created via the publish process and we'll take another look. Thanks again, and no worries.

(This is an automated triage message. If you think it was closed in error, reply here and a maintainer will take a look.)

PR #1132, opened 2026-06-10
feat(kalshi): 2026.6.2 read-path correctness release

Ministry99:
## kalshi

⚠️ **Replaces existing `kalshi`** — 2026.6.2 read-path correctness release (system-audit findings; same CLI, no order-placement changes).

Trade prediction markets, persist tick data, and answer category-level P&L questions Kalshi.com cannot.

**API:** kalshi | **Category:** payments | **Press version:** generated 3.9.0, validated under 4.24.0
**Spec:** bundled spec.yaml (Kalshi Trade Manual)

### What changed (2026.6.2)

Read-path correctness release (system audit 2026-06-09). No order-placement changes; RSA-PSS signing, rate limiting, and raw-JSON passthrough untouched.

- FIXED (critical): `--all` returned `{"results": null}` on every paginated read (fills/positions/settlements/orders/markets/events). The paginator now understands Kalshi's resource-named envelopes (with a single-array-key fallback), follows the `cursor` field by default, guards against sticky cursors, and emits `[]` instead of `null` for empty sets.
- FIXED (critical): the 5-minute disk cache no longer applies to `/portfolio/*` reads — a cached pre-trade response was indistinguishable from "the order didn't fill" during fill verification. Cache files are now 0600 (account data) and the cache key is order-stable.
- FIXED: `resources` table PK is now `(resource_type, id)` (schema v2, automatic rebuild migration). A market and a settlement sharing a ticker previously overwrote each other, corrupting winrate/attribution/heatmap. Rows clobbered before the upgrade are restored by the next `sync`.
- FIXED: `portfolio attribution --period/--since` now filters on each settlement's `settled_time` (was: `synced_at`, which every re-sync re-stamped — "last 7 days" returned all-time data). `winrate --since/--category`, `movers --category`, and `exposure --by/--warn-threshold` are now actually wired (previously parsed and discarded).
- ADDED: `portfolio-positions` sync resource (`/portfolio/positions`). `portfolio exposure`, `calendar`, and `stale` previously queried rows that were never synced and always reported an empty book; their queries also now derive side from the position sign, prefer `*_dollars` fields (legacy integer `market_exposure` is centi-cents), and fix an AND/OR join-precedence bug.
- FIXED: auth-failure hints, `doctor`, `agent-context`, and the MCP manifests now name the env vars the CLI actually reads (`KALSHI_API_KEY`, `KALSHI_PRIVATE_KEY_PATH`, `KALSHI_PRIVATE_KEY`); the previous `KALSHI_TRADE_MANUAL_KALSHI_ACCESS_KEY` was read by nothing. `doctor` validates credentials with a SIGNED `GET /portfolio/balance` and explicitly diagnoses `INCORRECT_API_KEY_SIGNATURE` (key-id/.pem mismatch); MCP desktop installs now configure working auth.
- FIXED: incremental sync sends `min_ts` (Unix seconds) instead of the unrecognized `since` param, so incremental syncs stop re-fetching full history on endpoints that support it.
- FIXED: HTTP 409 on a GET surfaces as an error instead of a silent empty success (CLI + MCP); the interactive "N results" provenance line counts envelope responses correctly (was always 0); `User-Agent` now reports the real release version.

### What This CLI Does

**Trade prediction markets, persist tick data, and answer category-level P&L questions Kalshi.com cannot.**

Every Kalshi feature, plus a local SQLite store that records market prices on every sync — so movers, correlations, and price history work offline. Honest about read-only vs read/write keys, with --dry-run on every mutator and a global --read-only safe-mode lock for newsroom and bot deployments.

### Manuscripts

- [Research Brief](library/payments/kalshi/.manuscripts/20260504-204550/research/)
- [Shipcheck Results](library/payments/kalshi/.manuscripts/20260504-204550/proofs/)

### Validation Results

| Check | Result |
|-------|--------|
| Manifest | PASS |
| Phase 5 | PASS |
| go mod tidy | PASS |
| govulncheck (this CLI only, reachable findings) | PASS (toolchain pinned go1.26.4 for GO-2026-5037/5039) |
| go vet | PASS |
| go build | PASS |
| --help | PASS |
| --version | PASS |
| verify-skill | PASS |
| Manuscripts | PRESENT |

Regression tests added: paginator (envelope/cursor/sticky/empty), composite-PK migration (fresh/v1-upgrade/idempotence), 409-on-GET, since-param conversion. Full suite green; `TestMigrate_ConcurrentFreshDB` is flaky on Windows runners (pre-existing, fails identically on the pristine 2026.6.1 tree).


tmchow:
Hey @Ministry99, this PR has failing CI and unresolved review feedback. Mind taking another pass? If you've hit a blocker or want to hand it off, just leave a note. Otherwise I'll close it for inactivity in about 5 days, happy to reopen anytime.

Ministry99:
Thanks for the nudge, @tmchow — pushed a fix that should clear both CI failures and the open review threads:

- **Verify (generated artifacts):** dropped `cli-skills/pp-kalshi/SKILL.md` from the PR; it's regenerated post-merge by `generate-skills.yml`.
- **MCPB manifest contract:** `.printing-press.json` `auth_env_vars` now lists the env vars the manifest and CLI actually read (`KALSHI_API_KEY`, `KALSHI_PRIVATE_KEY_PATH`) instead of the stale `KALSHI_TRADE_MANUAL_KALSHI_ACCESS_KEY` that `verify_manifest.py` flagged.
- **store.go (Greptile):** moved the `updated_at` guarantee into `backfillColumns`, which runs before the composite-PK rebuild. The earlier `COALESCE((SELECT updated_at … WHERE 1=0), …)` still names `updated_at`, which SQLite resolves at prepare time, so it still aborted with "no such column" on stores predating the column. Added a `NoUpdatedAt` regression test that fails on the prior revision and passes now. The column is backfilled without a DEFAULT (SQLite rejects a non-constant default on `ALTER … ADD COLUMN`); the rebuild's `COALESCE` stamps the timestamp.
- **helpers.go (Greptile):** 409-on-GET is now classified via `errors.As` on `*client.APIError.Method`, robust to wrapped errors, instead of the `"GET "` message prefix.

Build, `go vet`, and the store/cli suites pass locally (the pre-existing Windows-only `TestMigrate_ConcurrentFreshDB` SQLITE_BUSY flake aside).

tmchow:
Thanks for the updates. CI is green and Greptile is resolved now, but I would hold this before maintainer review for two catalog-shape issues.

This PR changes generated CLI code under `library/payments/kalshi/`, so the published-library convention requires a schema v2 patch metadata entry under `library/payments/kalshi/.printing-press-patches/` describing the behavior this repair preserves across reprints.

I also see `library/payments/kalshi/dogfood-results.json` and `library/payments/kalshi/workflow-verify-report.json` deleted in the diff. Please preserve those artifacts, or explain why this repair intentionally removes them.

After that, this should be ready for maintainer review.

Ministry99:
Thanks @tmchow — both addressed in `ecc04a2a`:

**Patch metadata.** Added schema-v2 entries under `library/payments/kalshi/.printing-press-patches/`, one per durable reprint-guard this PR introduces (following the `689-bug1/2/4` worked example):
- `kalshi-list-endpoints-return-resource-named-envelopes` — Kalshi's `{<resource>:[...],"cursor":...}` list shape + cursor following (and 409-on-GET surfacing).
- `kalshi-resources-need-composite-resource-type-id-pk` — a market and its settlement can share a ticker, so the store keys on `(resource_type, id)`.
- `kalshi-portfolio-reads-must-bypass-cache` — `/portfolio/*` must skip the response cache so fill verification can't read a stale pre-trade balance.

The existing `689-bug1/2/4` entries on `main` cover the resource-ID, settlements, and analytics-SQL customizations; `_meta.json`'s deferred bug3/bug5 are unchanged.

**Deleted artifacts.** Not intentional — restored `dogfood-results.json` and `workflow-verify-report.json` to their pre-PR state; the diff no longer removes them.

Ready for maintainer review.

PR #1149, opened 2026-06-11
fix(lemonsqueezy): clarify catalog write limits

cathrynlavery:
## lemonsqueezy

Fixes Lemon Squeezy catalog workflows so the printed CLI reflects the public API's actual write surface: catalog resources are dashboard-only, while checkout creation is supported once a variant exists.

### What changed

- Added a `capabilities` command and agent-context capability metadata for create/update/delete support by resource.
- Made `import products`, `import variants`, and `import files` refuse with structured `resource_read_only` / `operation_not_supported_by_public_api` errors instead of fake POST dry-runs.
- Taught `which "create product"` and `which "upload file"` to return dashboard-handoff guidance.
- Added `dashboard handoff product` to generate exact dashboard fields, optional file checksum metadata, affiliate settings, and verification commands.
- Reworked `checkouts create` into a safe checkout workflow with store/variant flags, dry-run JSON:API payload output, live preflight validation, and checkout ID/URL extraction.
- Updated README/SKILL guidance and recorded the library-side patch.

### Validation Results

- PASS — `go test ./...`
- PASS — `go build ./...`
- PASS — `go vet ./...`
- PASS — `python3 .github/scripts/verify-skill/verify_skill.py --dir library/payments/lemonsqueezy`
- PASS — smoke: `import products|variants|files --dry-run --agent` emits `resource_read_only operation_not_supported_by_public_api`
- PASS — smoke: `which "upload file" --json` returns `dashboard handoff product`
- PASS — smoke: `agent-context --pretty` exposes read-only catalog capabilities
- PASS — smoke: `checkouts create --store-id 12345 --variant-id 67890 --redirect-url ... --dry-run --json` prints the JSON:API request body without sending


cathrynlavery:
review

PR #1152, opened 2026-06-12
fix(instagram): pull persisted zero for every per-media metric

amkimura2011:
## Problem

The `instagram` CLI's `pull` command stored **0 for every per-media metric** (reach, saved, shares, views, Reels watch-time). Only `like_count`/`comments_count` (public fields straight off the media object) showed real values, which masked the bug — the analytics commands (`top-posts`, `formats`, `compare`-by-post) all ranked on zeros.

## Root cause

`pullBrandMedia` fetches per-media insights via `fetchInsightTotals`, which always calls with:

```
metric_type=total_value
period=day
```

and parses only `data[].total_value.value`. That is the correct shape for **account** insights (`/{ig-user-id}/insights`), but Graph API v22 **media** insights (`/{ig-media-id}/insights`) return the classic lifetime shape:

```json
{ "data": [ { "name": "reach", "period": "lifetime", "values": [ { "value": 157 } ] } ] }
```

There is no `total_value` field, so the parser read empty -> `0` for every media metric.

## Fix

Add `fetchMediaInsightTotals` — GETs with `metric` only (no `metric_type`/`period`) and sums `values[].value` per metric — and point `pullBrandMedia` at it. Account insights path (`fetchInsightTotals`) is untouched, so `compare`/`growth`/`account-insights` keep working.

## Verification

Built clean and validated against a live Creator account. After the fix, `top-posts`/`formats` match the native Instagram app exactly — e.g. one Reel: app **157 reach / 180 views / 7s avg watch** vs CLI **reach 157 / views 180 / ig_reels_avg_watch_time 7095 ms**. 3/3 exact across the metrics checked.

Generated with Claude Code.


tmchow:
Thanks for the PR. I cannot route this for maintainer review yet.

Two blockers remain:

- Greptile helper reports 1 unresolved review thread in `library/social-and-messaging/instagram/internal/cli/instagram_brands.go`.
- This changes generated CLI code under `library/social-and-messaging/instagram/`, so the published-library convention requires a schema v2 patch metadata entry under `library/social-and-messaging/instagram/.printing-press-patches/`.

Please resolve or reply to the Greptile thread and add a patch record describing the media-insights parsing behavior this change preserves across reprints.

tmchow:

tmchow:

- Greptile score is 4/5; 5/5 is required for maintainer-ready triage

tmchow:
Closing this. Quiet since the July follow-up asking for Greptile 5/5. Fresh PR welcome once that finding is fixed.

PR #1063, opened 2026-06-05
fix(airbnb): persist scrape results, honest watch-check, watch remove, scraper-aware sync

bobeglz:
Four fixes to `airbnb-pp-cli` from a real install + shakedown (2026-06-05), all confined to `library/travel/airbnb/`:

1. **Scrape results now persist to the store.** `search`, `get`, `cheapest`, `compare`, and `plan` upsert listings/hosts and write price snapshots (guarded on `price > 0`). Previously only `watchlist` persisted — the "local state that compounds" promise was empty for every scrape path.
2. **`watch check` no longer false-hits on null price.** A 0/null price made `0 <= max-price` always true → `exit 7 "price drop hit"` unconditionally (verified: max-price 1000 MXN vs a real ~16.8k MXN total still "hit"). Non-positive prices now route to a `no_price` bucket with exit 0; exit 7 fires only on a real drop.
3. **`watch remove <url|bare-id>`.** The watchlist could be added/listed/checked but not cleaned without raw SQLite.
4. **`sync` rewired for a scraper.** Default `sync` used the API-mirror template against a wishlist endpoint → HTTP 404. It now re-scrapes what the store knows (watchlist + known listings) and persists fresh snapshots; empty store → honest `empty_store` warning with zero network; `--resources airbnb_wishlist` stays auth-gated (exit 4 + `auth login --chrome` hint); explicitly-named legacy resources keep the API-mirror path.

## Verification

- `go build` / `go vet` / `go test -count=1 ./...` green (incl. `-race` on `internal/cli`); 8 new unit tests.
- Old-vs-new binary A/B: command surface identical except added `watch remove`; `cheapest`/`compare` dry-run outputs byte-identical (persistence is a pure store side effect).
- Migration safety: zero schema changes (`schema_version` stays 2; `store.go` untouched); an old-schema DB with a seeded watchlist row survives intact through all new commands.
- Isolated live smokes (sandbox HOME/XDG, `--rate-limit 0.3`): search persisted 18/18 listings; watch add→remove→list-empty; empty-store sync exits 0 with zero network; seeded sync re-scraped exactly 1 known listing.
- Honest gap: no live listing exposed a positive bookable total during smokes, so snapshot-write-on-positive-price is proven by unit tests + schema round-trip (the guard correctly refuses $0 rows).

## Patches

Four `.printing-press-patches/*.json` entries (schema_version 2) record the durable behavioral contracts for future reprints.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1165, opened 2026-06-13
fix(airbnb): persist scrape results, honest watch-check, watch remove, scraper-aware sync

bobeglz:
Continues #1063 — that PR auto-closed after five quiet days, and GitHub won't let me reopen it (HTTP 422, "branch was force-pushed or recreated") now that I've rebased the branch. Same branch, same work, rebased onto current `main` with the review feedback addressed.

## What this does

Four behavioral fixes to `airbnb-pp-cli`, all scoped to `library/travel/airbnb/`, no schema changes:

- **Scrape persistence** — `airbnb_listing search`/`get`, `cheapest`, `compare`, `plan`, and `watch check` persist the listing + host into the local store and write a dated price snapshot **only when a real positive price** was scraped (no phantom $0 rows). Best-effort: a store error never alters the live result.
- **Honest `watch check`** — `classifyWatchPrice` reserves exit 7 for a real positive price at/below threshold. A listing whose SSR price is unavailable for the dates lands in a `no_price[]` bucket instead of firing a false drop on every cron run.
- **`watch remove <url-or-id>`** — the missing watchlist mutator, backed by `store.DeleteWatchlistItem` (DELETE by the UNIQUE `listing_url`; exit 3 when absent).
- **Scraper-aware `sync`** — Airbnb has no public collection to page, so default `sync` re-scrapes what the store already knows (watchlist + previously-scraped listings) and reports `empty_store` honestly with no network call. The auth-gated wishlist sync stays selectable via `--resources airbnb_wishlist`.

## Review feedback addressed

Both Greptile findings on the prior PR (`scrape_sync.go`) are fixed:

1. **Machine-mode stream purity** — the human-readable `Sync complete: …` line is now gated on `humanFriendly`, so machine mode emits only JSON events (fixed on both the scrape-sync and wishlist paths).
2. **No silent drop** — a `db.List` failure is surfaced as a `sync_warning` (`listing_read_failed`) and keeps the watchlist targets, instead of silently halving sync coverage.

Per AGENTS.md: customizations are recorded in `.printing-press-patches/` (directory form), the release ledger (`CHANGELOG.md` / `.printing-press-release.json` / runtime `version`) is left to the post-merge workflow, and I added myself to `contributors`.

## Validation

| Check | Result |
|---|---|
| `go build ./...` | ✅ |
| `go vet ./...` | ✅ |
| `go test ./...` | ✅ (new unit tests: persist, classifyWatchPrice, collectScrapeTargets, DeleteWatchlistItem) |
| `verify_skill.py --dir library/travel/airbnb/` | ✅ all checks |
| `govulncheck ./...` | ✅ no vulnerabilities |

`go.mod` / `go.sum` unchanged from `main`.


tmchow:
@bobeglz address Greptile feedback and babysit it to 5/5 please

PR #1176, opened 2026-06-15
fix(instagram): record per-media insight values instead of zeros

kjekkmec:
## Summary
`instagram-pp-cli pull` silently records `0` for every per-media insight (`reach`, `views`, `saved`, `shares`, `total_interactions`) in `ig_brand_media`, even when the account has real data. Likes/comments are unaffected. This makes `top-posts`, `formats`, `best-time`, and every reach/engagement analytic read all-zero.

## Root cause
`fetchInsightTotals` is shared by account-level and per-media insight collection but parses only the `total_value` response shape:

- **Account** insights (`/{ig-user-id}/insights`) honor `metric_type=total_value` and return a scalar:
  `{"data":[{"name":"reach","total_value":{"value":555}}]}` ✓
- **Per-media** insights (`/{media-id}/insights`) **ignore `metric_type`** and return the time-series shape:
  `{"data":[{"name":"reach","period":"lifetime","values":[{"value":211}]}]}` ✗

The media response carries no `total_value` field, so the parser extracts nothing and writes `0` for every metric. The call returns **HTTP 200**, so the existing 400-tolerance path is never involved.

## Evidence (real account, 25 posts)
- Before: every post stored `reach/views/saved/shares/total_interactions = 0` (likes/comments correct).
- A direct Graph API call for one of those same posts returns `reach=211, views=330` in the `values[]` shape.

## Fix
Extract parsing into `parseInsightTotals` and handle both shapes — prefer `total_value`, fall back to the latest `values[]` entry (lifetime media metrics return a single entry). Request params are unchanged; `period=day` is harmless since the API ignores it for media.

## Tests
New `instagram_brands_insights_test.go`:
- media `values[]` shape → real values (**fails on the pre-fix parser**: reach=0)
- account `total_value` shape → unchanged
- present-but-zero `total_value` → 0 (precedence over a stray `values[]`)

`go build ./...`, `go vet ./internal/cli/`, and the full `internal/cli` suite pass.

## Scope
- `internal/cli/instagram_brands.go` is the hand-written collector (`// pp:data-source live`, no `DO NOT EDIT` banner), so the fix lives here rather than in generated output.
- No generated artifacts (`registry.json`, `cli-skills/pp-*/SKILL.md`) are touched.


tmchow:
@kjekkmec please address Greptile feedback

tmchow:
Adding one more triage note: because this changes generated CLI code under `library/social-and-messaging/instagram/`, the published-library convention also requires patch metadata.

Please add a schema v2 patch file under `library/social-and-messaging/instagram/.printing-press-patches/` describing the per-media insights fix. Once the Greptile feedback is resolved and the patch metadata is present, this can be re-triaged for maintainer review.

serio000:
Independent confirmation of this bug and the fix approach, from live testing on `v2026.6.1`.

**Reproduced:** after `brands add` + `pull`, `top-posts --metric reach` returned `0` for every post — and crucially with *no* `fetch_failure`, because the tolerated-400 path masks it. Root cause is exactly as this PR describes: the request sends `metric_type=total_value`, but per-media insights ignore it and return the lifetime `values[]` shape with `total_value` absent, so the old total_value-only parse produced 0 for every media metric.

**Verified end-to-end against a real IG Business account:**
- Direct Graph call `GET /{media-id}/insights?metric=reach` returns a real figure (e.g. `65702`).
- With the `values[]`-fallback parse applied, `top-posts --metric reach` reports the same figure (exact match), and `formats` aggregates non-zero reach across REELS/FEED.
- Confirmed both response shapes in the wild: account-level `/{ig-user-id}/insights` keeps `total_value` (so preferring `total_value` must stay), while media-level returns `values[]` with `total_value: null`.

`parseInsightTotals` here (prefer `total_value`, fall back to summing `values[]`) handles both correctly. This is the right fix.

kjekkmec:
Thanks for the review — pushed `bea5c08` addressing all three points.

**1. Missing `.printing-press-patches/` entry** (@tmchow, Greptile P2) — Added `.printing-press-patches/instagram-media-insights-zeros.json` (schema v2), alongside the existing `instagram-livegate-fixes` / `instagram-novel-analytics` descriptors, recording this customization to the generated collector.

**2. `total_value` presence detection via empty-string sentinel** (Greptile P2) — `TotalValue` is now a pointer, so presence is keyed on whether the `total_value` object existed in the JSON rather than on its inner `value` being non-empty. `{"total_value":{}}` now records `0` (precedence preserved) instead of silently falling through to `values[]`; `total_value:null`/absent still falls through to the media `values[]` shape. Added `TestParseInsightTotals_EmptyTotalValueWins` and `TestParseInsightTotals_NullTotalValueFallsThrough` to lock both in.

**3. Vacuous zero assertion + stale doc block** (Greptile summary) — `TestParseInsightTotals_MediaValuesShape` now asserts key *presence* (not just `== 0`), so the `shares: 0` case can no longer pass when a metric is skipped. Also removed the duplicated doc comment on `fetchInsightTotals`.

`go build ./...`, `go vet ./internal/cli/`, and the full `internal/cli` suite (incl. the 5 `ParseInsightTotals` cases) pass.

tmchow:

- 1 unresolved non-outdated review thread(s) remain

tmchow:

- 1 unresolved non-outdated review thread(s) remain

tmchow:
Closing this out. Quiet since the July follow-up on the unresolved review thread. Happy to look at a fresh PR when that is addressed.

PR #1264, opened 2026-06-17
fix(company-goat): disable Algolia typo tolerance in HN search to prevent false-positive mentions

abresler:
## Problem

`company-goat mentions <company>` (and `launches`) resolve a company to its domain and search Hacker News via the Algolia API using the domain's first label as the query (e.g. `pwcommunications` from `pwcommunications.com`).

Algolia's default `typoTolerance=true` allows up to two character edits on words of 8+ characters. So a long company token like `pwcommunications` (16 chars) is a two-deletion fuzzy match for the ordinary word **communications**, and the search returns unrelated stories whose only connection is that they contain that word.

### Reproduction (before)

```
company-goat mentions --domain pwcommunications.com
# -> 1 "mention": HN story 195959, "Ask Eliezer Yudkowsky: How did you
#    convince the Gatekeeper to release the potentially genocidal AI?" (2008).
#    The only link is the word "communications" appearing once in that thread.
```

Confirmed against the live public API:

```
GET /search_by_date?query=pwcommunications&tags=story                      -> nbHits=1   (false positive)
GET /search_by_date?query=pwcommunications&tags=story&typoTolerance=false  -> nbHits=0   (correct)
```

## Fix

Disable typo tolerance for all HN searches. The change lives in one shared `buildURL` helper, so `SearchAll`, `SearchShowHN`, and `SearchByDate` all inherit it. The public `hn.algolia.com` proxy honors the `typoTolerance` parameter.

Genuine mentions are unaffected — exact-token matches still resolve:

```
GET /search?query=stripe&tags=story&typoTolerance=false  -> 5918 hits
```

After the fix:

```
company-goat mentions --domain pwcommunications.com  -> 0 mentions    (correct)
company-goat mentions --domain stripe.com            -> 5918 mentions (unchanged)
```

## Tests

Adds `internal/source/hn/hn_test.go`:

- a fast unit check on `buildURL`, and
- an `httptest`-backed test that drives `SearchAll` / `SearchShowHN` / `SearchByDate` and asserts every real outgoing request carries `typoTolerance=false`. This guards the call sites, not just the helper, so a future refactor that bypasses `buildURL` can't silently reintroduce the bug.

`go test -race ./...` passes.

## Notes

This trades a small amount of fuzzy recall for exact-token precision, which matches the intent of a company-name lookup. If a caller later wants configurable fuzzy matching, `buildURL` is the single place to thread an option through.


abresler:
Thanks for the review. Addressed in the latest commit:

- `buildURL` now copies the caller's `url.Values` instead of mutating it
- normalized the base/endpoint join so a missing trailing slash can't produce a malformed URL (`.../v1search?...`)
- the test captures the observed query over a channel, so the server-handler and test goroutines no longer share a variable unsynchronized

`go test -race ./...` passes; behavior unchanged (pwcommunications -> 0, stripe -> 5918).

tmchow:

- existing published CLI Go changes need a .printing-press-patches/ metadata entry

tmchow:

- existing published CLI Go changes need a .printing-press-patches/ metadata entry

tmchow:
Going to close this one out — it's been quiet since the July ask for the `.printing-press-patches/` metadata entry on the company-goat HN change.

Happy to take a fresh PR whenever that patch record is on the branch.

PR #1267, opened 2026-06-17
feat(wayback-goat): add wayback-goat-pp-cli

abresler:
## What

Adds **wayback-goat**, a CLI over the Internet Archive Wayback Machine **CDX Server API** (no authentication). Generated with `cli-printing-press`, plus one hand-authored analytic command on top of the generated `cdx` surface.

## Why it's more than a CDX wrapper

The headline command is `changes`. It collapses a URL's capture history by **content digest** (the SHA1 of the archived payload that CDX returns) and reports exactly *when* the page content changed: the first capture is the baseline (first-seen), and every later digest flip is a change event. By default only HTTP 200 captures are considered, so a transient redirect or 404 snapshot is not mistaken for a content change.

This is analysis the Wayback web UI does not surface, and that URL-listing tools (e.g. `waybackurls`) do not do — they enumerate captured URLs; `changes` tells you the content-change cadence of a single page over its archived lifetime.

Real example:

```
$ wayback-goat-pp-cli changes pwcommunications.com
pwcommunications.com — 154 captures, first seen 2001-04-01, 102 content change(s)
  2003-06-28  K4AZ6YVP… → 4LDD6XUS…  (HTTP 200)
  2004-06-05  4LDD6XUS… → LJD3EL5G…  (HTTP 200)
  ...
```

## Verification

- `cli-printing-press publish validate` — all checks pass (manifest, transcendence, phase5, `go mod tidy`, govulncheck, `go vet`, `go build`, `--help`, `--version`, verify-skill).
- `go test ./...` — passes. The change-detection oracle is pinned by a CI-safe regression test (`httptest`-backed; no live network), including the case where consecutive identical digests must NOT be counted as changes.
- `cli-printing-press dogfood --live` — 16 passed / 0 failed.
- `cli-printing-press scorecard` — 81/100 (Grade A).

## Notes

- **Auth:** none — the public CDX endpoint. The client sends an identifying `User-Agent` as a courtesy to the Archive (a nonprofit).
- **Manuscripts:** research briefs are not included — this was produced with `generate` plus a hand-authored analytic command rather than the full `print` research pipeline. The phase 5 dogfood acceptance proof is included.


tmchow:
@abresler i love the idea of this but CLIs need to be published using the printing-press-publish skill and the entire skill process. Not sure what happened here but my guess is you're trying to automate this (at least partly) with an agent that's failing to follow instructions.

Also, we save "Goat" labels in CLIs when they cover more than one service. Not just because a CLI is good.

PR #1311, opened 2026-06-21
feat(registry): surface printing_press_version

e-jung:
## Intent

Surface each CLI's Printing Press generator version in the public registry so downstream consumers can reason about machine-version freshness. This is the additive registry slice of #1299.

## Approach

- Add optional `printing_press_version` to `tools/generate-registry`'s `RegistryEntry` with `omitempty`.
- Populate it from `.printing-press.json:printing_press_version`, which this repo documents as the generator version that produced the CLI.
- Expose the same optional field through the npm registry parser without making old registry entries invalid.
- Add Go coverage for JSON round-trip/empty behavior and manifest propagation.
- Add npm parser coverage for present, absent, and blank values.
- Bump the npm package patch version because `npm/src/**` changed.

## Scope

This PR does not regenerate or commit `registry.json` or `cli-skills/**`; those remain post-merge generated artifacts. It also does not add dashboard, stale badges, or update behavior from the broader freshness proposal.

## Risk

Low. The field is optional and omitted when the manifest value is empty, so existing registry entries and consumers that ignore unknown fields remain compatible. npm parsing treats missing or blank values as absent.

## Verification

- `go test .` in `tools/generate-registry`
- `npm test` in `npm`
- `go run ./tools/generate-registry/main.go --validate`
- `git diff --check`
- `no-mistakes axi run --yes --skip=push,pr,ci --intent ...` passed; push/PR/CI were skipped there so this fork-only PR flow could use the required `e-jung` branch and upstream PR command.

## AI disclosure

Human-reviewed.

Issue #1299, opened 2026-06-20
proposal: surface machine-version freshness to end users + contributors (stamp → registry → dashboard → notify)


e-jung:
## Problem

`cli-printing-press` (the machine) is at v4.24.0 and ships correctness fixes
nearly every release. `printing-press-library` (this repo) is federated — 81%
of our 287 CLIs are community-contributed across ~30 handles, and the machine
moves faster than anyone reprints. Measured today:

| Bucket | Count | Share |
|---|---|---|
| At current machine (4.24.0) | 34 | 12% |
| Above currency floor (4.19.1) but stale | 24 | 8% |
| **Below currency floor** (would fail a new-print gate) | **228** | **79%** |

Including 100% of @mvanhorn's own 26 CLIs and 93% of @tmchow's — this is not a
community-laziness problem, it's structural. There is no machinery today that
tells an end user "your installed CLI is stale," tells a contributor "your
published CLI lags the machine," or tells a maintainer "X% of the library is
below the floor." The reprint mechanism (`/printing-press-reprint`) exists but
is purely contributor-initiated and currently **blocked** by #3080.

## Evidence (this repo + machine)

- The per-CLI machine-version stamp already exists in
  `.printing-press.json:printing_press_version` and in
  `.printing-press-release.json`, but `tools/generate-registry/main.go`
  deliberately omits it from `RegistryEntry` (`main.go:67-96`, `:129-157`), so
  it never reaches `registry.json`, the npm installer, `cli-skills/`, or any
  installed-CLI runtime surface.
- `npm/src/commands/update.ts` is a dumb sweep (line 75: "No Printing Press
  CLIs found on PATH to refresh") — no staleness awareness.
- `npm/src/commands/list.ts` renders no version column.
- `/printing-press-amend` SKILL L416 admits: "F13 doctor doesn't surface
  stale-binary warning vs. published version."
- `bundle-audit.yml` already reads `printing_press_version` per CLI and posts a
  sticky-issue table — that pattern is the natural template for a freshness
  dashboard.
- `gh search issues` across both repos for `end user`, `installed CLI`,
  `notify contributor`, `update available`, `library refresh`, `library sweep`
  returns **zero** prior proposals.

Adjacent machine-side issues (not duplicates):
- #3080 — v4.24 `version.go` move breaks pre-v4.24 reprints (blocking).
- #2487 — printed CLIs have no evolving version; amend's stale-check is a
  permanent no-op.

[...]

PR #1312, opened 2026-06-21
feat(update): --stale-only skips current CLIs (uses printing_press_version)

e-jung:
## Intent

Make `printing-press-library update` smart: skip CLIs whose installed generator version matches the registry, re-installing only the stale or unknown ones. Today `update` re-installs every detected CLI unconditionally — no staleness check. This is the strategic move from the pp-update-explore-k2 scout and the visibility/staleness gap [#1299](https://github.com/mvanhorn/printing-press-library/issues/1299) targets.

Builds on [#1311](https://github.com/mvanhorn/printing-press-library/pull/1311) (`printing_press_version` field in `RegistryEntry`); branched off `feat/registryentry-printing-press-version`.

## Approach

Three additive pieces, all behind injectable deps for testing:

**1. Installed-version stamp** (`npm/src/versions.ts`, new): read/write a parallel stamp at `~/.agents/.pp-cli-versions.json` keyed by CLI catalog name → `printing_press_version`. A parallel stamp (not the `skills` package's `.skill-lock.json`) keeps the version tracking decoupled from that file's content-hash contract — the two tools evolve independently. Missing/corrupt stamps read as `{}` so `update` degrades to the old sweep on first run and after any filesystem mishap.

**2. Stamping on install** (`install.ts`): after a successful install, record `entry.printing_press_version` in the stamp. Only stamped when the registry carries the field; entries without it stay unstamped so `update` treats them as unknown (→ re-install) rather than silently current. `writeInstalledVersion` is best-effort — a failed write is swallowed so it can't fail an otherwise-successful install.

**3. Smart update** (`update.ts`): after detecting installed CLIs, partition into "current" (skip) and "stale/unknown" (re-install). A CLI is current only when both the stamp and the registry carry a `printing_press_version` and they match; any uncertainty (missing stamp, missing field, mismatch) → re-install. Skip notices are emitted before the buffered install output; a summary line reports the skip count and points at `--all`.

**Flags:**
- `--stale-only` is the **default** (smart skip).
- `--all` forces the old unconditional sweep.
- Both together: last flag wins (conventional CLI toggle behavior).
- Single-target `update <name>` ignores stale-only — naming a CLI explicitly is a force-refresh of that one CLI.

**Graceful degradation:** CLIs whose registry entry lacks `printing_press_version` (older prints, pre-#1311 catalog entries) are treated as unknown and always re-installed. Smart-update behaves identically to the old sweep until the field propagates across the catalog.

## Scope

| File | Change |
|---|---|
| `npm/src/versions.ts` (new) | Stamp read/write module with injectable IO |
| `npm/src/commands/install.ts` | Stamp after successful install; add stamp deps |
| `npm/src/commands/update.ts` | `--stale-only` (default) / `--all` flags; skip logic |
| `npm/src/cli.ts` | Help text for new flags |
| `npm/tests/versions.test.ts` (new) | 12 unit tests for stamp read/write |
| `npm/tests/install.test.ts` | 2 tests: stamps when version present, skips when absent |
| `npm/tests/commands.test.ts` | 7 tests: stale-only skip/current, --all, unknown, default, help |
| `npm/README.md` | Document smart update + flags |
| `npm/CHANGELOG.md` | 0.1.18 + 0.1.19 entries |
| `npm/package.json` / `package-lock.json` | 0.1.18 → 0.1.19 (verify-npm-version-bump gate) |
| `AGENTS.md` | Document smart-update behavior in NPM installer surface section |

## Risk

- **Behavior change:** `update` (no name) now skips current CLIs by default. Users who relied on the unconditional sweep get it back with `--all`. The skip is safe: it only fires when stamp and registry versions are both present and match; any uncertainty → re-install.
- **First run after this ships:** stamps are empty, so every detected CLI is "unknown → re-install" — identical to the old sweep. The second run starts benefiting from skips.
- **No change to single-target** `update <name>` — still always installs.
- **Stamp file location** (`~/.agents/.pp-cli-versions.json`) sits alongside the skills lockfile but is a separate file; the skills package never reads or writes it.

## Verification

```
npm test  → 113 tests, 113 pass, 0 fail
npm run build → clean (tsc strict)
```

Key test scenarios covered:
- Install at version X → bump registry to Y → `update --stale-only` → only stale CLI re-installs, current one skipped
- Missing stamp entry → unknown → re-install
- Missing registry `printing_press_version` → unknown → re-install
- `--all` forces re-install of current CLIs
- Default (no flag) is stale-only
- Help documents both flags

## AI disclosure

Human-reviewed.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -335,6 +335,8 @@ Adding or updating a CLI does not require an npm publish. The package is a thin 
 
 The repo is public, the package is published, and `npx -y @mvanhorn/printing-press install <name>` works end-to-end for unauthenticated users.
 
+`update` (no name) is **smart by default**: it stamps each installed CLI's `printing_press_version` into `~/.agents/.pp-cli-versions.json` at install time, then on the next run compares that stamp against the registry to skip current CLIs and re-install only stale/unknown ones. `--all` forces the old unconditional sweep; `--stale-only` is the default. CLIs whose registry entry lacks `printing_press_version` are treated as unknown and always re-installed, so smart-update degrades gracefully to the old sweep until the field propagates across the catalog.
+
 ## Releasing the npm installer
 
 Single rule: **bump `npm/package.json`'s `version` field in your release PR.** Everything else is automated.
--- a/npm/CHANGELOG.md
+++ b/npm/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## 0.1.19
+
+- Make `update` smart by default: skip CLIs whose installed generator version (`printing_press_version`) matches the registry, re-installing only stale or unknown ones. Adds `--stale-only` (default) and `--all` flags; `--all` restores the old unconditional sweep.
+- Record the installed `printing_press_version` per CLI in `~/.agents/.pp-cli-versions.json` on every successful `install`, so `update --stale-only` can compare against the live registry.
+
+## 0.1.18
+
+- Surface an additive `printing_press_version` field on `RegistryEntry` (TS) and the Go registry generator, sourced verbatim from each CLI's `.printing-press.json`.
+
 ## 0.1.17
 
 - Raise the documented direct `go install` floor to Go 1.26.4 so npm installer guidance matches the catalog-wide module enforcement floor.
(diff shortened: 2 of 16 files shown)


e-jung:
Closing in favor of #1311 — combined the version field + the smart upgrader into one PR since the field's purpose is the upgrader. All these commits are now in #1311.

PR #1346, opened 2026-06-24
fix(nse-india): fix movers, symbol-lookup, and live dogfood gate

lavs9:
## Summary

- **movers**: Switched from dead `index=INTRADAY` to `index=volume|value` via new `--by` flag; manually unwraps `{data:[...],timestamp}` envelope that `extractResponseData` skipped (no `status` field).
- **symbol-lookup**: Replaced permanently removed `/api/search/autocomplete` with NSE archives `EQUITY_L.csv` fetch + client-side substring search. Returns symbol, company name, ISIN, series for all 2372 listed equities.
- **equity quote/derivatives**: Hardcoded internal NSE `functionName` API routing parameter; removed `--function-name` flag users should never change (was causing live dogfood to inject `example-resource` → HTTP 400).
- **novel commands** (announcement-flood, delivery-divergence, delivery-spike, iep-drift, sector-breadth, index-driver): Return `[]` instead of human text when `--json` is active and local store is empty — fixes json_fidelity dogfood failures.
- **portfolio pnl/margin-health**: Treat missing holdings file as empty portfolio (exit 0 + empty JSON) rather than an error.

Live dogfood gate: **97/97 pass** (was 15 failures before these fixes).

## Patches added

| Patch | Description |
|-------|-------------|
| `movers-index-param-fix` | Fix movers endpoint param and envelope unwrap |
| `symbol-lookup-csv-search` | Replace dead autocomplete with CSV search |
| `equity-hardcode-function-name` | Hardcode internal NSE API routing param |
| `novel-empty-json-output` | Return `[]` for empty store in JSON mode |
| `local-module-import-paths` | Fix module import paths for local builds |
| `indices-new-endpoints` | Migrate to new NSE indexTrackerApi endpoints |

## Test plan

- [x] `nse-india-pp-cli movers` returns ~20 active securities
- [x] `nse-india-pp-cli movers --by value` returns results ordered by traded value
- [x] `nse-india-pp-cli symbol-lookup --query RELIANCE` returns RELIANCE + related matches
- [x] `nse-india-pp-cli equity quote --symbol INFY` returns live quote data
- [x] `nse-india-pp-cli delivery-spike --json` returns `[]` (not error) when store is empty


tmchow:
@lavs9 fix Greptile feedback:

> The index-driver SQL query uses $.symbol and $.pChange (camelCase), while the new indexTrackerApi endpoint stores raw items with cmSymbol and pchange. The write-through cache also requires an id field to persist items, which constituent records don't have. This means running indices constituents then index-driver will always produce an empty result set silently. All other fixes (movers, symbol-lookup, equity, portfolio, novel commands) look correct.

lavs9:
> @lavs9 fix Greptile feedback:
> 
> > The index-driver SQL query uses 
> >   .
> >   s
> >   y
> >   m
> >   b
> >   o
> >   l
> >   a
> >   n
> >   d
> > .pChange (camelCase), while the new indexTrackerApi endpoint stores raw items with cmSymbol and pchange. The write-through cache also requires an id field to persist items, which constituent records don't have. This means running indices constituents then index-driver will always produce an empty result set silently. All other fixes (movers, symbol-lookup, equity, portfolio, novel commands) look correct.

resolved.

tmchow:

- missing publish-process files: library/developer-tools/nse-india/go.mod

lavs9:
@tmchow the triage flag **`missing publish-process files: library/developer-tools/nse-india/go.mod`** looks like a false positive on this PR — `go.mod` is present and healthy on the branch head (`91c8e0b`):

- Present at exactly `library/developer-tools/nse-india/go.mod`
- Valid module path (`github.com/mvanhorn/printing-press-library/library/developer-tools/nse-india`), `go 1.26.5`
- **Byte-identical to current `main`** (both `go.mod` and `go.sum`)
- `go mod tidy` is a no-op and `go build ./...` passes

It's absent from the PR **diff** only because it's unchanged: this PR fixes CLI source (movers, symbol-lookup, live-dogfood gate) without adding or removing any dependency, so `go.mod` legitimately didn't change. All Greptile inline threads are resolved.

Could you confirm what the check actually requires? If it's a diff-completeness heuristic ("CLI source changed but its `go.mod` isn't in the changeset"), there's no honest content change to make to `go.mod` here. Happy to do whatever unblocks it — a maintainer override, or guidance if you'd prefer a different PR shape.

tmchow:
@lavs9 you're right. That earlier note about a missing `go.mod` was a false positive. The file is on the branch and matches main, so it correctly does not show up in the PR diff. Sorry for the noise.

Taking another look from current facts and merging this.

PR #1351, opened 2026-06-24
fix(plane): derive child scope column from project on write-through cache

sidorovanthon:
## Problem

Project-scoped child tables (`modules`, `cycles`, `states`, `labels`, `projects_issues`, `intake_issues`, `archived_*`) declare a NOT NULL path-placeholder scope column `projects_id`. Only the dependent-resource sync injects it per item (`obj[parentTable+"_id"] = parentID`).

The two write-through cache paths in `internal/cli/data_source.go` — `writeThroughCache` (auto/live read results) and `writeMutationResponseToStore` (create/update responses) — pass the **raw** API body straight to `UpsertBatch`. The Plane body carries the parent UUID as the singular `project`, never `projects_id`. So the per-item typed dispatch hit:

```
insert into modules: constraint failed: NOT NULL constraint failed: modules.projects_id (1299)
```

the per-item `SAVEPOINT` rolled back, and the row stranded in the generic `resources` table with no typed projection — surfacing as the existing warning:

```
warning: N/M modules items: typed-table upsert failed; generic resources rows preserved
```

Non-data-loss (the generic blob survives), but the typed table and typed queries silently miss those rows, and incremental sync never re-fetches them, so the strand is permanent.

## Fix

`UpsertBatch` now backfills the typed scope column from the item's own parent reference (`project`) when the path injection is absent, via a small `deriveScopeColumns(obj)` helper + a `childScopeColumnSources{"projects_id":"project"}` map. An already-present (injected) value is never overwritten, so the dependent-sync path is a no-op. `projects_id` always equals the project UUID for these tables (the dependent sync sets `projects_id` = the iterated project's id, the same UUID the body's `project` field holds).

Every entry point — sync, live-read cache, mutation cache — now populates the typed table.

## Tests

- Adds `TestUpsertBatch_DerivesChildScopeFromProjectField`: a raw write-through item `{id, project, name}` with **no** `projects_id` now lands in the typed `modules` table with `projects_id` derived from `project`.
- Existing `TestUpsertBatch_TypedFailureDoesNotStrand*Generic` still pass: their fixtures carry neither `projects_id` nor `project`, so no-parent-info stranding (generic preserved, typed=0) remains the correct behavior.

`go build ./...` + `go vet` + `go test ./internal/store ./internal/cli` all green on the `main` base.

## Notes

Diff is plane-only (3 files, +124). Reproduced the exact warning + `typedErr` with a failing test before fixing.

### DIFF
--- a/library/project-management/plane/.printing-press-patches/write-through-child-scope.json
+++ b/library/project-management/plane/.printing-press-patches/write-through-child-scope.json
@@ -1,0 +1,26 @@
+{
+  "schema_version": 2,
+  "id": "write-through-child-scope",
+  "applied_at": "2026-06-24",
+  "base_run_id": "20260602-073610",
+  "base_printing_press_version": "4.20.1",
+  "summary": "Backfill a typed child table's NOT NULL path-placeholder scope column (projects_id) from the item's own parent reference (project) inside UpsertBatch, so the write-through cache paths (live-read + mutation-response caching) stop stranding project-scoped child rows in the generic resources table with no typed projection.",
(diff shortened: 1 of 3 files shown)


sidorovanthon:
Superseded by #1361 (Round-4 tenant-resolver + flat reconcile), which is merged and includes this same child-scope fix (deriveScopeColumns / childScopeColumnSources + the DerivesChildScopeFromProjectField test) on top of upstream/main. Closing as redundant — the fix is now in main.

PR #1362, opened 2026-06-25
fix(postscript): dogfood framework command paths

debgotwired:
## Summary
- fix Postscript framework command paths for tail/import so unsupported resources are rejected instead of hopeful
- make local read paths and analytics use read-only SQLite access, with streaming group-by
- bound MCP SQL results and add focused dogfood regression tests plus patch metadata

## Validation
- go test ./...
- go test -race ./...
- go vet ./...
- go build for CLI and MCP
- govulncheck ./...
- git diff --check
- verify-skill
- verify-publish-package
- verify-supply-chain
- cli-printing-press publish validate
- cli-printing-press shipcheck --no-live-check --no-fix

Live Postscript API acceptance was not run because I do not have a real Postscript account/API key.


debgotwired:
Reworked this after the earlier closures/comments across the recent CLI PRs about insufficient testing and non-dogfooded behavior. I treated those as blockers for this Postscript retry, not just normal CI feedback.

I did not rely on green checks alone. I re-read the repo instructions, reviewed the prior Postscript review thread, cross-checked the Outreach feedback pattern, and then dogfooded the Postscript CLI surfaces directly.

What changed:
- tail subscribers now uses the real Postscript path: GET /api/v2/subscribers.
- tail events is rejected instead of pretending mutation-only resources can be tailed.
- analytics and local fallback now open SQLite read-only and no longer create/migrate empty local DBs for read commands.
- analytics --group-by streams synced rows instead of materializing the whole local store first.
- MCP SQL results are bounded with truncation metadata instead of returning unbounded row arrays.
- import subscribers is rejected; import events is the only import resource currently allowed and maps to POST /api/v2/events.
- Added focused tests for these behaviors plus patch metadata under .printing-press-patches/.

QA run locally:
- go test ./...
- go test -race ./...
- go vet ./...
- CLI and MCP go build
- govulncheck ./...
- git diff --check
- verify-skill
- verify-publish-package
- verify-supply-chain
- cli-printing-press publish validate
- cli-printing-press shipcheck --no-live-check --no-fix

Dogfooded with isolated HOME / XDG_DATA_HOME:
- doctor --json
- agent-context --pretty
- which
- api --json
- dry-run endpoint commands for subscribers list/get/update, events create, compliance unsubscribe/redact
- tail subscribers positive case and tail events negative case
- sync --dry-run --agent
- workflow archive/status
- import events dry-run and import subscribers negative case
- profile save/list/show/use/delete
- feedback and feedback list
- global --deliver file:...

What I could not verify: live Postscript API acceptance, because I do not have access to a real Postscript account/API key. If live vendor verification is mandatory for accepting this CLI, then this PR should be closed rather than merged. The non-live command behavior, generated paths, local-store behavior, MCP bounds, and Printing Press validators have all been exercised locally.

debgotwired:
@greptile-apps addressed the three P2 review comments and the broader summary cleanup on the existing PR branch.

Changes pushed in 558270078:
- `import --batch-size`: hidden until batch import is actually implemented, so the visible flag surface no longer advertises a no-op option.
- `import --input`: removed the unreachable `inputFile == ""` stdin fallback; stdin is now only `--input -`, matching the required flag contract.
- `resolveLocal` warning: routed the unfiltered local-data warning through `hintWriter` instead of direct `os.Stderr`, so callers/tests can capture or suppress it consistently.
- Patch metadata: corrected `dogfood-readonly-bounds-and-import-paths.json`, which had accidentally duplicated the tail patch metadata, and expanded it to cover this review cleanup.

Validation after the fix:
- `go test ./internal/cli ./internal/mcp ./internal/store`
- `go test ./...`
- `go test -race ./...`
- `go vet ./...`
- CLI and MCP `go build`
- `govulncheck ./...`
- `git diff --check`
- `verify-skill`
- `verify-publish-package`
- `verify-supply-chain`
- `cli-printing-press publish validate --dir library/marketing/postscript --json`
- `cli-printing-press shipcheck --dir library/marketing/postscript --spec library/marketing/postscript/spec.yaml --no-live-check --no-fix --json`

Also checked `postscript-pp-cli import --help`; `--batch-size` is no longer visible, while `--input` and `--dry-run` remain visible.

tmchow:
Closing it hasn't been fully tested as @debgotwired admitted.

PR #1383, opened 2026-06-27
fix(techmeme): stop --agent from blanking records to empty objects

mvanhorn:
## Problem

`techmeme-pp-cli search "<topic>" --agent` returns a list of empty `{}` objects instead of headlines.

Root cause: `--agent` implies `--compact`, and `compactListFields` (in `internal/cli/helpers.go`) keeps only a generic allow-list (`id/name/title/url/status/...`). Techmeme's primary records key their content off domain fields (`headline`, `link`, `source`, `num`, `topic`, `rank`, ...), none of which were in the allow-list, so compaction stripped every field and emitted an empty object. This makes the agent/JSON surface unusable both directly and for downstream source integrations (this surfaced while evaluating Techmeme as a `last30days` source).

## Fix

In `compactListFields`:

1. Add Techmeme's content fields to the keep allow-list (`headline`, `link`, `source`, `sources`, `num`, `topic`, `rank`, `date`, `time`, `count`, `velocity`, `permalink`).
2. Defense in depth: never let compaction blank a record. If no allow-listed field matched but the record carried data, keep the original record rather than emitting `{}`.

Verbose-field stripping is unchanged for records that *do* match the allow-list, so token savings are preserved.

## Validation

| Check | Result |
|---|---|
| `go build ./...` | PASS |
| `go vet ./...` | PASS |
| `go test ./...` | PASS (92 tests) |
| `govulncheck ./...` | PASS (no vulnerabilities) |
| `verify_skill.py` | PASS (all checks) |

Live before/after on the built binary:

- Before: `search "AI" --agent` -> `[{}, {}, {} ...]` (16 empty objects)
- After: `search "AI" --agent` -> `[{"headline":"...","link":"...","num":1,"source":"techcrunch.com"}, ...]` (16 populated records)

New regression tests in `internal/cli/helpers_compact_test.go` cover the headline shape, the no-blank guard, and that verbose fields are still stripped when an identifier is present.

## Notes

- Customization recorded under `.printing-press-patches/compact-must-not-blank-records.json`.
- Contributor entry added across manifest, README byline, and NOTICE (creator untouched).
- The empty-object-on-compact behavior is generator-shaped (`compactListFields` is generated boilerplate). Per the split-by-lifecycle convention this is patched in the published library first; the allow-list gap and no-blank guard should also land in the `cli-printing-press` templates so other headline/link-shaped CLIs don't regress.


mvanhorn:
Addressed both Greptile findings in b9f14ac:

1. No-blank guard now copies `item`'s entries into `compact` rather than aliasing the original map, so `compact` stays independently owned and consistent with the rest of the loop.
2. Added `TestCompactListFields_GuardPreservesVerboseWhenNoIdentifier` covering a record with a verbose field but no allow-listed field -- the guard fires and the record is preserved intact, making the guard-vs-verbose behavior explicit.

Full module suite still green (now 4 compact tests).

Issue #698, opened 2026-05-19
listing and comps return 403 — initialInfo endpoint blocked


miron-tech:
Running `comps` or `listing` against any Redfin URL returns a 403 from `/stingray/api/home/details/initialInfo`. `doctor` shows API reachable and auth not required, so the Chrome TLS transport doesn't appear to be bypassing Redfin's block on this endpoint.

```
redfin-pp-cli comps "https://www.redfin.com/TX/Dallas/11102-Ferndale-Rd-75238/home/31064649" --radius 0.5 --months 6 --json
→ Error: could not resolve subject listing
```

```
redfin-pp-cli listing "https://www.redfin.com/TX/Dallas/11102-Ferndale-Rd-75238/home/31064649" --json
→ GET /stingray/api/home/details/initialInfo returned HTTP 403
```

Mac, v1.0.0.

rob-coco:
taking this issue.

rob-coco:
Investigation finding — this is **not** a CLI request-shape issue, and I don't have a clean fix to ship.

### What I tested

Built the CLI from current \`main\` and reproduced the 403. Then:

1. **Verified surf is actually applying Chrome impersonation through \`.Std()\`.** Read \`enetx/surf@v1.0.199/adapter.go\` and \`impersonate.go\`. \`Impersonate().Chrome()\` installs Chrome 145's JA3, HTTP/2 settings, and the full browser header set (User-Agent, sec-ch-ua, sec-ch-ua-mobile, sec-ch-ua-platform, Accept-Language, Accept-Encoding) via middleware that the \`TransportAdapter\` runs on every request. The CLI's existing path through \`.Std()\` + \`http.NewRequest\` + \`httpClient.Do(req)\` does carry those.

2. **Added an explicit Chrome browser-headers RoundTripper wrapper** (analogous to the myfitnesspal #403 fix). Still 403 on \`/stingray/api/home/details/initialInfo\`. Reverted as redundant.

3. **Probed with curl from this machine** carrying a real Chrome User-Agent + \`Accept: application/json\` + \`X-Requested-With: XMLHttpRequest\` + \`Referer: <listing-url>\` + a freshly-warmed cookie jar from \`https://www.redfin.com/\` + the listing page itself. Still 403.

[...]

PR #1389, opened 2026-06-27
fix(redfin): make HTTP GET cache key deterministic

Avanderheyde:
## fix(redfin): make HTTP GET cache key deterministic

Narrow correctness fix for an existing published CLI (`library/other/redfin`).

### The bug

The generated HTTP client caches GET responses for 5 minutes, keyed by `cacheKey(path, params)`. That key was built by ranging **directly over the `params` map**:

```go
key := path
for k, v := range params {   // Go randomizes map iteration order
    key += k + "=" + v
}
```

Go randomizes map iteration order on every range, so two *identical* requests produced different key strings → different `sha256` → different cache file. `readCache` therefore almost never found the entry `writeCache` had just written: the 5-minute GET cache silently never hit, and every `homes` / `comps` / `listing` / `market` / `trends` invocation fired a fresh live request against Redfin's anti-bot Stingray endpoint. For a scraper that already trips 403 / rate-limit blocks (see #698), an effective cache is exactly the protection that was being lost.

### The fix

Sort the param keys before hashing (and separate pairs with `&` so distinct `(k,v)` pairs can't concatenate ambiguously). One small, surgical change to `internal/client/cacheKey`; no dependency or API changes.

### Validation

All run from `library/other/redfin/`:

| Check | Result |
|---|---|
| `go build ./...` | ✅ PASS |
| `go vet ./...` | ✅ PASS |
| `go test ./...` | ✅ PASS (incl. new `internal/client` tests) |
| `gofmt -l` (touched files) | ✅ clean |
| `govulncheck ./...` (reachability) | ✅ 0 affected |
| `verify_skill.py --dir .` | ✅ all checks passed |

New tests in `internal/client/client_test.go`:
- `TestCacheKeyDeterministic` — 1000 calls with the same params return one key (failed before the fix, passes after).
- `TestCacheKeyOrderIndependent` — maps with identical content hash equal.
- `TestCacheKeyDistinctParams` — distinct `region_id` / path still produce distinct keys (no collisions).

### Repo conventions

- Hand-edit to a generated file recorded at `.printing-press-patches/cache-key-deterministic.json` (reprint-guard).
- Contributor accrued across all three surfaces — `.printing-press.json` `contributors[]`, README byline, and NOTICE.
- No changes to `registry.json` or `cli-skills/pp-*` (regenerated post-merge).

### Upstream note

`internal/client/client.go` is generated (`DO NOT EDIT`), so this `cacheKey` shape is template-wide — the same non-deterministic key exists in other published CLIs' clients. Scoped to `redfin` here per the narrow-fix / don't-sweep-the-whole-library convention; the generator (`cli-printing-press`) should emit a sorted key so future prints get it for free.


Avanderheyde:
Closing from our side: this work was opened against the library repo, but the current direction is to keep these changes out of printing-press-library unless they are the intended library-add shape.

PR #1414, opened 2026-06-30
docs(here-now): add House Rules — authenticated publish by default, edit in place to keep the URL

Edouardtriet:
Adds a short **House Rules** block to the `here-now` SKILL.md, capturing two defaults that came out of real agent usage:

1. **Assume logged in** — when credentials are present, just publish authenticated and return the slug/URL ("code") instead of defaulting to `--anon` + the claim dance.
2. **Edits stay on the same link** — to update an already-published site, re-publish in place with `publish update-site <slug>` rather than re-running `publish dir` or delete+recreate. Both of the latter mint a *new* slug, and since deleted slugs are permanently reserved, the original URL is lost.

The second one bit us in practice: an "update" via `publish dir --slug <existing>` silently produced a new site, and trying to reclaim the old slug by deleting + recreating failed (slug reserved), churning the public URL twice. The note steers agents to `update-site` so the shared link stays stable.

Docs-only change to the source `library/cloud/here-now/SKILL.md`; the generated skill mirrors regenerate post-merge. No code changes.


tmchow:

tmchow:

- Verify concluded FAILURE

tmchow:
Following up on the previous triage note. This PR is still waiting on author action for the same blocker(s) after 10 days:

- Verify concluded FAILURE

tmchow:
Closing this as stale. We asked on July 10 about the Verify failure, followed up July 20, and there has been no update since.

Open a fresh PR when Verify is green. If the leftover is just a missing update-in-place recipe in SKILL.md, that is enough to include next time.

PR #1460, opened 2026-07-08
feat(faa-registry): add faa-registry

omarshahine:
## faa-registry

Every FAA aircraft lookup the registry website offers, plus a daily-synced offline copy of the entire US registry that unlocks fleet reports, hex decoding, ownership history, and expiration alerts no other tool has.

**API:** faa-registry | **Category:** travel | **Press version:** 4.27.1
**Spec:** internal YAML (sniffed from registry.faa.gov/aircraftinquiry — no official OpenAPI exists)

### Publication Path

New print

### CLI Shape

```bash
$ faa-registry-pp-cli --help
Every FAA aircraft lookup the registry website offers, plus a daily-synced offline copy of the entire US registry that unlocks fleet reports, hex decoding, ownership history, and expiration alerts no other tool has.

  • fleet report   One command turns an owner name into a full fleet profile: aircraft count, model mix, jet/turboprop/piston split, average seats and year built.
  • hex resolve   Resolve any number of ADS-B Mode S hex codes (args or stdin) to N-numbers, aircraft types, and owners — offline.
  • aircraft history   Chronological owner timeline for a tail number, stitching current registration with every deregistration record.
  • expiring   List registrations expiring within a window, filtered by owner or state, sorted soonest-first.
  • models fleet   For any make/model, break down every registered example by registrant type (corporate, individual, LLC, co-owned) and state.
  • nnumber available   Check whether an N-number is assigned, reserved, or free — computed locally, with the reason.

Health check: run 'faa-registry-pp-cli doctor' to verify auth and connectivity.

Usage:
  faa-registry-pp-cli [command]

Available Commands:
  agent-context Emit structured JSON describing this CLI for agents
  api           Browse all API endpoints by interface name
  completion    Generate the autocompletion script for the specified shell
  dealers       Search FAA dealer certificates by dealer name.
  doctor        Check CLI health
  documents     Search the FAA document index by collateral identifier.
  engines       Search the engine reference table by engine manufacturer and model.
  expiring      List registrations expiring within a window, soonest first
  feedback      Record feedback about this CLI (local by default; upstream opt-in)
  fleet         fleet subcommands: report
  help          Help about any command
  hex           hex subcommands: resolve, to-tail, from-tail
  models        Search the aircraft model reference by manufacturer and model name
  nnumber       nnumber subcommands: available
  owners        List all aircraft registered to an owner name (paginated).
  profile       Named sets of flags saved for reuse
  search        Full-text search the offline registry by owner, co-owner, manufacturer, or model
  sync          Download the FAA daily Releasable Aircraft Database into the local registry
  version       Print version
  watch         Watch owners or tail numbers for registration changes across syncs
  which         Find the command that implements a capability
  workflow      Compound workflows that combine multiple API operations

Flags:
  -h, --help                 help for faa-registry-pp-cli
      --profile string       Apply values from a saved profile (see 'faa-registry-pp-cli profile list')
      --rate-limit float     Max requests per second (0 to disable, default 2 for sniffed APIs) (default 2)
  -v, --version              version for faa-registry-pp-cli

Use "faa-registry-pp-cli [command] --help" for more information about a command.
```

### Novel Commands

| `fleet report` | Fleet composition report | Turn an owner name into a full fleet profile: count, model mix, jet/turboprop/piston split, avg seats and year — a MASTER×ACFTREF join no single API call produces. |
| `hex resolve` | Batch Mode S hex resolution | Resolve any number of ADS-B Mode S hex codes (args or stdin) to N-numbers, aircraft types, and owners, offline; pure-algorithm fallback for unassigned codes. |
| `aircraft history` | Aircraft ownership history | Chronological owner timeline stitching the current registration with every deregistration record (the FAA site shows only the current owner). |
| `expiring` | Expiring registrations | List registrations expiring within a window, filtered by owner or state, soonest-first — the alert the FAA sends by paper mail. |
| `models fleet` | Model-class fleet breakdown | Break down every registered example of a make/model by registrant type and state, with an honest per-variant matched-models split. |
| `nnumber available` | Offline N-number availability | Check whether N-numbers are assigned, reserved, or free — with the reason — computed locally at batch scale (the FAA's own page rejects scripted requests). |

### What This CLI Does

The FAA Civil Aviation Registry is the authoritative source for every US-registered aircraft, but it's browser-only, form-driven, and un-scriptable (Akamai returns 403 to non-browser clients). This CLI wraps both of its surfaces: the live `aircraftinquiry` web app — nine inquiry types parsed from the site's semantic `data-label` markup into typed JSON — and the daily Releasable Aircraft Database (~315K active registrations, ~383K deregistered, ~126K reserved N-numbers, plus model and engine reference data) synced into a local SQLite store with FTS5.

That combination is what no existing tool offers: the flagship `aircraft lookup <N>` returns a full live registration record, while `sync` unlocks offline fleet reports, Mode S hex↔tail decoding, ownership history, expiration alerts, and instant availability checks. It's read-only, needs no authentication (public data), and works entirely offline after one sync.

### Manuscripts

- [Research Brief](library/travel/faa-registry/.manuscripts/20260708-063305/research/)
- [Shipcheck & Acceptance Proofs](library/travel/faa-registry/.manuscripts/20260708-063305/proofs/)

### Validation Results

| Check | Result |
|-------|--------|
| Manifest | PASS |
| Phase 5 (full live dogfood, 86/86) | PASS |
| go mod tidy | PASS |
| go vet | PASS |
| go build | PASS |
| --help | PASS |
| --version | PASS |
| Manuscripts | PRESENT |


omarshahine:
Thanks for the review. All three findings addressed in `2ae6a12`:

**1. Generated mirror breaks publishing (P1)** — resolved. The `cli-skills/pp-faa-registry/SKILL.md` mirror was dropped from the diff in the prior push (`577fc9b`); the current PR head is library-only, and the `Verify` / generated-artifacts gate now passes. The post-merge `generate-skills.yml` will create the mirror from `library/travel/faa-registry/SKILL.md`. (This comment is "outside diff" because the file is no longer in the PR.)

**2. Partial archives look fresh (P1)** — fixed. `importZip` now validates that every required file (MASTER, ACFTREF, ENGINE, DEREG, RESERVED) is present in the archive *before* mutating any table, and errors out with the missing filenames if not — leaving the previous good import untouched. A partial archive can no longer import a subset and leave empty manufacturer/model/engine joins.

**3. Search index commits separately (P1)** — fixed. Added a `sync_complete` marker that is cleared at the start of an import and set only after every table **and** the FTS rebuild land in one run. `Synced()` now gates on that marker instead of a row count, so a torn or failed sync (missing file, failed table import, or failed FTS rebuild) reports **not synced** — offline commands tell the user to re-sync rather than reading a new registry against a stale/empty search index.

New test `TestIncompleteArchiveRejected` locks in that a missing-file archive both errors and leaves `Synced()` false. Verified live: a full `sync --force` against the real FAA archive stamps the marker and offline commands work; the incomplete-archive path errors cleanly.

omarshahine:
Fully atomic now in `da4ad8b`. `importZip` runs every table DELETE/INSERT, the FTS rebuild, and the `sync_complete` marker inside a **single transaction** — `importCSV` and `rebuildFTS` take the caller's `*sql.Tx` and no longer commit on their own. If anything fails (bad row, FTS rebuild error, killed process), the whole transaction rolls back and the previous good sync is preserved intact — base tables and `faa_master_fts` never diverge. New `TestFailedReimportRollsBack` asserts that a failed re-import leaves the original data present, the partial row absent, and the search index consistent. Verified live: a full `sync --force` (~900K rows) commits in one transaction in ~31s.

PR #1461, opened 2026-07-08
feat(installer): parallel installs, --all/--category selectors, progress output

cuzo151:
## Summary

- `install` now runs up to 6 CLIs concurrently, reusing the existing `mapWithConcurrency` helper with the same cap and buffered-output pattern `update` already uses. Per-CLI output is buffered and replayed in input order so parallel runs never interleave lines.
- New selectors: `--all` (whole catalog) and `--category <cat>` (repeatable; unknown categories error with the list of known ones). Overlapping selectors dedupe to first occurrence.
- `[k/N] <name>: ok|failed` progress lines on stderr keep long bulk installs live (suppressed under `--json`).
- Go is now detected once up front for the whole invocation instead of once per CLI, so a missing toolchain fails fast with a single message.

## What Changed

- `npm/src/commands/install.ts`: registry fetch moved before selector expansion; serial for-loop replaced with a concurrent map (single-target path still streams directly); `--all`/`--category` parsing; single Go precheck; memoized `detectGo`.
- `npm/src/cli.ts`: help text for the new flags.
- `npm/tests/install.test.ts`: 7 new tests — concurrency overlap (asserts >1 in flight), contiguous per-CLI output under out-of-order completion, `--all`, `--category` (match + unknown), progress lines, single-Go-check fail-fast. 91 → 98, all passing.

## Benchmarks (M2-class Mac)

| Scenario | serial v0.1.18 | this branch |
|---|---|---|
| 6 fresh CLIs, cold modules, `--cli-only` | ~2 min+ (~21 s/CLI measured baseline) | **51.7 s** (575% CPU) |
| 3 CLIs, fully warm | 4.8 s | 3.6 s |

## Validation

- [x] `npm test` — 98/98 pass
- [x] Real-world run: `install agent-capture cal-com docker-hub dub ebay postman-explore --cli-only` — 6/6 installed, output contiguous per CLI

(Deleted the Published CLI Metadata section — this PR touches only `npm/`, no `library/` entries.)

## AI / Automation Disclosure

This PR was implemented with Claude Code, reviewed and directed by @cuzo151.

### DIFF
--- a/npm/CHANGELOG.md
+++ b/npm/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.1.19
+
+- Multi-name installs run up to 6 CLIs concurrently (same cap and buffered-output pattern as `update`); per-CLI output is replayed in input order and `[k/N]` progress lines on stderr keep long runs live.
+- New install selectors: `--all` (whole catalog) and `--category <cat>` (repeatable). Overlapping selectors dedupe.
+- Go is detected once per invocation instead of once per CLI, so a missing toolchain fails fast with a single message.
+
 ## 0.1.17
 
 - Raise the documented direct `go install` floor to Go 1.26.4 so npm installer guidance matches the catalog-wide module enforcement floor.
--- a/npm/README.md
+++ b/npm/README.md
@@ -82,11 +82,18 @@ npx -y @mvanhorn/printing-press-library install espn
 npx -y @mvanhorn/printing-press-library install airbnb-pp-cli
 ```
 
-Several at once (bundles and CLI names mix freely):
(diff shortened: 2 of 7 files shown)


tmchow:
Maintainer fix pushed in `028277040`: the PR's parallel bulk-install approach is good, but Greptile correctly caught that the original worker pool also parallelized `installSkill`, which can race against the shared global skills store. I kept Go installs concurrent, wrapped the bulk-path skill installer in a per-invocation async queue, and added a regression test showing Go installs can overlap while skill installs stay serialized.\n\nLocal validation: `npm test` passes 99/99 and `git diff --check` is clean. Greptile re-reviewed the new head at 5/5 with no unresolved threads; GitHub Actions were still queued repo-wide when I last checked.

PR #1471, opened 2026-07-09
feat(plane): reprint at engine 4.29.0 with publish evidence (R6/R7 preserved)

sidorovanthon:
# plane — Printing Press CLI (reprint at 4.29.0 with full publish evidence)

**Publication path:** Reprint/replace — update of existing PR #1471 (`plane` is already merged in `library/project-management/plane`; this PR replaces it with a reconciled reprint).

Every Plane workspace as a typed, offline-syncable CLI — plus issue relations and module membership the REST API hides.

- **API:** plane · **Category:** project-management · **CLI:** `plane-pp-cli` · **Auth:** api_key
- **Printing Press version:** 4.29.0 · **Run:** `20260724-124302-67128fc0`

## What this update is

A full reprint of the previously reviewed R6/R7 tree on cli-printing-press 4.29.0, seeded from this PR's own head (`cc4c6aed2`) and reconciled so the hand-written novel work is preserved on top of current generator output. The prior head's edits are superseded semantically, not byte-for-byte: 123 files differ against the old head (generated `internal/cli/**`, `SKILL.md`, `go.mod`, `.printing-press.json`). Three historical manuscript files from run `20260602-073610` are preserved byte-identical. The gates below verify the result end to end.

## Novel commands (from `.printing-press.json`, manifest order)

1. **`relations` — Issue Relations:** Flat list/set/unset of issue blocking, blocked_by, duplicate, relates_to and temporal links without hand-driving the relations endpoint (unset bridges Plane CE's missing DELETE via a configurable Django-shell call).
2. **`module` — Module Membership & Sync Enrichment:** Manage an issue's module membership (of/create-issue/sync) and auto-enrich issue module_ids during sync (on by default), since Plane's issue API never returns module membership.

The tree also carries hand-written `attach-file` (presign + upload + mark-uploaded in one call), `workspaces`/`init` onboarding (Plane's API cannot enumerate workspaces by key), and a sync wrapper for module enrichment — all recorded in `.printing-press-patches/`.

## Validation (publish validate, all checks PASS)

manifest · transcendence · phase5 · go mod tidy · govulncheck · go vet · go build · --help · --version · verify-skill · manuscripts — **passed: true**.

```
Plane CLI — Every Plane workspace as a typed, offline-syncable CLI — plus issue
relations and module membership the REST API hides.
Health check: run 'plane-pp-cli doctor' to verify auth and connectivity.
```

## Live end-to-end gate (Step 4.5)

`dogfood --live --level full` against the real Plane API with a live api_key: **303/303 tests PASS, verdict PASS** (`publish-live-gate.json`), acceptance marker `phase5-acceptance.json` = `pass` / `full`. The earlier attempts (`publish-live-gate.attempt1–4.json`) are committed as the drive-to-green history.

## Fixes discovered by the publish gates

The live gate surfaced two real bugs in the tree; both are fixed here and ledgered as patches:

- **`plane-init-no-config-clobber-on-failure`** — a failed `init` used to persist a base_url derived from the default cloud host before probing, silently repointing a self-hosted setup at Plane Cloud (this is what failed 12 downstream live tests). Failed `init` now leaves config and credential files byte-for-byte untouched; covered by `internal/cli/init_config_preserve_test.go`.
- **`plane-sync-warning-nul-json`** — dependent-resource sync/reconcile NDJSON events interpolated NUL-composite storage ids raw into JSON lines, breaking `sync --json` fidelity. All id/scope fields now render through `json.Marshal`; covered by `internal/cli/sync_event_json_test.go`.

## Manuscripts (run `20260724-124302-67128fc0`)

Research:
- [research.json](https://github.com/mvanhorn/printing-press-library/blob/c66fad3a7a6b895de2fce19e3cca51160be93979/library/project-management/plane/.manuscripts/20260724-124302-67128fc0/research.json)
- [plane-spec.yaml](https://github.com/mvanhorn/printing-press-library/blob/c66fad3a7a6b895de2fce19e3cca51160be93979/library/project-management/plane/.manuscripts/20260724-124302-67128fc0/plane-spec.yaml)

Proofs:
- [publish-live-gate.json](https://github.com/mvanhorn/printing-press-library/blob/c66fad3a7a6b895de2fce19e3cca51160be93979/library/project-management/plane/.manuscripts/20260724-124302-67128fc0/proofs/publish-live-gate.json) (303/303 PASS)
- [phase5-acceptance.json](https://github.com/mvanhorn/printing-press-library/blob/c66fad3a7a6b895de2fce19e3cca51160be93979/library/project-management/plane/.manuscripts/20260724-124302-67128fc0/proofs/phase5-acceptance.json)
- [publish-live-gate.attempt1.json](https://github.com/mvanhorn/printing-press-library/blob/c66fad3a7a6b895de2fce19e3cca51160be93979/library/project-management/plane/.manuscripts/20260724-124302-67128fc0/proofs/publish-live-gate.attempt1.json)
- [publish-live-gate.attempt2.json](https://github.com/mvanhorn/printing-press-library/blob/c66fad3a7a6b895de2fce19e3cca51160be93979/library/project-management/plane/.manuscripts/20260724-124302-67128fc0/proofs/publish-live-gate.attempt2.json)
- [publish-live-gate.attempt3.json](https://github.com/mvanhorn/printing-press-library/blob/c66fad3a7a6b895de2fce19e3cca51160be93979/library/project-management/plane/.manuscripts/20260724-124302-67128fc0/proofs/publish-live-gate.attempt3.json)
- [publish-live-gate.attempt4.json](https://github.com/mvanhorn/printing-press-library/blob/c66fad3a7a6b895de2fce19e3cca51160be93979/library/project-management/plane/.manuscripts/20260724-124302-67128fc0/proofs/publish-live-gate.attempt4.json)
- [regen-merge-report.json](https://github.com/mvanhorn/printing-press-library/blob/c66fad3a7a6b895de2fce19e3cca51160be93979/library/project-management/plane/.manuscripts/20260724-124302-67128fc0/proofs/regen-merge-report.json)
- [regen-merge-dry-run-report.json](https://github.com/mvanhorn/printing-press-library/blob/c66fad3a7a6b895de2fce19e3cca51160be93979/library/project-management/plane/.manuscripts/20260724-124302-67128fc0/proofs/regen-merge-dry-run-report.json)
- [regen-merge-apply.err](https://github.com/mvanhorn/printing-press-library/blob/c66fad3a7a6b895de2fce19e3cca51160be93979/library/project-management/plane/.manuscripts/20260724-124302-67128fc0/proofs/regen-merge-apply.err)
- [regen-merge-dry-run.err](https://github.com/mvanhorn/printing-press-library/blob/c66fad3a7a6b895de2fce19e3cca51160be93979/library/project-management/plane/.manuscripts/20260724-124302-67128fc0/proofs/regen-merge-dry-run.err)
- [drift-files.txt](https://github.com/mvanhorn/printing-press-library/blob/c66fad3a7a6b895de2fce19e3cca51160be93979/library/project-management/plane/.manuscripts/20260724-124302-67128fc0/proofs/drift-files.txt)

Historical run archive `20260602-073610` (preserved byte-identical from the prior head):
- [phase5-acceptance.json](https://github.com/mvanhorn/printing-press-library/blob/c66fad3a7a6b895de2fce19e3cca51160be93979/library/project-management/plane/.manuscripts/20260602-073610/proofs/phase5-acceptance.json)
- [2026-06-02-073610-feat-plane-pp-cli-brief.md](https://github.com/mvanhorn/printing-press-library/blob/c66fad3a7a6b895de2fce19e3cca51160be93979/library/project-management/plane/.manuscripts/20260602-073610/research/2026-06-02-073610-feat-plane-pp-cli-brief.md)
- [research.json](https://github.com/mvanhorn/printing-press-library/blob/c66fad3a7a6b895de2fce19e3cca51160be93979/library/project-management/plane/.manuscripts/20260602-073610/research/research.json)

## README (excerpt)

> A terminal CLI for [Plane](https://plane.so) — open-source project management for issues, cycles, modules, sub-issues, and workspaces; a self-hostable Jira / Linear / ClickUp alternative. Ships with offline SQLite sync for instant search and analytics, plus an MCP server (`plane-pp-mcp`) so agents can drive Plane directly.
>
> Pre-built binaries for Linux, macOS, and Windows ship in the `plane-current` release, so you can install with no Go or Node toolchain.

## Gaps

- `spec_url` is not present in the manifest (the spec itself is committed as `.manuscripts/20260724-124302-67128fc0/plane-spec.yaml`).
- `help_output`'s "Highlights" bullets render the novel command names without descriptions; the manifest `novel_features` descriptions are the authoritative text.

https://claude.ai/code/session_016U2vwYCDk5UnrAUXBAsvXN


sidorovanthon:
Addressed the Greptile P1 (workspace-boundary) in `c95f37522`: `plane_execute` now declares the top-level `workspace` argument in its tool schema, so schema-validating MCP clients no longer strip the per-call override before `applyWorkspaceArg` runs. The arg is consumed by the handler and never forwarded to the endpoint path/query/body. Verified live — `tools/list` now advertises `plane_execute.workspace`, and a per-call `workspace` override routes correctly (`plane_execute projects.list workspace=bbm` → 8 BBM projects vs `workspace=doctor-school` → 10 Doctor-School projects).

tmchow:

sidorovanthon:
Addressed the two remaining Greptile findings in `0c1655a9e`:

**① `store.go` `isoDatePattern` regex — FIXED.** The time-portion char class was `[0-9:.+-Zz]`, where `+-Z` is an unintended ASCII range (0x2B–0x5A) that swallowed `, / ; < = > ? @ A-Z`. `shouldIndexSearchString` therefore misclassified date-prefixed non-date strings (`2024-01-15 PLANNED`, date-prefixed slugs) as dates and excluded them from the FTS content index. Moved `-` to the end of the class (`[0-9:.+Zz-]`) so it's a literal. Real ISO dates/timestamps still match, so nothing correctly excluded starts leaking in. Added `store_novel_test.go` as a RED-without-the-fix regression guard (fails on the old range for `2024-01-15 A?/=@` and `2024-01-15 PLANNED`, passes after). Live A/B on the rebuilt CLI (isolated `--home`, prod `data.db` untouched): bbm sync 319 records 10/10 resources 0 warned/errored, local FTS `search` returns hits, exit 0.

**② `code_orch.go` cursor removed from list-endpoint QueryParams — working as intended, not a regression.** In engine 4.27 per-endpoint `cursor` params were replaced by the resumable-cursor pagination that `makeAPIHandler` drives via `mcpPageConfig` + the `bound` envelope (opaque cursor encode/decode, `MaxBytes` budget, `count` metadata). Pagination still works at the handler level (Greptile's own note); the schema no longer advertises a redundant `cursor` knob that the new envelope owns. No action taken here by design.

Please re-review the current head `0c1655a9e`.

@greptileai review

tmchow:

- missing publish-process files: library/project-management/plane/README.md, library/project-management/plane/SKILL.md

sidorovanthon:
Thanks. Fixed in a91ca91c1, and you were right about the symptom even though the files weren't literally absent.

**What actually happened.** `README.md` and `SKILL.md` exist in the branch, but they were byte-identical to `main`: the 4.27.1 reprint overwrote the templated Go and left the docs at their 4.20.1 shape. So the docs never entered the PR diff, which is what the missing-publish-process-files check saw.

**Why this isn't a straight `generate --force` of the two files.** I built the 4.27.1 engine and regenerated `plane` from this branch's `spec.yaml` to see exactly what the templates produce. A wholesale overwrite would have deleted content the 4.27.1 templates don't emit:

- the curated README lead and the `### Pre-built binary (no Node, no Go)` install path, plus the matching three-path install block in SKILL.md. These are library-side shape owned by `tools/sweep-canonical`, and they aren't in `readme.md.tmpl` / `skill.md.tmpl` even at 4.28.0.
- the hand-authored `## Workspace targeting` section in SKILL.md (slug precedence, the `PLANE_SLUG`-masquerading-as-403 trap, `plane_execute`'s `workspace` arg).
- the `### relations` and `### module` novel-command sections in README.md.

So I grafted the genuine 4.27 doc delta onto the published docs instead:

- **README**: new `## Paths & environment variables` covering the four path kinds, the per-kind-env to `--home` to `PLANE_HOME` to XDG to platform-default ladder, relocation through the MCP host `env` block, the fleet-precedence note, and the one-way-relocation warning. Credential persistence now points at `plane-pp-cli auth set-token` / `credentials.toml` under the data dir; `## Configuration` points at `doctor` for the resolved dirs.
- **SKILL**: new `## Paths and state` stating the same contract in agent terms, including the schema v4 `paths` block in `agent-context`. Credential-persistence and `feedback.jsonl` locations corrected to the resolved data dir.
- Both pick up the template's `--ignore-missing` wording fix.

Net diff is +82/-6 across the two files, no line-ending churn.

**Verification**

| Check | Result |
|---|---|
| `verify_skill.py --dir library/project-management/plane` | pass (flag-names, flag-commands, positional-args, shell-var-quotes, unknown-command) |
| `check_skill_guard_literals.py .../SKILL.md` | pass |
| `verify_publish_package.py --base-ref upstream/main` | pass (1 touched CLI dir, 0 new) |

Also fixed in 3d9283f95, since it turned out to be decidable without your input: `.printing-press.json` carried `run_id: 20260709-030047`, which matches `generated_at` to the second and corresponds to no pipeline run, so the manifest pointed at a research/proofs corpus that does not exist. Every other run-id reference in the tree already said `20260602-073610`: the shipped `.manuscripts/20260602-073610/{research,proofs}`, the `run_id` inside `proofs/phase5-acceptance.json`, `.printing-press-release.json`, and the `base_run_id` on all eleven `.printing-press-patches/` entries. The manifest was the lone outlier, so I pointed it back at the manuscripts the tree actually ships.

That also matches what the generator writes for a regen with no run state: `pipeline.WriteManifest` carries the existing manifest's `run_id` forward instead of minting one from the clock, precisely so the manifest keeps resolving to the shipped manuscripts. `generated_at` and `printing_press_version` stay as the record of when this tree was printed and on which engine. No manuscripts were fabricated for a run that never happened, and `.printing-press-release.json` is left to the post-merge automation that owns it.


Third item, `0ce62486f`. While checking whether the changed `spec_url` / `spec_path` / `spec_checksum` in this PR were legitimate, two things turned up.

The `spec_url` change is required, not incidental. The value on main pointed at `cli-printing-press/main/catalog/specs/plane-spec.yaml`, which now returns 404: the embedded catalog was removed upstream in mvanhorn/cli-printing-press#3385. So main's manifest referenced a dead source, and its `spec_checksum` described that catalog file rather than the `spec.yaml` shipped beside it. Repointing both at the library's own `spec.yaml` is the fix.

The `spec_checksum` value was wrong in a subtler way. It recorded `sha256:ca2852c5...`, the hash of `spec.yaml` on a Windows checkout with `core.autocrlf=true`. The blob in git, which is what CI and every Linux consumer sees, is LF and hashes to `sha256:61501162...`. That field is load-bearing: `sameGenerateManifestLineage` returns on checksum equality, and the `--force` regen cross-spec guard compares the snapshot manifest's checksum against a freshly computed one, so a same-spec regen on a different platform concludes "different spec" and drops carried-forward manifest metadata. Corrected to the LF value, which now matches `git show HEAD:library/project-management/plane/spec.yaml | sha256sum` exactly.

The generator-side cause is filed as mvanhorn/cli-printing-press#3689: all three checksum sites hash raw file bytes with no line-ending normalization, so any print from a Windows checkout embeds a platform-dependent value. This PR fixes the one printed CLI; the machine fix is what stops it recurring.

sidorovanthon:
Addressed the P1 (`Comments Outside Diff`) in `cc4c6aed2`.

**The finding was correct, and it reproduces.** `module of` joined `modules m ON m.id = mi.module_id`, but the two sides stopped agreeing at engine 4.27: `modules` is a dependent resource, so `resourceStorageID` keys its rows `<uuid>\x00<projects_id>`, while the `module_issues` junction carries the bare uuid — `localModules` strips the suffix before the insert precisely so a NUL never reaches an API path or a junction key. The JOIN matched nothing and `COALESCE(..., '')` turned every miss into an empty string, so the command printed correct ids and blank names.

Verified on a seeded fixture before changing anything:

```
OLD JOIN -> id="m1" name=""
NEW      -> id="m1" name="Sprint 1"
```

**Fix.** Replaced the JOIN with the same Go-side mapping the enrichment already uses for `projects_issues` (the `storedByBare` map), for the same stated reason — SQLite string funcs mis-handle the embedded NUL, so the composite key can't be stripped in SQL. `modulesForIssue` reads the junction rows, builds bare-id → name from `modules`, and resolves each link by `(project, module)` where the junction carries a project, falling back to the bare id for rows whose nullable `project_id` is unset (the `module create-issue` cache path can write one). Ordering is preserved — name, then id as a deterministic tie-break, which the old `ORDER BY 2` left unspecified — and the JSON shape is unchanged.

**Test.** `TestModulesForIssue_ResolvesCompositeModuleName` seeds a composite-keyed module against a bare junction row and asserts the resolved name, covering both the parent-qualified lookup and the bare-id fallback plus the ordering. It also guards the premise: if `modules` ever stops being parent-keyed, it fails loudly instead of passing for the wrong reason.

I swept the rest of the tree for the same shape — `grep` for SQL touching `modules` / `projects_issues` returns three sites: `localModules` (strips correctly), the enrichment's id map (already fixed), and this JOIN. No other hand-authored query compares a bare id against a composite-keyed table.

`go build ./...`, `go vet ./...`, and `go test ./internal/cli/ ./internal/store/ ./internal/client/` pass on the change. (Local-only note: `internal/learn` and `internal/mcp` tests fail on a Windows checkout because `t.Setenv("HOME", ...)` doesn't redirect `os.UserHomeDir()` there, so they read the real state dir and DB. Pre-existing on the base commit, unrelated to this change, and green in CI — I'm filing it machine-side, like the `spec_checksum` issue.)

sidorovanthon:
Follow-up: verified the fix against a **real synced store**, not just the unit fixture. Names redacted; ids are real.

The interesting part is that a store which has lived across the engine upgrade is *mixed* — 147 module rows, 122 still under the old bare id and 25 re-keyed to the 4.27 NUL-composite id (a full re-sync hasn't reconciled the stale rows away yet). Of 547 junction links, **246 point at a module that exists only under a composite id**, which is exactly the set the old JOIN could not resolve.

Same issue, same database, before and after:

```
old JOIN  -> [("1f8c44cf-…", ""), ("dccfb326-…", "")]
module of -> [{"id": "dccfb326-…", "name": "<module A>"},
              {"id": "1f8c44cf-…", "name": "<module B>"}]
```

So on a real upgraded store this was not an edge case — it was blank names on roughly half the links, silently. `moduleNamesByBareID` resolves all three store states (bare-only, composite-only, and the mixed state above), since it keys off `store.BareResourceID` rather than assuming either shape.

Worth noting separately, and *not* something this PR changes: the mixed state itself means an upgraded store keeps duplicate module rows (stale bare + new composite) until a reconciling `sync --full`. That is engine behavior, not CLI behavior, and `module of` is now correct either way.

sidorovanthon:
Ready for maintainer review. The `needs-author` label and the last triage marker (`sha=0c1655a9e259`) both predate the two pushes since; I don't have triage permission to relabel.

State at head `cc4c6aed2`:

- **Greptile 5/5**, reviewed the current head, "Safe to merge — both previously-flagged blocking issues are fully addressed". No actionable markers, no unresolved threads, nothing in `Comments Outside Diff`.
- **All 7 checks pass**: Verify, Govulncheck, Scan, Greptile Review, Greptile policy gate, Validate MCPB manifest contract, Validate SKILL.md against shipped CLI source. `MERGEABLE / CLEAN`.
- `python3 .github/scripts/pr-review-state/greptile_feedback.py 1471` → `READY: Greptile feedback is resolved for the current PR head`.

Both triage blockers from 2026-07-20 are closed: the publish-process files (`README.md`, `SKILL.md`) entered the diff in `a91ca91c1` — they existed on the branch but were byte-identical to `main`, so a docs-shape reprint miss read as absent files — and the `module of` P1 is fixed in `cc4c6aed2` with a regression test and live verification against a real synced store.

PR shape re-checked against `AGENTS.md`: no `registry.json` or `cli-skills/` in the diff, release-ledger files (`CHANGELOG.md`, `.printing-press-release.json`) untouched, `.manuscripts/20260602-073610/{research,proofs}` shipped and matching the manifest `run_id`, and `spec_checksum` verified byte-for-byte against the `spec.yaml` blob in this commit.

tmchow:

- missing .manuscripts publish evidence

sidorovanthon:
@tmchow The missing publish evidence is now in place. This head is a full reprint at cli-printing-press 4.29.0, and fresh manuscripts are committed under [`.manuscripts/20260724-124302-67128fc0/`](https://github.com/mvanhorn/printing-press-library/blob/c66fad3a7a6b895de2fce19e3cca51160be93979/library/project-management/plane/.manuscripts/20260724-124302-67128fc0/research.json) — research plus proofs, including [`publish-live-gate.json`](https://github.com/mvanhorn/printing-press-library/blob/c66fad3a7a6b895de2fce19e3cca51160be93979/library/project-management/plane/.manuscripts/20260724-124302-67128fc0/proofs/publish-live-gate.json) (live `dogfood --live --level full`: 303/303 PASS) and [`phase5-acceptance.json`](https://github.com/mvanhorn/printing-press-library/blob/c66fad3a7a6b895de2fce19e3cca51160be93979/library/project-management/plane/.manuscripts/20260724-124302-67128fc0/proofs/phase5-acceptance.json) (`pass` / `full`). `publish validate` is fully green (manifest, phase5, manuscripts, verify-skill and the build gates all PASS). The historical run archive [`.manuscripts/20260602-073610/`](https://github.com/mvanhorn/printing-press-library/blob/c66fad3a7a6b895de2fce19e3cca51160be93979/library/project-management/plane/.manuscripts/20260602-073610/research/research.json) is preserved byte-identical. Two real bugs the gates surfaced (failed `init` clobbering config, NUL-composite ids breaking `sync --json` NDJSON) were fixed and ledgered as `.printing-press-patches/plane-init-no-config-clobber-on-failure.json` and `plane-sync-warning-nul-json.json`. Full details in the updated PR description.

tmchow:

sidorovanthon:
@tmchow The `needs-author` triage here is reading a stale check, not a new problem on the branch. No commit has landed on `c66fad3a7a6b895de2fce19e3cca51160be93979` since 2026-07-24; what changed is that Greptile finally answered.

Timeline on this head:

- **07-24 08:23** `Greptile policy gate` starts ([run 30078768658](https://github.com/mvanhorn/printing-press-library/actions/runs/30078768658/job/89435324761)) and begins polling for a `Greptile Review` check.
- **07-24 08:26** the gate posts its own `@greptileai review` nudge (#issuecomment-5067763162), because at 305 files this PR is over Greptile's auto-review size cap.
- **07-24 08:49** the gate hits its 25-minute inner deadline and fails with `Timed out waiting for Greptile Review to complete for c66fad3a…`. That timeout is the only reason for the FAILURE. No score was ever read, no finding was raised.
- **07-29 01:33** Greptile acts on the nudge: the `Greptile Review` check on `c66fad3a` concludes **success**, and its summary comment is rewritten against this head with **Confidence Score: 5/5, "Safe to merge"**.

So the red `Greptile policy gate` predates the green `Greptile Review` by five days, and nothing re-ran it.

Re-running the gate as-is should pass. Walking `greptile-policy-gate.yml` against current state:

1. The poll loop finds the `Greptile Review` check for `c66fad3a` with `status=completed`, `conclusion=success`, and breaks instead of timing out.
2. The Python step needs the latest `greptile-apps[bot]` comment carrying `Confidence Score:` to contain the head SHA. Comment 4921558945 does contain `c66fad3a7a6b895de2fce19e3cca51160be93979` (it was rewritten at 01:33 today).
3. It parses `5/5` against `MIN_GREPTILE_SCORE=4`, so `5 >= 4` passes.

Every other check on this head is green: Verify, Govulncheck, Scan, Validate MCPB manifest contract, Validate SKILL.md against shipped CLI source. The PR is `MERGEABLE` with no conflicts against `main`; `mergeStateStatus` is `BLOCKED` solely on the stale gate.

**What I need from a maintainer:** a re-run of [run 30078768658](https://github.com/mvanhorn/printing-press-library/actions/runs/30078768658), or a `workflow_dispatch` of `Greptile policy gate` with `pr=1471`. My permissions on this repo are `pull` only, so I can trigger neither.

I am deliberately not pushing an empty commit to force fresh checks. A new SHA goes back over the Greptile size cap, the gate waits 25 minutes for a review that arrives days later, and we land in exactly this state again, except without a review that matches the head.

Separately, this looks like a reproducible trap in `greptile-policy-gate.yml` for large reprint PRs: the gate knows how to nudge Greptile past the size cap, but its deadline is far shorter than the nudge turnaround, and nothing re-evaluates the gate when the review lands afterward. Happy to file that as its own issue if useful.

sidorovanthon:
@tmchow Gentle ping on this one. Nothing has changed on my side since 2026-07-29, so this is a reminder rather than new information.

Head is still `c66fad3a7a6b895de2fce19e3cca51160be93979`, unchanged since 07-24. On that exact SHA:

- `Greptile Review` — SUCCESS (07-29 01:33). No actionable markers, no unresolved threads, nothing in `Comments Outside Diff`.
- `Greptile policy gate` — FAILURE (07-24 08:49), solely because its 25-minute inner deadline expired while it was waiting for that `Greptile Review` check to appear. At 305 files this PR is over Greptile's auto-review size cap, so the gate posted its own `@greptileai review` nudge and waited; the review landed five days after the gate had already given up. The gate does not recompute retroactively, so it still shows the stale failure.
- The other six checks are green and the PR is `MERGEABLE`.

A re-run of the `Greptile policy gate` job on `c66fad3a` should pass now that the `Greptile Review` check it polls for exists. I don't have permission to re-run the workflow or to clear the `needs-author` label.

I have deliberately not pushed an empty commit to force a fresh run: a new head SHA would discard the review that already landed and would very likely hit the same size-cap timeout again. If you'd prefer a different route — splitting the reprint, or anything else that keeps the gate under the cap — I'm happy to do that work.

tmchow:
You were right about the stale Greptile policy gate. Live gate is 303/303, Greptile is 5/5, and the other checks are green. Merging.

sidorovanthon:
Thanks!

PR #1513, opened 2026-07-13
feat(snipd): add snipd

maxswinguy:
## snipd

Turn your Snipd snips into a local, ranked, full-text-searchable corpus — search a concept across every note, quote, and transcript, pull the exact quote, and synthesize across shows from the command line, plus an MCP for agents.

**API:** snipd | **Category:** media-and-entertainment | **Press version:** 4.28.0
**Spec:** reverse-engineered internal spec (`snipd-internal-spec.yaml`, `spec_format: internal`) — Snipd has no public/official API, so the spec is reconstructed from its sanctioned Obsidian-export endpoints. A reverse-engineered internal spec is a first-class Printing Press input, the same path Medium Reader and Substack Reader took.

### Publication Path

New print.

### Update — token-retrieval docs fix (`f81d62f0`)

A docs-only follow-up on the reviewed tree. v1.0 described the `SNIPD_TOKEN` as "obtained by pairing from the Snipd mobile app" — that's inaccurate. The token is minted by a one-time **browser sign-in** to your Snipd account; it needs neither the mobile app, Obsidian, nor the Snipd Obsidian plugin (the page is titled "Obsidian integration" only because that's Snipd's internal name for the export API). Verified three ways: Snipd's open-source Obsidian plugin source, a live view of the sign-in page, and the original probe artifacts.

Corrected across README (Authentication + 401), SKILL (Auth Setup), the `pull`/`client` runtime hints, and `auth setup` (previously a "No setup URL is configured" dead-end). The generated-file edit (`auth.go`) is recorded in `.printing-press-patches/snipd-auth-token-docs-fix.json`; no tool-surface, manifest, or dependency change. Re-validated: `publish validate` 11/11, live `--level quick` 16/16, govulncheck clean. *(This push re-ran Greptile, which re-surfaced advisory comments on the v1.0 reader layer — date filters, snip-split, synthetic snip id — not the docs change; the policy gate passed.)*

### How this was built — generated chassis + hand-extended reader

A note on process, in my own words: I'm a UX/service designer, not a developer. I build these CLIs with Claude (Claude Code) doing the implementation while I scope and review each step. I'm stating the human/AI split and the generated/hand split up front so the provenance is legible rather than something a reviewer has to reconstruct.

- **Chassis — generated by `cli-printing-press` 4.28.0.** 126 `.go` files carry the `// Generated by CLI Printing Press (…). DO NOT EDIT.` header — config, the JSON client, the generic SQLite store, the MCP cobratree + typed-tool surface, the learn/teach/recall/playbook surface, and root/doctor/version. The provenance is press-emitted, not hand-authored: `.printing-press.json` carries a sub-second `generated_at`, a real `spec_checksum`, and run_id `20260713-190747-d8d38c3e`; `.manuscripts/` (research brief, absorb manifest, `research.json`) and the Phase-5 acceptance proof all come from that one generate run.

- **Hand-extension — the export-reader + retrieval layer, by necessity.** Snipd's export API is **RPC-shaped, not REST-shaped**: a `GET` returns the export catalog (episode ids, snip counts, batches) and a `POST` returns a **ZIP of labelled-delimiter Markdown**, not JSON. The generated JSON client can't consume a ZIP-of-Markdown body, so the source + retrieval layer is hand-written — the same division as Medium Reader and Substack Reader (both accepted here), where an undocumented/RPC surface always leaves the reader as the hand layer. 8 source files carry `// Hand-authored novel command — NOT generator output`: `internal/snipd/{parse,client}.go` and `internal/cli/{pull,quote,synthesize,filter,aggregate,snipd_corpus}.go` (plus their tests). Each is indexed in `.printing-press-patches/snipd-source-and-retrieval-commands.json`, the durable hand-edit contract a future regen reads.

- **Scope guardrail — read-only and generic.** One personal downstream workflow — exporting curated snips into my own Tana Outliner — is deliberately **excluded** so the published surface stays generic. The CLI never writes to Snipd; it reads your own account's export via your Bearer token (`SNIPD_TOKEN`), obtained by a one-time browser sign-in to your Snipd account (no mobile app, Obsidian, or plugin required — see the README's Authentication section).

### CLI Shape
```bash
$ snipd-pp-cli --help
Snipd CLI — Turn your Snipd snips into a local, ranked, full-text-searchable corpus — search a concept across every note, quote, and transcript, pull the exact quote, and synthesize across shows from the command line, plus an MCP for agents.

  • pull   Pull your Snipd snips into a local SQLite mirror you can search and query offline — the app has no export-to-query path.
  • quote   Get the verbatim pull-quote and speaker for matching snips — the distinct unit other tools concatenate into transcript prose and lose.
  • synthesize   Gather a deduped, snippet-only evidence set (title + note + quote + deep-link) scoped by query, show, or topic — sized for an agent to reason over.
  • filter   Slice the corpus by show, favorite, tag, date, or duration and return compact rows or counts.
  • aggregate   Roll up counts and distributions across your corpus — snips per show, top tags, favorites, episode density, and timeline.

Health check: run 'snipd-pp-cli doctor' to verify auth and connectivity.

Usage:
  snipd-pp-cli [command]

Available Commands:
  aggregate      Roll up counts and distributions across your corpus — snips per show, top tags, favorites, snips per episode.
  analytics      Run analytics queries on locally synced data
  auth           Manage authentication for Snipd
  catalog        List your episodes available for export (ids, snip counts, batches).
  filter         Slice the corpus by show, favorite, tag, or date and return compact rows or a count.
  pull           Pull your Snipd snips into a local SQLite mirror you can search and query offline — the app has no export-to-query path.
  quote          Get the verbatim pull-quote and speaker for matching snips — the distinct unit other tools concatenate into transcript prose and lose.
  search         Search locally synced data
  synthesize     Gather a deduped, snippet-only evidence set (title + note + quote + deep-link) scoped by query, show, or topic.
  tail           Stream live changes by polling the API at regular intervals
  version        Print version

Flags:
  -h, --help                 help for snipd-pp-cli
      --no-learn             Disable the teach/recall learning loop for this invocation
      --profile string       Apply values from a saved profile (see 'snipd-pp-cli profile list')
  -v, --version              version for snipd-pp-cli

Use "snipd-pp-cli [command] --help" for more information about a command.

```

### Novel Commands

| `pull` | Build the local snip corpus | Pull your Snipd snips into a local SQLite mirror you can search and query offline — the app has no export-to-query path. |
| `quote` | Pull the exact quote | Get the verbatim pull-quote and speaker for matching snips — the distinct unit other tools concatenate into transcript prose and lose. |
| `synthesize` | Scoped synthesis feed | Gather a deduped, snippet-only evidence set (title + note + quote + deep-link) scoped by query, show, or topic — sized for an agent to reason over. |
| `filter` | Structured filter | Slice the corpus by show, favorite, tag, date, or duration and return compact rows or counts. |
| `aggregate` | Corpus rollups | Roll up counts and distributions across your corpus — snips per show, top tags, favorites, episode density, and timeline. |

### What This CLI Does

Your snips are trapped in a mobile app. This CLI pulls them into a local SQLite mirror with a full-text index, then lets you search, quote, filter, and aggregate them in kilobytes instead of scrolling an app. The same commands are exposed as MCP tools so an agent can reason over your own listening.

Store: episodes + snips live in the generic `resources` table with a porter FTS index; retrieval is bm25-ranked over note/quote/transcript. `search` and `sql` are framework tools; `pull` and the four retrieval commands above are the hand-built transcendence surface. Read-only throughout — the CLI never mutates your Snipd account.

### Manuscripts

- [Research Brief](https://github.com/mvanhorn/printing-press-library/blob/1e9f8f8474accfad2d3bdc3fb39f25fcd8dffa29/library/media-and-entertainment/snipd/.manuscripts/20260713-190747-d8d38c3e/research/2026-07-13-feat-snipd-pp-cli-brief.md)
- [Absorb Manifest](https://github.com/mvanhorn/printing-press-library/blob/1e9f8f8474accfad2d3bdc3fb39f25fcd8dffa29/library/media-and-entertainment/snipd/.manuscripts/20260713-190747-d8d38c3e/research/2026-07-13-feat-snipd-pp-cli-absorb-manifest.md)
- [Research JSON (absorb / transcend ledger)](https://github.com/mvanhorn/printing-press-library/blob/1e9f8f8474accfad2d3bdc3fb39f25fcd8dffa29/library/media-and-entertainment/snipd/.manuscripts/20260713-190747-d8d38c3e/research.json)
- [Phase 5 Acceptance (live, quick)](https://github.com/mvanhorn/printing-press-library/blob/1e9f8f8474accfad2d3bdc3fb39f25fcd8dffa29/library/media-and-entertainment/snipd/.manuscripts/20260713-190747-d8d38c3e/proofs/phase5-acceptance.json)

### Validation Results

| Check | Result |
|-------|--------|
| Manifest | PASS |
| Transcendence | PASS |
| Phase 5 | PASS |
| go mod tidy | PASS |
| go vet | PASS |
| go build | PASS |
| --help | PASS |
| --version | PASS |
| verify-skill | PASS |
| Manuscripts | PRESENT |

`publish validate` passes 11/11 (with `GOFLAGS=-buildvcs=false`, since the local library tree isn't a git repo — a known 4.28.0 build-stamping quirk, not a CLI defect).

### Publish Live Gate

Live dogfood reran against the real Snipd export API at publish time and passed: **`--level quick`, 16/16 passed, 8 skipped**. Proof: [phase5-acceptance.json](https://github.com/mvanhorn/printing-press-library/blob/1e9f8f8474accfad2d3bdc3fb39f25fcd8dffa29/library/media-and-entertainment/snipd/.manuscripts/20260713-190747-d8d38c3e/proofs/phase5-acceptance.json).

Disclosure on the level: I ran `--level quick`, not `--level full`, deliberately. On `cli-printing-press` 4.28.0 the default-on self-learning commands (`teach`, `teach-pattern`, `teach-playbook`, `playbook amend`) are required-flag commands that a `--level full` happy-path probe invokes bare, so they exit 2 and are counted as failures — a generator-side matrix issue affecting every 4.28.0 CLI carrying the learn surface, not a defect in this CLI. `--level quick` is a gate-accepted level and exercises the real read path end to end. (Same disclosure as the Substack Reader 4.28.0 reprint.)

### Gaps

- **`tools-manifest.json` regenerated from the live MCP.** The press emits a spec-derived `tools-manifest.json` (1 endpoint) and `.printing-press.json` `mcp_tool_count: 1`; the live MCP exposes many more. I regenerated `tools-manifest.json` from the live `tools/list` so it reflects the real surface: the 14 domain + framework tools (`pull`, `search`, `sql`, `sync`, `context`, `workflow*`, and the retrieval commands), with the framework self-learning meta-surface (`teach*`/`learnings*`/`recall`/`playbook*`) excluded — matching the accepted Substack Reader (14) and Medium Reader (10) manifests. I left the press-emitted `mcp_tool_count` untouched rather than hand-editing provenance (Substack Reader ships `mcp_tool_count: 3` with a 14-tool manifest for the same reason).


maxswinguy:
Thanks for the review — all four findings were real and are fixed in `737f766`. They all landed in the hand-authored source/retrieval layer (my code, not generator output), and each fix ships with a regression test. The live pull is behavior-preserving: a real `pull` still returns the identical corpus (3 episodes / 52 snips on the `--limit 3` probe), because real snips carry UUIDs and the server always emits `@@SNIP@@` on its own line.

**P1 — Fallback snip IDs collide** (`internal/cli/pull.go`)
Extracted the synthetic-ID logic into `fallbackSnipID(s)` and now hash the full content (`episode + start + title + note + quote + transcript`) instead of just `episode+start+title`. A re-pull of the same snip is still content-stable (updates, not duplicates), but two distinct UUID-less snips can no longer collide and overwrite each other — a collision now means every content field matches, i.e. a genuine duplicate. Test: `TestFallbackSnipIDDistinguishesDistinctSnips`.

**P1 — Content can split snips** (`internal/snipd/parse.go`)
Replaced `strings.Split(body, "@@SNIP@@")` with a line-anchored split (`(?m)^@@SNIP@@$`). The export template emits `@@SNIP@@` alone on its own line, so this is identical for real exports but no longer treats the token appearing inside a note/quote/transcript as a boundary. Test: `TestParseEpisodeSeparatorInContentNotSplit` (an inline `@@SNIP@@` now stays one snip; the old bare split produced two).

**P1 — Cursor uses string ordering** (`internal/cli/pull.go`)
The incremental `--updated-after` cursor now compares parsed instants via `tsLater`, falling back to lexical order only when a value isn't a parseable timestamp (e.g. the empty initial cursor). Mixed offsets or fractional precision can no longer park the cursor before episodes it should cover. Test: `TestTsLater` (includes a same-instant-different-offset case that lexical comparison would get wrong).

**P2 — Invalid dates become filters** (`internal/cli/filter.go`)
`--since`/`--until` are now validated (`validateDateFlag`) before building the SQL predicate. A typo like `--since zzzz` returns `invalid --since "zzzz": expected a date (2006-01-02) or an RFC3339 timestamp` instead of silently running as a lexical predicate and returning an empty/misleading slice. Test: `TestValidateDateFlag`.

Full suite, `go vet`, `publish validate` (11/11), and a fresh live `dogfood --level quick` (16/16) are green after the changes.

— Disclosure, same as the PR description: I'm a designer, not a developer; Claude (Claude Code) implemented these fixes under my direction and I reviewed each one.

maxswinguy:
Good catch on the follow-up — fixed at the root in `cc84c276`.

Rather than chase delimiters, the parser now uses the export's structural invariant: **every real snip chunk opens with a `snip_<token>=<<` field** (verified against the live export — even non-favorited snips emit an empty `snip_favorite_star=<<…>>` first). So after splitting on standalone `@@SNIP@@` lines, any chunk that does *not* open a snip field is a phantom produced by a `@@SNIP@@` sitting inside a multiline value; it's glued back onto the previous chunk (restoring the exact separator) instead of becoming its own snip. This also covers a `@@SNIP@@` inside an episode-level field (it reattaches to the head).

Regression test `TestParseEpisodeStandaloneSeparatorInContentNotSplit` pins a note that contains a lone `@@SNIP@@` line: one snip, note content intact. A real `pull` still returns the identical corpus (3 episodes / 52 snips on the probe). Full suite + `go vet` + `publish validate` (11/11) + live `dogfood --level quick` (16/16) green.

maxswinguy:
Right again — accepting a timestamp while comparing it lexically against a date-only `publish_date` was a bug I introduced with the previous validation. Fixed at the root in `aa25624c`.

`validateDateFlag` is now `normalizeDateFlag`: it parses a date or RFC3339 timestamp and returns it truncated to `YYYY-MM-DD`, the granularity the stored `publish_date` uses. Both sides of the SQL comparison are now plain dates, so lexical order is chronological and no boundary-day episode is dropped. Verified live: `--since 2026-01-01` and `--since 2026-01-01T00:00:00Z` both return 624; `--until 2026-12-31` and `--until 2026-12-31T23:59:59Z` both return 2279. Invalid input (`--since zzzz`) still errors up front. Test updated to `TestNormalizeDateFlag` (asserts the timestamp→date normalization and the rejection).

maxswinguy:
You've now surfaced both horns of the same dilemma, which is the useful part — a UUID-less snip has **no stable server identity**, so any synthetic key trades one failure for the other:

- key on **content** (my previous commit) → distinct snips never collide, but editing a note/quote re-keys the row and a re-pull duplicates it (this finding);
- key on **metadata only** → edit-stable, but two snips sharing that metadata collide (the earlier "Fallback Snip IDs Collide" finding).

Fixed at the root in `1e9f8f84` by keying on the snip's **position — `episode + start + end`** (the clip bounds):

- **Edit-stable:** editing content never moves the clip, so a re-pull of an edited snip updates the same row instead of duplicating it. Resolves this finding.
- **Distinguishes real snips:** two snips at different clip spans get different ids. Covers the realistic collision case from the earlier finding (distinct snips are distinct clips).
- **Residual (documented, inherent):** two snips with a byte-identical `start` AND `end` in one episode collide — but that is the same captured clip, i.e. a genuine near-duplicate. There is no key that avoids both duplication-on-edit and collision-on-identical-position for a record with no stable id; position optimizes for the common operation (editing) over the pathological one (snipping the exact same clip span twice). And this path only touches the handful of UUID-less snips — the vast majority carry a stable deep-link UUID and never reach it.

Test is now `TestFallbackSnipIDIsPositionalAndEditStable` (edited content keeps the id; different clip spans differ). Full suite + `go vet` + `publish validate` (11/11) + live `dogfood --level quick` (16/16) green; a real `pull` returns the identical corpus. Flagging the residual explicitly so it's a maintainer call rather than a silent choice.

tmchow:

maxswinguy:
Thanks — this is the last horn of the UUID-less-snip dilemma, and you're right that episode + start + end still collapses when a snip exports with BOTH timestamps empty. Fixed at the root in `2d923eae8`.

A UUID-less snip has no server identity, so its key can't use content (that re-duplicates on edit — the earlier "Edited Snips Duplicate" finding) and can't use start+end alone (this finding). The fix keeps start+end as the primary key and adds a per-episode ordinal ONLY for the both-empty case:

- Snips that carry any timestamp are untouched — same key as before, so no re-keying churn and the edit-stable / position-stable properties hold.
- When both `start` and `end` are empty, the snip gets its position among the both-empty snips in that episode (export order) appended to the key, so two distinct such snips get distinct ids instead of one overwriting the other.
- The ordinal comes from export order, not content, so editing a note/quote/transcript doesn't change it — a re-pull still updates the row rather than duplicating it.
- The residual is now a harmless duplicate row (only if a both-empty snip is inserted or removed and shifts positions), never silent data loss. For a record with no stable id and no timestamps, that's the safe direction to fail.

Regression test `TestFallbackSnipIDBothEmptyTimesDistinguishedByOrdinal`: two both-empty snips in one episode get distinct ids; the same snip at a fixed position stays stable across a content edit.

Full suite + `go vet` + `publish validate` (11/11) + live `dogfood --level quick` (16/16) green. A real `pull` is behavior-preserving — identical corpus on the probe (3 episodes / 52 snips), because real snips carry a UUID (or at least a start time), so this path only guards the pathological both-empty case.

Issue fixed and comment written by Claude, reviewed by me

tmchow:

tmchow:

- missing publish-process files: library/media-and-entertainment/snipd/dogfood-results.json

maxswinguy:
Thanks — that one is on me: `dogfood-results.json` never made it into the PR. It existed locally but `publish package` doesn't copy it into the publish clone, so it silently stayed behind on my machine while every other artifact came across.

Fixed in `9f7c450c5`. Rather than commit the stale local copy (it predated the last four commits on this branch), I regenerated it against the current head from the publish tree itself:

- **verdict: WARN**, one advisory issue — `defaultSyncResources empty`. That is structural for this CLI, not a regression: snipd populates its corpus through the hand-written `pull` command (Obsidian ZIP export → parse → upsert), so the generated `sync` pipeline has no advertised resources. Happy to annotate it differently if you'd prefer the gate to read clean.
- **example_check: 10/10** commands have examples — this ran clean here, where the earlier local run had skipped it (the source master isn't a git checkout, so `go build` failed VCS stamping; running from the publish clone builds fine).
- No local paths in the artifact.

Branch head is now `9f7c450c5`; nothing else changed.

— Same disclosure as before: I'm a designer, not a developer; Claude Code did the work under my direction and I reviewed it.

maxswinguy:
Checking in on this one — no urgency, and I know the queue is long.

Where it stands: the missing `dogfood-results.json` is tracked as of `9f7c450c5`, and all seven checks are green on that head, Greptile included. The last triage comment predates that push — its marker still points at `2d923eae83f1` — so I suspect the re-triage just hasn't re-fired rather than there being something outstanding. Flagging in case that's useful.

Is there anything else you'd need from me to help move this forward? Happy to add evidence, adjust the artifact, or split anything out.

The reason I'm asking rather than just waiting: v1.1 is spec'd — a guided `auth login` that automates Snipd's browser pairing flow and replaces the manual token-copy step — and I'm deliberately holding it until this merges, so it lands as its own reviewable PR off a synced fork instead of growing this one. Happy to wait either way; I just don't want to sit on it if something is actually needed from my side.

tmchow:
The Aug 10 note asked for `dogfood-results.json`. That file is on this head, and the actual live proof was already in the manuscripts: `phase5-acceptance.json` is pass / quick / 16 of 16 against the real Snipd export API. Greptile's leftover 4/5 is the documented UUID-less no-time ordinal residual, not a new defect, and the policy gate passed. Routing to maintainer.

PR #1572, opened 2026-07-22
fix(janeapp): reject anonymous sessions on browser cookie import

omarshahine:
## The bug

`janeapp-pp-cli auth login --chrome` (and `--cookies-file`) checks only that a cookie **named** `_front_desk_session` exists, then saves it and prints `Session saved for clinic "..."`.

Jane sets that cookie for **logged-out visitors too**, and Chrome's on-disk cookie DB lags the live in-memory session. So importing while signed out — or while Chrome simply hasn't flushed a fresh login yet — stores an anonymous session and reports success.

Every command afterwards fails with a bare 401 whose hint is *"run `auth login --chrome`"* — the exact command that just reported success. That loop is genuinely hard to diagnose: a fresh `cookies.json` mtime and an `OK` message both say auth is fine, so the 401 reads like an account or permissions problem rather than a failed capture. I lost a while to it before finding the cause.

## The fix

The password path already handles this correctly — `janeLogin` probes `/api/v2/appointments` and fails on 401/403. This lifts that check into `janeVerifySession` and calls it from the browser-import path **before** saving.

`/api/v2/appointments` is the right probe: it returns 401 for absent, invalid, and valid-but-anonymous sessions alike, and 200 only when signed in. The HTML `/account` page is **not** usable — it returns 200 for an invalid cookie, silently redirecting to `/cookies_required`, so a status check against it passes when it shouldn't. (I made that mistake first; the comment records it so the next person doesn't.)

Server errors and transport failures are kept distinct from "not authenticated", so a 500 or a network blip never sends the user off to re-authenticate a session that's actually fine.

## Before / after

Against a live Jane instance with a stale cookie for a signed-out clinic:

```console
# before — silently stores an anonymous session
$ janeapp-pp-cli auth login --clinic leahkangas --chrome
OK Found 1 cookie(s) for leahkangas.janeapp.com
Session saved for clinic "leahkangas."
$ echo $?
0

# after
$ janeapp-pp-cli auth login --clinic leahkangas --chrome
ERROR Imported a cookie for leahkangas.janeapp.com, but it is not a signed-in session.

Jane sets this cookie for logged-out visitors too, and Chrome's
saved-to-disk copy can lag the session in the live browser window.

Confirm you are signed in here (you should see your name):

  https://leahkangas.janeapp.com/account

Then re-run this command. If it still fails, the on-disk cookie is
stale — export from the live session and import the file instead:

  janeapp-pp-cli auth login --clinic leahkangas --cookies-file <storage-state.json>
Error: imported session for leahkangas.janeapp.com is not authenticated
$ echo $?
4
```

A signed-in clinic still imports cleanly and exits 0 (verified separately).

## Tests

Six `httptest`-backed cases in `jane_auth_verify_test.go`: authenticated accepted, anonymous rejected, empty/whitespace cookie strings rejected, multi-cookie strings parsed, and 500 / unreachable-host both kept distinct from "anonymous".

- `go build ./...` ✅
- `go vet ./...` ✅
- `go test ./...` ✅ (full package suite, not just the new tests)
- `gofmt -l` clean

No changes to `registry.json` or `cli-skills/` per CONTRIBUTING.

## Note

This is a fix to an existing library CLI rather than a new submission, so it came in via a manual fork/PR rather than `/printing-press publish`. Happy to reshape it if you'd prefer this land differently.

### DIFF
--- a/library/health/janeapp/.printing-press-patches/verify-authenticated-session-on-browser-import.json
+++ b/library/health/janeapp/.printing-press-patches/verify-authenticated-session-on-browser-import.json
@@ -1,0 +1,15 @@
+{
+  "schema_version": 2,
+  "id": "verify-authenticated-session-on-browser-import",
+  "applied_at": "2026-07-22",
+  "base_run_id": "20260707-170547",
+  "base_printing_press_version": "4.27.1",
+  "summary": "The browser cookie-import login path (auth login --chrome / --cookies-file) must verify the imported session is actually authenticated before saving it, not merely that a cookie named _front_desk_session is present.",
+  "reason": "Jane sets _front_desk_session for logged-out visitors too, and Chrome's on-disk cookie DB lags the live in-memory session, so importing while signed out (or before Chrome flushes a fresh login) stored an anonymous session and printed 'Session saved'. Every later command then 401s with a hint to run the command that just 'succeeded'. A regen must keep janeVerifySession probing /api/v2/appointments (401 for absent/invalid/anonymous, 200 only when signed in) before setClinicSession, and must NOT switch the probe to the HTML /account page, which returns 200 for an invalid cookie by redirecting to /cookies_required. Server errors and transport failures are kept distinct from 'not authenticated' so a 500 or a network blip never tells the user to re-authenticate a working session. The verification honors the global --timeout, consistent with the two-step password path.",
+  "files": [
+    "internal/cli/jane_auth.go",
+    "internal/cli/auth.go",
+    "internal/cli/jane_auth_verify_test.go"
+  ],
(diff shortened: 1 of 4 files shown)


omarshahine:
Ran an independent `codex review` on the change and addressed both findings in `06d59ab`:

- **[P2] Honor the configured `--timeout`.** The verification request hardcoded `30s`, ignoring the global `--timeout` (default `60s`) — a slow endpoint could fail early, and `--timeout 1s` could still block up to 30s. Now passes `flags.timeout`, matching the two-step password path (`janeLogin`) and the rest of the CLI.
- **[P2] Record the customization in the patch catalog.** Added `.printing-press-patches/verify-authenticated-session-on-browser-import.json` so a future regeneration preserves the authenticated-session probe instead of silently reverting to the presence-only check. (This is also why the probe must stay on `/api/v2/appointments` and not move to `/account`, which 200s for an invalid cookie — recorded in the `reason` field.)

Re-verified after the fix: `--timeout 5s` on a dead session exits 4 in <0.5s; a signed-in clinic still imports and exits 0. `go build`, `go vet`, full package `go test`, and `gofmt` all pass.

Note on the red **Greptile policy gate**: it timed out *waiting for Greptile to start* (no review check ever appeared) and auto-nudged `@greptileai review`; it isn't a finding against the diff. Greptile doesn't seem to auto-review this fork PR, so that gate likely needs a maintainer-side run or waiver.

tmchow:

omarshahine:
@greptileai review

PR #1591, opened 2026-07-24
feat(facebook-marketplace): add direct buyer thread reader

mgalpert:
## Summary
- add a new `inbox thread` command that reads a known Marketplace Messenger thread route directly by thread id
- parse thread summary, contact rows, and message rows from the direct `/messages/t/<thread>/` shell instead of relying on the generic inbox feed
- document the new command and cover the parser with targeted client tests

## Validation
- env GOCACHE=/private/tmp/go-build-cache /Users/pal/.homebrew/bin/go test ./internal/client ./internal/cli
- env GOCACHE=/private/tmp/go-build-cache /Users/pal/.homebrew/bin/go run ./cmd/facebook-marketplace-pp-cli inbox thread --thread 27799138726406206 --agent


tmchow:
@Greptileai review

PR #1390, opened 2026-06-27
feat(vibe-signal): add vibe-signal

Avanderheyde:
## vibe-signal

One question, every low-friction source: a recency-aware, cited signal report instead of ten raw source dumps.

**API:** vibe-signal | **Category:** media-and-entertainment | **Press version:** 4.27.0
**Spec:** Internal aggregator spec (no upstream OpenAPI) — Hacker News (Algolia + Firebase) and Techmeme RSS

### Publication Path

New print — no merged CLI and no existing PR matched this slug.

### CLI Shape

```bash
$ vibe-signal-pp-cli --help
Vibe Signal CLI — One question, every low-friction source: a recency-aware, cited signal report instead of ten raw source dumps.

Highlights (not in the official API docs):
  • report   Ask one question across Hacker News and Techmeme and get a recency-aware signal report with per-source coverage.
  • evidence   List the raw, cited items (post URL, author, timestamp, points, comments) backing a topic from a chosen source.
  • sources list   Show which sources are wired in, their auth needs, and which command syncs them.

Agent mode: add --agent to any command for JSON output + non-interactive mode.
Health check: run 'vibe-signal-pp-cli doctor' to verify auth and connectivity.
See README.md or the bundled SKILL.md for recipes.

Usage:
  vibe-signal-pp-cli [command]

Available Commands:
  agent-context Emit structured JSON describing this CLI for agents
  completion    Generate the autocompletion script for the specified shell
  doctor        Check CLI health
  evidence      List the raw, cited items backing a topic from the local store
  feedback      Record feedback about this CLI (local by default; upstream opt-in)
  help          Help about any command
  hn            Hacker News source: search (Algolia) and item lookup (Firebase)
  profile       Named sets of flags saved for reuse
  report        Cross-source signal report for a topic, with per-source coverage
  search        Full-text search across synced data or live API
  sources       Inspect and sync the wired sources (list, sync)
  sync          Sync API data to local SQLite for offline search and analysis
  version       Print version
  which         Find the command that implements a capability
  workflow      Compound workflows that combine multiple API operations

Flags:
      --agent                Set all agent-friendly defaults (--json --compact --no-input --no-color --yes)
      --compact              Return only key fields (id, name, status, timestamps) for minimal token usage
      --config string        Config file path
      --csv                  Output as CSV (table and array responses)
      --data-source string   Data source for read commands: auto (live with local fallback), live (API only), local (synced data only) (default "auto")
[...]

PR #791, opened 2026-05-23
feat(prediction-goat): search relevance, freshness, venue scoping, LLM-driven learning

mvanhorn:
Stacked on top of #780. **Do not merge until #780 lands.**

This PR ships eight implementation units that address the search-relevance, sync-coverage, freshness-safety, venue-scoping, and learning gaps in prediction-goat-pp-cli — all surfaced in a debugging session this morning where the CLI silently returned wrong or stale answers on representative queries.

Plan: [`docs/plans/2026-05-23-001-feat-prediction-goat-search-relevance-and-freshness-plan.md`](docs/plans/2026-05-23-001-feat-prediction-goat-search-relevance-and-freshness-plan.md).

## What this fixes

| Query | Pre-fix top Kalshi hit | After fix |
|---|---|---|
| `portugal world cup` | (nothing — not synced + ranking) | KXMENWORLDCUP-26-PT at position 0 |
| `oscars best picture 2026` | KXFUSION (Nuclear fusion!) | KXOSCARSCORE / KXOSCARCOUNT* at position 0 |
| `bitcoin 100k` | BTCETHATH-29DEC31 (Ethereum vs BTC) | KXBTCMAX100 at position 0 |
| `fed rate cut june` | "government spending cut" event | KXRATECUT / KXFEDRATEMIN |
| `nba championship 2026` | "NBA team" series | KXNBA finals series |

And: cached prices no longer mislead. Concurrent reads no longer fail with SQLITE_BUSY. Agents can scope to one venue. The CLI now learns from the LLM's response so future identical questions short-circuit discovery.

## Commits (8 units, 374 tests passing)

| Unit | Commit | What changed |
|---|---|---|
| U1 | `225e50eb` | WAL pragmas via `sqlite.RegisterConnectionHook` — DSN-level `_journal_mode=WAL` was silently ignored by modernc.org/sqlite v1.37.0, leaving the live store on journal_mode=delete with no busy_timeout |
| U2 | `0add272d` | Curated `resources_fts.content` (title + ticker + category + tags, dropping raw-JSON noise) + schema v2→v3 reindex migration |
| U3 | _(skipped)_ | Title-boost ranking — U2 made BM25 over the curated content sufficient; verified via [`fts_speedrun_test.go`](library/payments/prediction-goat/internal/store/fts_speedrun_test.go) |
| U4 | `1c7f22b1` | `--venue`, `--polymarket`, `--kalshi` flags on topic, compare, liquid, movers, resolving, trending, new |
[...]

PR #1594, opened 2026-07-25
fix(slack): assemble POST bodies from flags; unwrap live search envelope; surface ok:false

dericknwq:
## Summary

Three bugs in `slack-pp-cli` that each produce a wrong answer silently rather than an error. No
write operation in the CLI could succeed — every `POST` shipped an empty `{}` body because the
flag values were never read. Live `search` reported "No results" with exit 0 for queries with
thousands of matches, because Slack's result envelope was unwrapped incorrectly and then
discarded by an emptiness filter. And Slack's `{"ok":false}` application errors, which arrive
with HTTP 200, were reported as success on every write path, so a rejected write looked like a
completed one.

All three are fixed CLI-locally, with the generator's own current output as the reference for
the write-body fix. Local-mode `search` is **not** fixed — see Scope below.

## Findings

|---|---|---|---|
| F1 | request-body assembly | bug | Write fields are declared under `params:` in the spec, and the generated body builder iterated only `Endpoint.Body`, so all 33 POST commands sent `{}`. Required-flag validation also ran unconditionally, so `--stdin` alone failed validation and the only working call duplicated every value in both places. |
| F2 | search response envelope | bug | `search.messages` nests hits at `messages.matches`, which no generic wrapper key reaches; the envelope then failed `isNilOrEmpty` and was dropped, yielding "No results" and exit 0. |
| F4 | API error surfacing | bug | Slack signals application errors as HTTP 200 + `{"ok":false,"error":...}`. `internal/client/client.go` has no `ok` handling on any verb, and the write and live-search paths never called the CLI-layer checker, so failures rendered as success. |

## Changes

41 files, +1089 / -162.

- **F1** — 30 command files now assemble the request body from their flag values. `unfurl_links`
  and `is_private` are booleans and `post_at` and `num_minutes` are integers in the Slack API, so
  those four are coerced from their string flags and reject non-parsing input with a named error
  rather than sending a JSON string the API refuses. Required-flag validation moved inside the
  non-stdin branch, making flags and `--stdin` independent routes; stdin wins when both are given.
  Three commands (`auth.test`, `dnd.endDnd`, `dnd.endSnooze`) take no body parameters, so their
  empty map was already correct and they are unchanged apart from F4.
- **F2** — `extractSearchResults` understands the `messages.matches` nesting, and `isNilOrEmpty`
  recognizes a Slack match by its `text`/`ts`/`permalink` fields. Both were required: unwrapping
  alone would still have had every hit discarded by the filter.
- **F4** — the 33 write commands, `import`, and the live-search path now call the existing
  `checkSlackAPIError` helper. This reuses the detector already used by `resolveRead` and the DM
  helpers rather than adding a second one. The prior `fix-ok-false-detection` patch, which covered
  `internal/cli/sync.go` only, is untouched.
- **`internal/cli/import.go` — the one change outside the F1/F2/F4 framing, called out so it is not
  missed in review.** It discarded the API response entirely (`_, _, err := c.Post(...)`), so a
  record Slack rejected with HTTP 200 + `{"ok":false,...}` was still counted as an imported success.
  It is the same silent-failure class as F4 and now performs the same check, reporting the Slack
  error and incrementing `failed` instead of `success`. Included deliberately: fixing silent write
  failures everywhere except one file would have left the PR incoherent.
- **Docs** — cookbook recipes for the write path and `--stdin`, a "Writing to Slack" section in
  both `README.md` and `SKILL.md`, and a **Known Limitations** section stating plainly what is
  still broken.
- **Tests** — `internal/cli/search_amend_test.go`, six cases covering envelope unwrapping, filter
  survival, the two fixes composed, a genuinely empty result still reading as empty, and `ok:false`
  detection including that `missing_scope` names the needed scope.

## Scope — what this PR does not fix

**Local-mode `search` remains broken.** `search --data-source local` still returns nothing for
messages. `sync` populates the `messages` and `messages_fts` tables, but `Store.SearchMessages`
(`internal/store/store.go:711`) has **zero callers** — `search.go`'s empty-type branch queries
only `SearchUsergroups` and `SearchFiles`, and `--type messages` falls through to the generic
`db.Search`. So the local index can hold matches the CLI cannot reach. This is a deliberate
omission, not an oversight: it needs a decision about whether the typed-table architecture is
kept, which is out of scope here. Documented under Known Limitations so nobody reads "search
fixed" as covering both modes.

Also known-remaining, now documented and filed as follow-ups: `files upload` targets Slack's
retired `files.upload` and fails `method_deprecated` even with a correct body; `auth test` is not
registered on the command tree; `sync --resources channels` fails `unknown_method`.

## Why not just reprint?

The generator bug behind F1 was fixed upstream (cli-printing-press#790, PR #791, press ~4.2.0)
and this CLI still carries it only because it has not been reprinted since press `1.2.1`. A
reprint was evaluated and rejected on evidence. Generating this CLI's spec under press `4.29.0`
shows:

- The prior spec **no longer generates** — a new reserved-name guard rejects the `auth` and
  `search` resources, so a reprint requires spec edits that rename user-visible commands.
- **All 8 novel features are absent** from generator output (`digest`, `health`,
  `response_times`, `trends`, `quiet`, `user_activity`, `funny`, `threads_stale`) — most of this
  CLI's differentiated value.
- The regenerated `sync.go` has **no messages sync at all** (no `conversations.history`
  reference, no `messages` resource), and the regenerated store drops all 14 typed tables for a
  generic one — so a reprint would leave nothing for local message search to find.
- `regen-merge` classifies **23 files as published-only and 8 as needing hand merge**, including
  `sync.go`, `store.go`, `client.go`, and `conversations_history.go` — precisely the files
  carrying the six existing patches.
- Three of the six recorded patches (`fix-conversations-types-scope`, `fix-ok-false-detection`,
  `fix-im-history-channel-not-found`) are **dropped** by regeneration; the other three are
  absorbed.
- **F3 would still not be fixed** — the regenerated empty-type branch calls only the generic
  `db.Search` while its comment claims it searches typed tables too.

So the reprint's one confirmed win is F1, at the cost of losing the novel features, re-applying
three patches, hand-merging eight files, breaking command names, and forcing a full local re-sync.
Backporting F1 from the generator's own output is the cheaper, safer path. A reprint remains
worth doing as its own scoped piece of work, once someone can steward those eight merges and
decide the store architecture.

## Verification

- `go build ./...` — PASS
- `go vet ./...` — PASS
- `gofmt` — clean
- `go test ./internal/cli/` — PASS (6 new cases)
- `govulncheck` — PASS
- `cli-printing-press publish validate` — **byte-identical to the pre-change baseline.** The
  suite reports 4 pre-existing failures on this CLI (`manifest` wants `schema_version: 2`,
  `transcendence` finds no recorded novel features, `phase5` is missing its acceptance artifact,
  and `verify-skill` reports 3 README flag-attribution errors). All 4 fail identically on a
  pristine `upstream/main` checkout of this directory, so **this PR introduces no new
  validation failures**. They are pre-existing publish-readiness gaps and are listed in the
  follow-up issue rather than fixed here.
- `verify_skill.py --dir library/productivity/slack` — 3 errors, unchanged from baseline. One
  additional item appears in the "likely false positive" group, from a cookbook line; the tool
  mis-attributes flags of underscore-named subcommands (`messages post_message --text`) to the
  parent command, which is why the 3 pre-existing errors exist too.
- Empirical check of F1, dry-run with a dummy token (no network I/O — `client.do` returns at its
  `DryRun` branch before a request is built):

  ```
  $ slack-pp-cli messages post_message --channel C00000000 --text "hello world" \
      --thread-ts 1700000000.000000 --dry-run
  POST https://slack.com/api/chat.postMessage
    Body:
  {
      "channel": "C00000000",
      "text": "hello world",
      "thread_ts": "1700000000.000000"
    }
  ```

  Previously this printed `{}`. `--stdin` alone now also builds a correct body without tripping
  required-flag validation, and `--is-private true` serializes as a JSON boolean rather than the
  string `"true"`.

## Evidence

Patch records at this PR's HEAD:

- [`slack-post-body-from-flags.json`](https://github.com/dericknwq/printing-press-library/blob/a92f7435a9136ce03f0621bbca7b0a137d913787/library/productivity/slack/.printing-press-patches/slack-post-body-from-flags.json)
- [`slack-search-messages-matches.json`](https://github.com/dericknwq/printing-press-library/blob/a92f7435a9136ce03f0621bbca7b0a137d913787/library/productivity/slack/.printing-press-patches/slack-search-messages-matches.json)
- [`slack-ok-false-on-writes-and-search.json`](https://github.com/dericknwq/printing-press-library/blob/a92f7435a9136ce03f0621bbca7b0a137d913787/library/productivity/slack/.printing-press-patches/slack-ok-false-on-writes-and-search.json)
- [`internal/cli/search_amend_test.go`](https://github.com/dericknwq/printing-press-library/blob/a92f7435a9136ce03f0621bbca7b0a137d913787/library/productivity/slack/internal/cli/search_amend_test.go)
Each patch records a `deferred_to_upstream` entry naming the machine-level condition that should
supersede it, so none of these three is left looking like a permanent local fork:

- `slack-post-body-from-flags` → **cli-printing-press#790** (closed). Already fixed in the
  generator; this patch is a backport for a CLI printed under 1.2.1 and is superseded the moment
  `slack` is reprinted.
- `slack-search-messages-matches` → **cli-printing-press#3795** (opened with this PR). The
  generated search template has no way to express a per-API response path or result identity
  fields, so any API whose search envelope falls outside the hardcoded generic lists silently
  returns zero results. A spec-declared `response_path` + `identity_fields` would supersede this
  patch. That issue also flags that #1390's code/comment-alignment acceptance criterion is not
  currently met in 4.29.0 output.
- `slack-ok-false-on-writes-and-search` → **cli-printing-press#3740** (open, additive
  error-classifier seam). One client-level hook would replace the 35 call-site checks this patch
  adds.

### DIFF
--- a/.github/scripts/verify-skill/verify_skill.py
+++ b/.github/scripts/verify-skill/verify_skill.py
@@ -1170,6 +1170,27 @@ def check_flag_commands(cli_dir: Path, sources: list[Path], cli_binary: str, rep
         for raw_cmd_path, _positional, flags, _surface in extract_cli_invocations(src, cli_binary, cli_dir):
             cmd_path = list(raw_cmd_path)
             path_str = " ".join(cmd_path)
+            # The extractor under-resolves subcommands whose names contain an
+            # underscore — `messages post_message --text` yields cmd_path
+            # ["messages"] with "post_message" left among the positionals — so
+            # flags belonging to the subcommand were attributed to its parent and
+            # reported as "declared elsewhere but not on <parent>".
+            #
+            # check_positional_args already treats this shape as a false positive
+            # (see the snake_case note there), but that only downgraded its own
+            # finding; this check still emitted hard errors.
+            #
+            # Resolve it instead of suppressing it, but only as a fallback once the
+            # parent has been ruled out. Trying the child first would misread
+            # commands that take a resource name positionally — `export messages
+            # --format` and `tail messages --interval` both resolve
+            # ["export", "messages"] to a messages source file, which does not
+            # declare --format or --interval, turning working docs into errors.
+            fallback_path: list[str] | None = None
+            if len(cmd_path) == 1 and _positional and re.match(r"^[a-z][a-z0-9_-]+$", _positional[0]):
+                candidate = cmd_path + [_positional[0]]
+                if find_command_source(cli_dir, candidate)[0]:
+                    fallback_path = candidate
             for raw_flag in flags:
                 flag = raw_flag.lstrip("-")
                 key = (path_str, flag)
@@ -1182,6 +1203,15 @@ def check_flag_commands(cli_dir: Path, sources: list[Path], cli_binary: str, rep
                     continue
                 if cmd_files and flag_declared_via_helper(cli_dir, cmd_files, flag):
                     continue
+                # Parent ruled out — the flag may belong to an under-resolved
+                # subcommand left in the positionals (see fallback_path above).
+                if fallback_path:
+                    sub_files, _, _ = find_command_source(cli_dir, fallback_path)
+                    if sub_files and (
+                        flag_declared_in(sub_files, flag)
+                        or flag_declared_via_helper(cli_dir, sub_files, flag)
+                    ):
+                        continue
                 seen.add(key)
                 if flag_declared_in(all_files, flag):
                     report.findings.append(
--- a/library/productivity/slack/.printing-press-patches/slack-ok-false-on-writes-and-search.json
+++ b/library/productivity/slack/.printing-press-patches/slack-ok-false-on-writes-and-search.json
@@ -1,0 +1,57 @@
+{
+  "schema_version": 2,
+  "applied_at": "2026-07-25",
+  "base_run_id": "20260409-073625",
+  "base_printing_press_version": "1.2.1",
+  "id": "slack-ok-false-on-writes-and-search",
+  "summary": "Surface Slack ok:false application errors on writes, imports, and live search.",
+  "reason": "Slack answers application errors with HTTP 200 and {\"ok\":false,\"error\":\"...\"}, and internal/client/client.go has no ok handling on any verb, so the client reports these as success. The CLI layer compensates via checkSlackAPIError, but only on paths that route through resolveRead or the DM helpers. Every POST command called c.Post directly and never checked, so the JSON envelope reported success:true and exit 0 for a write Slack had rejected \u2014 the failure mode in cli-printing-press#790's own reproduction. search's live path likewise called c.Get directly, bypassing resolveRead, which is why a failed search coerced to an empty result set rather than an error. import discarded the response entirely and counted rejected records as imported. All now reuse the existing checkSlackAPIError helper rather than adding a second detector; the prior fix-ok-false-detection patch covered internal/cli/sync.go only and is unchanged by this patch.",
+  "files": [
+    "internal/cli/auth_test.go",
+    "internal/cli/conversations_archive.go",
+    "internal/cli/conversations_create.go",
+    "internal/cli/conversations_invite.go",
+    "internal/cli/conversations_mark.go",
+    "internal/cli/conversations_set_purpose.go",
+    "internal/cli/conversations_set_topic.go",
+    "internal/cli/conversations_unarchive.go",
+    "internal/cli/dnd_end_dnd.go",
+    "internal/cli/dnd_end_snooze.go",
+    "internal/cli/dnd_set_snooze.go",
+    "internal/cli/dnd_team_info.go",
+    "internal/cli/files_delete.go",
+    "internal/cli/files_upload.go",
+    "internal/cli/messages_delete_message.go",
+    "internal/cli/messages_list_scheduled.go",
+    "internal/cli/messages_post_message.go",
+    "internal/cli/messages_schedule_message.go",
+    "internal/cli/messages_update_message.go",
+    "internal/cli/pins_add.go",
+    "internal/cli/pins_remove.go",
+    "internal/cli/reactions_add.go",
+    "internal/cli/reactions_remove.go",
+    "internal/cli/reminders_add.go",
+    "internal/cli/reminders_complete.go",
+    "internal/cli/reminders_delete.go",
+    "internal/cli/stars_add.go",
+    "internal/cli/stars_remove.go",
+    "internal/cli/usergroups_create.go",
+    "internal/cli/usergroups_update.go",
+    "internal/cli/usergroups_users_update.go",
+    "internal/cli/users_profile_set.go",
+    "internal/cli/users_set_presence.go",
+    "internal/cli/import.go",
+    "internal/cli/search.go"
+  ],
+  "validated_outcome": "Table-driven tests cover ok:false plain errors, missing_scope (asserting the needed scope is named in the message), ok:true success, and non-Slack payloads. go vet and go build clean. --dry-run is unaffected because the dry-run payload carries no ok field.",
+  "findings_addressed": [
+    "F4"
+  ],
+  "deferred_to_upstream": [
+    {
+      "feature": "An additive error-classifier seam so 200-with-error-envelope APIs are handled in the client",
+      "reason": "Slack is one of many APIs that signal failure inside a 200 body. Without a generator-level seam, each printed CLI must hand-patch detection at every call site \u2014 this patch touches 35 files to do what one client-level hook would. Tracked upstream as the additive error-classifier seam request."
+    }
+  ],
+  "upstream_issue": "https://github.com/mvanhorn/cli-printing-press/issues/3740"
+}
--- a/library/productivity/slack/.printing-press-patches/slack-ok-false-on-writes-and-search.json
+++ b/library/productivity/slack/.printing-press-patches/slack-ok-false-on-writes-and-search.json
@@ -5,7 +5,7 @@
   "base_printing_press_version": "1.2.1",
   "id": "slack-ok-false-on-writes-and-search",
   "summary": "Surface Slack ok:false application errors on writes, imports, and live search.",
-  "reason": "Slack answers application errors with HTTP 200 and {\"ok\":false,\"error\":\"...\"}, and internal/client/client.go has no ok handling on any verb, so the client reports these as success. The CLI layer compensates via checkSlackAPIError, but only on paths that route through resolveRead or the DM helpers. Every POST command called c.Post directly and never checked, so the JSON envelope reported success:true and exit 0 for a write Slack had rejected \u2014 the failure mode in cli-printing-press#790's own reproduction. search's live path likewise called c.Get directly, bypassing resolveRead, which is why a failed search coerced to an empty result set rather than an error. import discarded the response entirely and counted rejected records as imported. All now reuse the existing checkSlackAPIError helper rather than adding a second detector; the prior fix-ok-false-detection patch covered internal/cli/sync.go only and is unchanged by this patch.",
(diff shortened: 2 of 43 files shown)


dericknwq:
Known-remaining items for this CLI, deliberately out of scope here, are tracked in #1595 — including local-mode `search` (`Store.SearchMessages` has zero callers), `files upload` hitting the retired `files.upload`, `auth test` not being registered because its file is named `auth_test.go`, and the pre-existing `publish validate` failures this PR leaves byte-identical to baseline.

The generator-level cause of the search fix is mvanhorn/cli-printing-press#3795, referenced from `slack-search-messages-matches.json` as `deferred_to_upstream`.

dericknwq:
### On the failing `Validate SKILL.md against shipped CLI source` check

This check **fails identically on `upstream/main`** — it is not introduced by this PR. Because
`verify-skills.yml` only runs the verifier against CLIs whose files changed, it has not run
against `slack` before, so the pre-existing failure surfaces here for the first time.

Running the repo's own verifier against both trees:

```
$ python3 .github/scripts/verify-skill/verify_skill.py --dir library/productivity/slack

upstream/main   ✘ 3 error(s), 3 likely false-positive(s)   exit 1
this PR         ✘ 3 error(s), 4 likely false-positive(s)   exit 1
```

The **3 hard errors are unchanged**. The only delta is one additional item in the
*likely-false-positive* group, from a new README cookbook line documenting the write path.

All of these items are the same verifier bug: it mis-parses subcommands whose names contain an
underscore, reading `slack-pp-cli messages post_message --text …` as command `messages` with a
stray positional argument, then reporting `"messages" expects 0–0`. The verifier labels each one
`[likely false positive]` itself:

```
[positional-args] slack-pp-cli messages: got 1 positional args; Use: "messages" expects 0–0  [likely false positive]
  evidence: README.md: post_message
```

Single-word subcommands parse correctly, which is why the `SKILL.md` examples in this PR use
`reactions add` / `pins add` and contribute nothing here. The remaining hits are all on
underscore-named commands (`post_message`, `lookup_by_email`).

I deliberately did not reword the README line to reach exact 3+3 parity. Documenting
`messages post_message` is the substance of this PR — it is the flagship command that could not
previously send a body — and weakening the docs to satisfy a known verifier mis-parse seemed the
wrong trade. Happy to change it if you would rather the delta be zero.

Two options that would actually clear this, both outside this PR's scope:

1. Teach the verifier to resolve multi-word subcommand paths before counting positional args.
2. Have it exclude `[likely false positive]` items from the exit code, so genuine errors still
   fail the build.

Filed as item 9 of #1595. Not fixed here since it is repo tooling rather than this CLI.

dericknwq:
### Added commit: `ok:false` errors now carry the documented exit code

Found while building on this branch. The `ok:false` detection in this PR surfaced the right message
but returned bare `fmt.Errorf` values, so `ExitCode` found no `*cliError` and fell through to its
generic `1`.

Read paths exited `5` anyway — but only incidentally, because they hand the error back to
`classifyAPIError`, whose `default` branch wraps with `apiErr`. Write paths return it directly:

| Path | Before | After |
|---|---|---|
| `conversations history` (read) | 5 | 5 |
| `messages post_message` (write) | **1** | 5 |
| `files upload` (write) | **1** | 5 |

So a Slack-rejected write was indistinguishable from a usage error for anything branching on exit
codes, against a README that documents `5` as the API-error code — which matters most for the agent
callers `--agent` mode exists to serve.

Fixed by wrapping both returns in `apiErr` inside `checkSlackAPIError`, which corrects all 36 call
sites at once rather than editing each. Read-path behaviour is unchanged: `classifyAPIError`'s
re-wrap resolves to the same code. Error text is byte-identical.

Verified against a nonexistent channel: all three paths now exit 5, and `search --data-source live`
plus a valid `conversations history` still exit 0. Folded into the existing
`slack-ok-false-on-writes-and-search` patch record rather than filed as a new patch, since it is the
same logical change.

Also opened #1596, stacked on this branch — it migrates `files upload` off Slack's retired
`files.upload`. It depends on this PR and should merge after it.

dericknwq:
### On the `Greptile policy gate` check

Flagging this so the two red checks on this stack aren't mistaken for problems with the changes.

`Greptile policy gate` runs for ~25 minutes, posts an `@greptileai review` auto-nudge when no
Greptile Review check appears, and then fails. That has happened twice here. It doesn't look
specific to this PR, or to fork PRs:

| PR | Author | Gate | Greptile reviews |
|---|---|---|---|
| #1598 | @mvanhorn (merged) | fail | 0 |
| #1604 | @MBemera | fail | 0 |
| #1603 | @giuseppebisemi | fail | 0 |
| #1594 / #1596 / #1605 / #1606 | this stack | fail | 0 |

No PR I checked has a Greptile review, including maintainer-authored #1598 — which merged with the
gate red, so it evidently isn't merge-blocking. Looks like the integration isn't responding
repo-wide rather than anything about these branches. Worth a look if the gate is meant to be
meaningful, since it currently fails for everyone.

I'd have happily addressed review findings; there simply aren't any to address. If you can trigger
Greptile another way, I'll work through whatever it raises.

For completeness, the other red check — `Validate SKILL.md against shipped CLI source` — is
pre-existing and fails identically on `upstream/main`, with the evidence in a comment above.

tmchow:

dericknwq:
### Correction: my earlier comment about the Greptile gate was wrong

Retracting [my comment above](https://github.com/mvanhorn/printing-press-library/pull/1594#issuecomment-5082631853) claiming the Greptile integration wasn't responding repo-wide.

It responded. `Greptile Review` has since passed on all four PRs in this stack and produced substantive, correct findings — including four real bugs of mine. What I actually observed was the `Greptile policy gate` timing out at ~25 minutes while Greptile replied later than that window, and I over-generalised from a snapshot taken during it. The comparison table in that comment was a point-in-time artifact, not evidence of a broken integration. Apologies for the noise.

The findings are legitimate and I'm working through them now:

- `import.go` — rejected records don't affect the exit code
- `canvases read --quiet` — still prints content
- `canvases read --dry-run` — parses the dry-run envelope as `files.info` and errors
- `SearchMessages` — no FTS5 sanitisation, unlike sibling indexes
- Stale `Known Limitations` entries contradicting fixes shipped later in the same stack
- Double-buffering of upload payloads
- Dry-run skipping `--canvas` validation

I'll also make `Validate SKILL.md against shipped CLI source` pass rather than arguing it's pre-existing. It is pre-existing — it fails identically on `upstream/main` — but that isn't the bar, and the three errors come from a verifier limitation worth fixing properly.

Will re-request review once all of it is green.

dericknwq:
All checks are green and the Greptile confidence score is **5/5**, so the two blockers from the `needs-author` triage above are resolved.

| Check | |
|---|---|
| Greptile Review | pass |
| Greptile policy gate | pass (5/5 ≥ 4/5) |
| Validate SKILL.md against shipped CLI source | pass |
| Verify | pass |
| Govulncheck | pass |

Fixed since triage:

- **`import` exited 0 with records rejected** — detecting per-record `ok:false` only mattered if the exit code followed. A partial or total import failure is now exit 5.
- **`Validate SKILL.md`** — this had been failing on `main` for every CLI, not just here. `check_flag_commands` attributed a subcommand's flags to its parent when the name contains an underscore, so `messages post_message --text` reported `--text` as "declared elsewhere but not on messages". Resolved by looking the flag up on the child, strictly as a fallback after the parent is ruled out — trying the child first misreads commands taking a resource name positionally, where `export messages --format` and `tail messages --interval` resolve to a messages source file declaring neither. Measured across 40 CLIs: slack 3→0 errors, `amazon-ads` 4→2 (its two were the same misattribution on `reports wait --wait-timeout`), the other 38 unchanged. No genuine finding is suppressed.

The verifier change is the one file outside this CLI's directory. Happy to split it into its own PR, though it fixes the check for every CLI as it stands.

The rest of the stack — #1596, #1605, #1606 — is also fully green now. Merge order is #1594 → #1596 → #1605 → #1606.

dericknwq:
Closing as superseded by #1605, which is rebased onto post-reprint `main`.

#1652 (the v4.29.0 reprint) landed underneath this stack. The generator now assembles POST bodies from
flags natively, which was the bulk of #1594 — so most of this PR is upstream already. The parts still
genuinely missing on `main` (`ok:false` on the command path, the live-search response path, and the
external upload flow) are consolidated into #1605 rather than left in a three-PR stack whose ordering
was itself the source of the CI confusion here.

Nothing is dropped: every surviving fix is in #1605 with its own patch record, and the search fix is
re-derived rather than carried over, because the reprint changed how that bug presents. Reopen if you'd
rather review it split.

PR #1596, opened 2026-07-25
feat(slack): migrate files upload to Slack's external upload flow

dericknwq:
> [!IMPORTANT]
> **Stacked on #1594.** This branch is based on `amend/slack-post-body-live-search-ok-false`,
> because `internal/cli/files_upload.go` carries that PR's write-body and `ok:false` patches.
> Review or merge #1594 first; the diff below is the one commit on top of it.

## Summary

`slack-pp-cli files upload` cannot succeed. It posts to `/files.upload`, which Slack has retired —
the endpoint answers `{"ok":false,"error":"method_deprecated"}` for every request regardless of
body, so the command was already dead before #1594's write-body fix and remains dead after it.

This replaces it with Slack's three-step external upload flow, and adds the `--file` flag the
command never had.

## The flow

| Step | Call | Note |
|---|---|---|
| 1 | `GET files.getUploadURLExternal?filename=&length=` | returns `upload_url` + `file_id` |
| 2 | `POST` raw bytes to `upload_url` | **unauthenticated, deliberately** |
| 3 | `POST files.completeUploadExternal` | `files=[{id,title}]`, optional `channel_id` / `initial_comment` / `thread_ts` |

Step 2 does not route through `client.do`, for two reasons: it targets a host other than
`slack.com`, and it must not carry the `Authorization` header. The returned URL carries its own
single-use token; attaching the workspace credential to a third-party host would leak it. It uses
the client's exported `HTTPClient` directly with a multipart body.

This is why the fix cannot live in the generator as-is — see *Upstream* below.

## Flags

- **`--file`** *(new)* — path to a local file. The command previously had **no way to upload a file
  from disk**; only inline `--content` was expressible, which is a strange gap for a command named
  `upload`.
- **`--thread-ts`** *(new)* — `completeUploadExternal` accepts it, so a file can be shared into a
  thread rather than only into a channel.
- `--channels` — retained as a deprecated alias for the retired endpoint's comma-separated
  parameter. It now **errors when given more than one channel** instead of silently dropping the
  extras, because `completeUploadExternal` shares into one conversation per call. Silently
  uploading to fewer channels than asked is the same class of quiet-wrong-answer this CLI has been
  cleaning up.
- `--stdin` was dropped: the request is no longer a single JSON body, so a JSON-body override has
  nothing to override.

Input is validated before any network call, so bad arguments fail without consuming a reserved
upload URL. An empty payload is rejected outright — Slack accepts the reservation and produces an
unusable zero-byte file.

## Verification

Confirmed end to end against a live workspace. The flow was first walked by hand to establish the
contract, then exercised through the command:

- 77-byte binary PDF via `--file` — uploaded, registered, correct name and byte count
- 18-byte inline payload via `--content` + `--filename` — same
- `--thread-ts` upload — landed in the target thread with the expected `thread_ts`
- Slack rejections surface via `checkSlackAPIError` and exit 5
- `--dry-run` prints all three planned steps, including that step 2 is unauthenticated, and
  contacts nothing

Each validation path produces a named error: no input; `--file` with `--content`; `--content`
without `--filename`; a multi-channel `--channels` list; a nonexistent `--file` path.

`go build`, `go vet`, `gofmt`, `go test ./internal/cli/` all clean. All test files and messages
created during verification were deleted afterwards.

## Upstream

Recorded in `.printing-press-patches/slack-external-file-upload.json` with two
`deferred_to_upstream` entries, neither of which has an issue yet:

1. **Multi-step upload flows aren't expressible in the generated command template.** The generator
   emits one request per command, so "reserve a URL, send bytes to a foreign host, then confirm"
   has no representation. It also has no way to mark a step as deliberately unauthenticated, which
   this flow needs for correctness rather than convenience. Any API with presigned-upload semantics
   — S3-style, Slack, Google Drive resumable — hits the same wall.

2. **The catalog generated a command for a retired endpoint.** No verification step caught that
   `files.upload` returns `method_deprecated` for every call, so a published CLI shipped a command
   that could never succeed. A catalog-level deprecation or liveness check would have caught it
   before publish.

Happy to open either as an issue if you'd like them tracked separately.

## Note for whoever tidies up test uploads

`files.delete` removes the file but leaves any `initial_comment` message behind as a standalone
message, so cleaning up takes both `files.delete` and `chat.delete`. A file posted with no
`initial_comment` does disappear along with its file.


tmchow:

dericknwq:
All checks green, Greptile **5/5**. Three review rounds on this one, each narrowing the same problem: step 1 reserves a length that step 3 registers, and neither can be renegotiated, so the body must match it exactly.

1. `os.Stat` sized the *path* and a deferred `os.Open` read it later — a file appended to, truncated, rotated or replaced in between gave a different snapshot than was reserved. Now opened once and sized through that handle.
2. One handle still wasn't enough: contents stay live across the step 1 round trip and the copy. The copy is bounded and the result verified.
3. Bounding alone **silently truncated a grown source** — Slack would store the prefix while the command reported success. A probe read after the bounded copy distinguishes "exactly right" from "grew", since `LimitReader` alone reports both as the reserved size.

Both directions now fail with a named error stating the reserved length. Nine tests cover the contract against a stub for both hosts, and they catch both regressions: the unbounded copy sends 45 extra bytes, and a bare cap without the probe truncates silently.

Also in scope, since fixing only the CLI left this half-done: the **MCP `files_upload` tool** was registered as `makeAPIHandler("POST", "/files.upload")` — the retired endpoint — so every agent following the documented MCP setup got `method_deprecated`. The three steps now live in `client.UploadExternal` and both surfaces call it. The MCP tool gained `file`/`channel`/`thread_ts` to match the command, rejects a multi-channel list instead of dropping the extras, and surfaces `ok:false`, which the generic handler never did.

Thanks for the persistence on the reservation contract — the third round was the one that mattered, since capping was itself a silent wrong answer.

dericknwq:
Closing as superseded by #1605, which is rebased onto post-reprint `main`.

#1652 (the v4.29.0 reprint) landed underneath this stack. The generator now assembles POST bodies from
flags natively, which was the bulk of #1594 — so most of this PR is upstream already. The parts still
genuinely missing on `main` (`ok:false` on the command path, the live-search response path, and the
external upload flow) are consolidated into #1605 rather than left in a three-PR stack whose ordering
was itself the source of the CI confusion here.

Nothing is dropped: every surviving fix is in #1605 with its own patch record, and the search fix is
re-derived rather than carried over, because the reprint changed how that bug presents. Reopen if you'd
rather review it split.

Issue #1556, opened 2026-07-21
granola: `sync` fails with `storage.dek not found` on Granola desktop 7.427.x (DEK moved into Keychain)


giuseppebisemi:
## Summary

`granola-pp-cli sync` (and every cache-backed command) fails to decrypt the local
Granola cache on a healthy, signed-in Granola desktop. The CLI aborts because it
requires an on-disk `storage.dek` file, but Granola desktop **7.427.6** no longer
writes one — the encryption key now lives only in the macOS Keychain
(Electron `safeStorage`). The CLI bails before it ever uses the Keychain key it
could access, so `sync` is impossible. Updating the CLI to the latest release did
not help.

## Environment

- `granola-pp-cli`: **2026.7.1** (also reproduced on 2026.6.2)
- Granola.app: **7.427.6**
- macOS: **26.5.2**, arm64 (Apple Silicon)
- Install method: Printing Press library (`--cli-only`)

## Error

`granola-pp-cli sync`:

```
Error: loading Granola cache at ~/Library/Application Support/Granola/cache-v6.json.enc:
reading encrypted cache ~/Library/Application Support/Granola/cache-v6.json.enc:
Keychain access unavailable (sign into Granola desktop or run `granola-pp-cli sync`
to authorize Keychain access): safestorage: Keychain key or DEK unavailable:
storage.dek not found at ~/Library/Application Support/Granola/storage.dek
(Granola not installed or pre-encryption version)
```

`granola-pp-cli doctor` (relevant lines):

```
OK   Config: ok
FAIL Auth: not configured
OK   API: reachable (HTTP 404 at /)
FAIL Encrypted store: ERROR last sync failed to decrypt (key_unavailable)
     version: 2026.7.1
INFO Cache: unknown  (sync_state is empty)
```

## Root-cause evidence

The `(Granola not installed or pre-encryption version)` hint is misleading — Granola
**is** installed, signed in, running, and actively writing the encrypted cache. The
real cause is an encryption-scheme change on the Granola side:

1. **The encrypted cache is present and fresh** — the desktop is healthy and writing it:
   ```
   cache-v6.json.enc   (~1 MB, recently written)
   supabase.json.enc   (~3 KB, recently written)
   ```

2. **No `storage.dek` (or any `*.dek`) file exists** anywhere under
   `~/Library/Application Support/Granola/`.

3. **The Keychain DOES contain the safeStorage key** — a generic-password item:
   ```
[...]

PR #1598, opened 2026-07-25
fix(granola): restore data access after Granola's DEK migration

mvanhorn:
Granola desktop moved its data-encryption key somewhere this CLI cannot follow, and every cache-backed command has been dead since. This restores access through the public API and makes the local path fail honestly where it can no longer work.

## Root cause

Granola 7.447.1 migrated the 32-byte DEK out of `storage.dek` and into the macOS **data-protection** keychain, under service `com.granola.app.dek` in access group `QZ7DHHLN25.granola`, then unlinked the file. The decompiled bundle shows the migrate-then-unlink flow directly:

```js
let {result:n} = await Q1e(readFileSync(t));   // decrypt storage.dek
r.length===32 && (await e.importDek(r), Z(`info`,`dek-migrated`,{}));
unlinkSync(t);                                 // delete it
```

`keychain.node` links `kSecUseDataProtectionKeychain` and `kSecAttrAccessGroup`, and `Granola.app` is signed `TeamIdentifier=QZ7DHHLN25` with a `keychain-access-groups` entitlement. That access group is gated by an entitlement only Granola's Apple team can issue, so no third-party binary can read the key. A legacy-API lookup of `com.granola.app.dek` returns item-not-found, confirming it lives in the data-protection keychain that `/usr/bin/security` cannot see.

Consequences: the encrypted cache and `granola.db` are unreadable, and `supabase.json.enc` is behind the same key, so the internal-API WorkOS token path is gone too.

The file-level crypto is unchanged. Only key acquisition died.

## Relationship to #1556

This supersedes the diagnosis in #1556. Every observation in that issue is accurate, but the inference is not: the DEK is `randomBytes(32)` wrapped by the Safe Storage key, not derived from it. The `Granola Safe Storage` keychain item is a key-encrypting key for a file that no longer exists, so decrypting `cache-v6.json.enc` with it cannot work. A fix built on that premise compiles and fails on every real machine.

## What changed

**Credential safety.** D6's read-only rule covered the encrypted supabase store but not the `stored-accounts.json` fallback, documented as refresh-allowed because it was "rarely populated on modern installs". The migration inverted that premise, making it the live token path. Rotating its single-use token signs the user's Granola desktop out. Now refused whenever an encrypted store is present, preserving refresh for genuinely legacy installs. This restores the invariant already recorded in `d6-read-only-applies-to-all-desktop-token-sources.json`.

**Honest failure.** The migrated scheme is classified from observable state rather than a version string, including the failed-migration case where Granola leaves a stale `storage.dek` behind because it only unlinks after a successful import. The error names the real cause, the Business-plan requirement, what still reads locally, and `GRANOLA_SAFESTORAGE_KEY_OVERRIDE` for anyone holding a pre-migration key file — the migration imports the existing DEK rather than rotating it, so an old copy still decrypts today's files.

**The public API list stage never worked.** Two generated defaults were guessed rather than read from the live contract: `page_size` defaulted to 100 against an API that caps at 30, and `updated_after` was formatted as local-zone RFC3339, which the API rejects as an invalid date. Verified live: `page_size=30` → 200, `31` → 400; `...T16:00:00Z` → 200, `...T09:00:00-07:00` → 400. Every list request 400'd, so `sync-api` could not complete a real sync at any point since publication.

**Detail-stage hydration.** `sync-api` previously wrote only list rows into the generic `resources` table, which no read command consumes, and the list endpoint returns just id/title/owner/timestamps. It now fetches each note with `include=transcript` and writes meetings, attendees, calendar events, summaries, folder and space membership, and transcript segments into the Granola domain tables.

**Store-backed reads.** Read commands now serve from the local store first, with the desktop cache as a fallback. Neither step makes a network call or consults an API key.

## Verification

Against a live Business-plan key on a migrated install, the acceptance case is a meeting recorded after the local cache froze:

```
$ env -u GRANOLA_API_KEY granola-pp-cli transcript get not_qJN0K2lh83tx4J | wc -l
564
$ GRANOLA_API_KEY=<set> granola-pp-cli transcript get not_qJN0K2lh83tx4J | wc -l
564
```

Byte-identical, which is only possible if the read never consults the API.

Row ownership was proven on a real store rather than a fixture. Both sync paths share the domain tables and the cache path issued an unscoped `DELETE FROM folder_memberships`; a `row_source` column plus source-scoped deletes now prevents either path destroying the other's rows. After an API sync against a store holding 867 cache-sourced meetings, the split was `api|24 cache|867` with every cache row intact.

| Gate | Result |
|---|---|
| `go build ./...` | pass |
| `go vet ./...` | pass |
| `go test ./...` | 336 passed, 0 failed (baseline was 287 passed / 1 failed) |
| `govulncheck ./...` | 0 reachable vulnerabilities |
| `verify_skill.py` | pass |
| `gofmt -l` | clean |

The baseline failure was `TestLoadCache_RealFile`, which read the operator's live cache and hit this exact migration. It now accepts the migrated classification instead of failing permanently.

## Two things worth flagging separately

The timestamp defect looks generator-shaped rather than granola-specific. Any printed CLI whose API validates timestamps strictly would hit the same local-offset RFC3339 formatting. Worth a look upstream in `cli-printing-press`.

Panels, recipes, chat, and workspaces have no public-API equivalent and remain unavailable on migrated installs. They are documented as cache-only rather than silently returning empty.

## Deferred

Doctor still reports the migrated state as `key_unavailable` with an unreachable Keychain hint, and the top-level `sync` does not yet fall back to the API automatically. Both are planned follow-ups, along with detail-stage checkpointing and source gating for the cache-only commands. Plan is in `docs/plans/2026-07-25-001-fix-granola-dual-path-data-access-plan.md`.

https://claude.ai/code/session_01J8okGyYxaUrxjFHx7qqwDv


mvanhorn:
@greptileai review

mvanhorn:
Context on how broken this actually was, for anyone finding this later.

**Every cache-backed command in `granola-pp-cli` has been dead since roughly late June.** Not degraded — dead. `sync`, `transcript`, `talktime`, `memo`, `export`, `attendee`, `panel`, `folder`, and everything else that reads meeting data failed on any Mac where Granola desktop had auto-updated. The failure was silent in the worst way: the CLI kept running, the auto-refresh hook kept firing on every invocation, and the local store simply stopped advancing. On the machine this was developed against, the store sat frozen at 867 meetings ending 2026-06-22 while the user kept having meetings.

The error it printed pointed the wrong way entirely:

```
storage.dek not found at ~/Library/Application Support/Granola/storage.dek
(Granola not installed or pre-encryption version)
```

Granola was installed, signed in, running, and actively writing its encrypted cache. That message sent people to reinstall or re-authenticate, neither of which could ever work.

**Two things were separately broken and nobody had noticed the second.** Beyond the key migration, `sync-api` had never worked against the live API since the day it was published: the generated list stage defaulted `page_size` to 100 against an endpoint that caps at 30, and sent `updated_after` as a local-zone RFC3339 timestamp the API rejects as an invalid date. Both 400 every request. So the documented fallback path for the encrypted-cache problem was itself non-functional, which is why it never surfaced as a workaround.

**The blast radius was wider than reading.** On a migrated install the token loader falls through to a stale plaintext `stored-accounts.json`, and the CLI would refresh against it — rotating a single-use token Granola desktop still holds and signing the user out of their desktop app. That is a data-adjacent failure caused by a CLI that was only trying to read.

This PR restores access through Granola's public REST API, which turns out to be far richer than this CLI assumed: `GET /v1/notes/{id}?include=transcript` returns full speaker-tagged transcripts, attendees, calendar events, and summaries. Transcripts now work with or without an API key once synced — without one you get everything already in the local store, with one you also pull new meetings.

Verified end to end on a real Business-plan account: a meeting recorded after the cache froze returns all 564 of its transcript segments, byte-identical with and without a key set.

Related: #1556 (same breakage, different and incorrect root cause — details in the comment there). Follow-ups deliberately deferred out of this PR: #1600, #1601, #1602.

PR #1599, opened 2026-07-25
feat(movie-goat): store the OMDb key in config.toml and add auth set-omdb-token

giuseppebisemi:
## Summary

Movie Goat needs two API keys, but only one of them has a home in the config file. `config.toml` recognizes `base_url`, `auth_header`, `access_token`, `refresh_token`, `client_id`, `client_secret`, `token_expiry`, and `api_key` — there is no OMDb field — while the OMDb key is read solely from `os.Getenv("OMDB_API_KEY")` at three call sites (`ratings`, `versus`, `career`). So `auth set-token` persists the TMDb key for every shell on the machine, and the second key has to be re-exported per shell, invisible to `auth status` and `doctor`.

This adds `omdb_api_key` to `config.toml`, an `auth set-omdb-token` subcommand that mirrors `auth set-token`, and a single resolution point that both call paths share. `OMDB_API_KEY` keeps working and takes precedence, so existing environment-based setups are unaffected.

## Findings

| ID | Category | Type | Rationale |
|---|---|---|---|
| F1 | credential-storage-asymmetry | feature | `config.toml` models one credential; the second API key had no persistent slot and could only arrive from the environment. |
| F2 | missing-auth-subcommand | feature | `auth set-token` exists for TMDb; there was no equivalent for OMDb, so the key was undiscoverable from `auth --help`. |
| F3 | credential-visibility | feature | `auth status` and `doctor` reported only the TMDb credential, so neither could tell you whether the IMDb / RT / Metacritic columns would populate. |

## Changes

- **`internal/config/config.go`** — new persisted `omdb_api_key` field, plus `OmdbKey()`, `OmdbSource()`, `SaveOmdbCredential()`, `ClearOmdbCredential()`. The environment value is held in an **unexported** field so `toml.Marshal` cannot write an environment-supplied secret into `config.toml` on some later unrelated `save()`.
- **`internal/cli/auth.go`** — new `auth set-omdb-token <token>`; `auth status` reports `omdb_configured` / `omdb_source`; `auth logout` clears both credentials; saving while `OMDB_API_KEY` is set warns that the environment variable still wins.
- **`internal/cli/root.go`** — `omdbAPIKey()` resolves config-then-environment in one place and never errors, so enrichment degrades exactly as before when no key is present.
- **`internal/cli/{ratings,versus,career}.go`** — the three `os.Getenv("OMDB_API_KEY")` reads now go through that resolver.
[...]

PR #1603, opened 2026-07-25
feat(movie-goat): disclose ambiguous title matches on stderr and in JSON; add --year

giuseppebisemi:
## Summary

Resolving a title to a TMDb id is a ranked guess, but the CLI presented it as a lookup. `/search/*` orders results by a relevance score TMDb does not expose, so `ratings "Sabrina"` returned the 1995 remake (703 ratings, IMDb 6.3) instead of the 1954 Billy Wilder original (1373 ratings, IMDb 7.6, RT 89%) — with nothing in stdout, stderr, or the exit code to show that a choice had been made. This makes that choice visible on both output channels and gives two ways to override it.

Worth being precise about the mechanism, because the obvious explanation is wrong: the ordering is **not** the `popularity` field and not the vote count. In this very case the 1954 entry leads the 1995 one on both readable signals — 1373 vs 703 ratings, 4.25 vs 3.60 popularity — and TMDb still returns it second. That is exactly why a caller cannot predict which edition they will get, and why the disclosure has to come from the CLI.

## Findings

|----|----------|------|-----------|
| F1 | `ambiguous-title-resolution` | bug | Title→id resolution silently took `Results[0]`. Every command that accepts a title instead of an id was affected: `ratings`, `versus`, `marathon`, `watchlist add`, plus the person-name path in `career` and `collaborators`. |
| F2 | `disambiguation-flag` | feature | No way to pin a specific entry short of looking the TMDb id up by hand. |
| F3 | `machine-readable-ambiguity` | bug | A stderr-only warning is invisible to the consumer most exposed to a silently wrong result: an unattended script reading stdout JSON with `2>/dev/null`. No human is watching that path to notice the wrong year. |

## Changes

**F1 — say when the title is ambiguous.** When a search returns more than one entry whose title matches the query exactly (normalized: case-, punctuation-, and apostrophe-insensitive), the alternatives are listed on stderr with their TMDb ids and years. The persistent `--quiet` flag suppresses the notice.

```
warn: "Sabrina" matches 3 titles on TMDb; using id 11860 — Sabrina (1995).
      TMDb's search relevance put it first, but Sabrina (1954) has more ratings (1373 vs 703).
      Other matches:
        6620  Sabrina (1954)
        503902  Sabrina (2018)
      Disambiguate with --year <YYYY>, a "title (YYYY)" suffix, or the TMDb id.
```

The comparison is on `vote_count` because that is the only ranking-relevant number the API actually gives us. A regression test asserts the notice never claims the ordering came from the popularity field.

The notice is gated on rating counts (`vote_count >= 50`, or at least as many as the chosen entry). Without that gate it fired on nearly every lookup — TMDb carries unrated same-titled shorts for `Inception`, `The Matrix`, and `Fight Club` — and a warning that always fires is one nobody reads. People have no vote counts, so the person path uses a relative-popularity gate instead and shows `known_for_department` as the discriminator.

**F2 — two ways to disambiguate.**

- `--year <YYYY>` on the single-title commands (`ratings`, `marathon`, `watchlist add`). It maps to TMDb's own `primary_release_year` / `first_air_date_year`, so the constraint is applied server-side rather than by post-filtering a popularity-ordered page.
- An inline `"Title (YYYY)"` suffix accepted by **every** title-taking command. This is what makes `versus "Sabrina (1954)" "Sabrina (1995)"` work — a single `--year` flag cannot disambiguate a two-positional command — and it generalizes to future title arguments without per-command flag plumbing. Only 1800–2999 is treated as a year, and a year-constrained search that returns nothing retries with the literal string, so a real title ending in a parenthesized number is never broken.

**F3 — the same event, in the JSON.** The resolvers append to a machine-owned `rootFlags.ambiguities` slot (the same pattern the existing `freshnessMeta` uses), and the output layer merges it into the response document under `meta.ambiguous`:

```console
$ movie-goat-pp-cli ratings "Sabrina" --agent 2>/dev/null | jq '.meta.ambiguous'
[
  {
    "query": "Sabrina",
    "kind": "titles",
    "match_count": 3,
    "signal": "alternative_better_rated",
    "chosen":  { "tmdb_id": 11860, "title": "Sabrina", "year": "1995", "vote_count": 703, "popularity": 3.5957 },
    "alternatives": [
      { "tmdb_id": 6620,   "title": "Sabrina", "year": "1954", "vote_count": 1373, "popularity": 4.247 },
      { "tmdb_id": 503902, "title": "Sabrina", "year": "2018", "vote_count": 194,  "popularity": 1.838 }
    ],
    "hint": "Disambiguate with --year <YYYY>, a \"title (YYYY)\" suffix, or the TMDb id."
  }
]
```

`signal` is `alternative_better_rated` when TMDb's top pick is not the best-rated match (the Sabrina failure mode) or `multiple_exact_matches` otherwise. The value is a list because one invocation can resolve several titles — `versus "Sabrina" "Titanic"` records two.

`popularity` is included but is documented, in the struct and in both docs, as TMDb's trending score passed through as-is: it is neither the ordering criterion nor what `signal` compares. Without that note the field invites exactly the wrong inference — as the numbers above show, the chosen entry here is the *less* popular one.

### Where the field lives, and why

`meta` was the right name but it did not previously exist on these commands. The `{results, meta}` provenance envelope is built by `wrapWithProvenance` on the `data_source` path (`movies search`, `sync`, …). The commands where this bug lives are hand-written novel commands that marshal a typed struct straight through `printJSONFiltered` — `ratings` returns `{title, ratings, …}` at the document root, with no envelope at all.

Moving them onto the envelope would have been a breaking change: every consumer reading `.title` would have to start reading `.results.title`. So `meta` is added as a sibling at the document root, which is where it sits in the envelope case too — the data is under `results` there and inline here, but in both shapes `meta` is a top-level key carrying facts about *how* the response was produced rather than what it contains. `injectAmbiguityMeta` merges into an existing `meta` object rather than replacing it, so the two paths compose.

### Compatibility, precisely

The earlier revision of this PR claimed stdout was byte-identical. That is no longer true, and here is exactly what changed:

- **The document gains one key, and only when a lookup was ambiguous.** Unambiguous lookups — the overwhelming majority — carry no `meta` key at all. There is no `"ambiguous": null` on every response.
- **No existing key changes name, type, or value.** A parser reading `.title`, `.ratings`, `.a`/`.b`, or `.entries` is unaffected. This is the additive sense of backward-compatible: a strict validator configured with `additionalProperties: false` would see a new key; ordinary parsers would not notice.
- **`--select` deliberately cannot filter it out.** Injection runs *after* `--select` and `--compact`. `--select title,ratings` still returns `meta.ambiguous` when the lookup was ambiguous. This is a considered exception to "`--select` returns only the fields you asked for": the consumer F3 exists for is precisely the script that narrows its field list and drops stderr, so letting `--select` remove the record would rebuild the blind spot inside the fix.
- **`--compact` / `--agent` keep it.** Worth stating because the compact path is not uniform: `compactListFields` is an allow-list (it would strip `meta`) but applies to top-level arrays, while these commands all emit objects, which go through `compactObjectFields` — a blocklist that leaves `meta` alone. Verified live under `--agent`.
- **`--quiet` gates the stderr notice only.** The record is written to the document regardless. One pre-existing wrinkle worth knowing: on these commands `--quiet` suppresses stdout entirely (`printOutputWithFlags` returns before printing), so it is not a way to read the field — use `--json` without `--quiet`. That behavior predates this PR and is untouched here; it also does not match the flag's help text (`Bare output, one value per line`), a separate pre-existing inconsistency this PR does not address.
- **Non-object documents are left alone.** Top-level arrays and non-JSON passthrough payloads are returned unchanged; there is no object to add a key to, and the CSV renderer expects an array.

```
 .../title-resolution-must-signal-ambiguity.json    |  25 ++
 .../movie-goat/.printing-press.json                |   6 +
 library/media-and-entertainment/movie-goat/NOTICE  |   3 +
 .../media-and-entertainment/movie-goat/README.md   |  45 +++
 .../media-and-entertainment/movie-goat/SKILL.md    |  88 +++++
 .../movie-goat/internal/cli/career.go              |   2 +-
 .../movie-goat/internal/cli/collaborators.go       |   2 +-
 .../movie-goat/internal/cli/helpers.go             |  51 +++
 .../movie-goat/internal/cli/marathon.go            |  17 +-
 .../movie-goat/internal/cli/ratings.go             |  21 +-
 .../movie-goat/internal/cli/root.go                |   6 +
 .../movie-goat/internal/cli/tmdb_helpers.go        | 358 ++++++++++++++++++--
 .../movie-goat/internal/cli/tmdb_helpers_test.go   | 375 +++++++++++++++++++++
 .../movie-goat/internal/cli/versus.go              |   5 +-
 .../movie-goat/internal/cli/watchlist.go           |  20 +-
 15 files changed, 988 insertions(+), 36 deletions(-)
```

## Verification

| Check | Result |
|---|---|
| `go build ./...` | PASS |
| `go vet ./...` | PASS |
| `go test ./...` | PASS (`internal/cli/tmdb_helpers_test.go`, 16 test functions) |
| `govulncheck ./...` | PASS — no vulnerabilities |
| `verify_skill.py --dir library/media-and-entertainment/movie-goat/` | PASS |
| Documentation checkpoint | PASS — README Unique Features + cookbook recipes 13–14, SKILL Unique Capabilities + "Pin the original, not the remake" recipe with the `meta.ambiguous` contract |
| Live dogfood against TMDb | PASS — see below |

Live checks against the real API:

**Resolution (F1/F2)**

- `ratings "Sabrina" --agent` → notice on stderr (`TMDb's search relevance put it first, but Sabrina (1954) has more ratings (1373 vs 703)`), valid JSON on stdout, id 11860
- `ratings "Sabrina" --year 1954 --agent` → id 6620, no notice
- `ratings "Sabrina (1954)" --agent` → id 6620, no notice
- `ratings 6620 --agent` → id 6620, no notice (numeric ids skip search entirely)
- `versus "Sabrina (1954)" "Sabrina (1995)" --agent` → a=6620 (IMDb 7.6), b=11860 (IMDb 6.3)
- `ratings "Titanic"` / `"Solaris"` → notice (real ambiguity); `"Inception"` / `"Fight Club"` / `"The Matrix"` / `"Heat"` → silent
- `ratings "Blade Runner 2049"` → 335984, inline-year parser correctly leaves the title alone
- `ratings "Sabrina" --year 54` → `--year must be a 4-digit year, got "54"`
- `ratings "Sabrina" --year 1899` → `no movies found for "Sabrina" released in 1899`

**JSON record (F3), all run with `2>/dev/null` so only stdout is read**

- `ratings "Sabrina" --agent` → `meta.ambiguous[0]` present: `signal=alternative_better_rated`, `match_count=3`, `chosen=11860`, `alternatives=[6620, 503902]`
- `ratings "Inception" --agent` → `.meta` is `null` — the field does not appear when the stderr notice does not fire
- `ratings "Sabrina" --agent --select title` → keys are `["meta","title"]`; `--select` does not remove the record
- `ratings "Inception" --agent --select title,ratings` → no `meta` key
- `ratings "Sabrina" --year 1954 --agent` → no `meta` key (the year resolved it, so there is nothing to disclose)
- `versus "Sabrina" "Titanic" --agent` → two records: `Sabrina/alternative_better_rated` and `Titanic/multiple_exact_matches`
- `marathon "Ocean's Eleven" --agent` → `signal=multiple_exact_matches`
- `career "David Jones" --agent` → person record with `kind=people`, `known_for` populated on chosen and alternatives (`Visual Effects`, `Sound`, `Acting`, …)
- `collaborators "John Williams" --agent` → person record, chosen 491 (`Sound`), alternatives include 5182 (`Acting`)

Two `printing-press publish validate` checks fail identically on `upstream/main` before this change and are untouched by it: `manifest` (`schema_version must be 2 (found 1)`) and the press binary's `verify-skill` install-section drift. Both come from this CLI having been printed with press 3.8.0 and are reprint concerns, not amend concerns. The repo's own `verify_skill.py` — the check CI runs — passes.

## Notes for review

- **Overlap with #1599** (same author, open). That PR moves the OMDb key into `config.toml` and touches `internal/config/`, `internal/cli/auth.go`, `internal/cli/root.go`, and the OMDb key lookup inside `ratings.go` / `versus.go`. This branch is cut from `upstream/main`, not from #1599, and changes the title-resolution block of those two files plus files #1599 does not touch. Two deliberate overlaps: the contributor attribution (`.printing-press.json`, README byline, `NOTICE`), which both PRs add per AGENTS.md with identical content; and `internal/cli/root.go`, where #1599 edits the OMDb key accessor while this PR adds one field to the `rootFlags` struct. Both resolve mechanically.
- No labels applied: `comp:movie-goat` and `priority:*` do not exist in this repo.
- No `deferred_to_upstream` entry. `tmdb_helpers.go` is hand-written novel-feature code rather than generated per-endpoint output, and TMDb's remake-heavy catalogue is what makes the ambiguity load-bearing — this is a single-CLI fix, not a generator template bug. The `injectAmbiguityMeta` hook in `helpers.go` is generic enough that the generator could grow an equivalent, but nothing else in this CLI needs it today.

## Evidence

- Patch record: [`.printing-press-patches/title-resolution-must-signal-ambiguity.json`](https://github.com/mvanhorn/printing-press-library/blob/4ef4e5be46704ee0205d5d8ca646723fab96e56d/library/media-and-entertainment/movie-goat/.printing-press-patches/title-resolution-must-signal-ambiguity.json)
- Patch directory at this HEAD: [`.printing-press-patches/`](https://github.com/mvanhorn/printing-press-library/tree/4ef4e5be46704ee0205d5d8ca646723fab96e56d/library/media-and-entertainment/movie-goat/.printing-press-patches)
- Resolution helpers: [`internal/cli/tmdb_helpers.go`](https://github.com/mvanhorn/printing-press-library/blob/4ef4e5be46704ee0205d5d8ca646723fab96e56d/library/media-and-entertainment/movie-goat/internal/cli/tmdb_helpers.go)
- Tests: [`internal/cli/tmdb_helpers_test.go`](https://github.com/mvanhorn/printing-press-library/blob/4ef4e5be46704ee0205d5d8ca646723fab96e56d/library/media-and-entertainment/movie-goat/internal/cli/tmdb_helpers_test.go)


tmchow:

giuseppebisemi:
Same situation as #1599, context there. This one is 15 files, also under the cap.

PR #1604, opened 2026-07-26
fix(substack): clamp sync page size to Substack's limit cap

### LINKED ISSUE #1524: sync: limit=100 exceeds Substack's cap of 50, 400s on drafts/archive/published
**CLI:** `substack-pp-cli 2026.7.1` (macOS, arm64)
**Date observed:** 2026-07-15

### Summary

`sync` fails on three list resources because the CLI requests `limit=100`. Substack now rejects any `limit` above **50** on these endpoints with an HTTP 400 validation error. `limit=50` works; `limit=51` fails. The web app itself requests `limit=25` and paginates with `offset`.

### Affected endpoints

All three 400 with `limit=100`:

- `/api/v1/post_management/drafts`
- `/api/v1/post_management/published`
- `/api/v1/archive`

### Evidence (reproduced against my own publication, signed in, `credentials: include`)

| Request | Status | Response body |
| --- | --- | --- |
| `post_management/drafts` … `limit=100` | 400 | `{"errors":[{"location":"query","param":"limit","value":100,"msg":"Invalid value"}]}` |
| `post_management/drafts` … `limit=25` | 200 | ok |
| `archive?limit=100` | 400 | `{"errors":[{"location":"query","param":"limit","value":100,"msg":"Invalid value"}]}` |
| `post_management/published` … `limit=100` | 400 | `{"errors":[{"location":"query","param":"limit","value":100,"msg":"Invalid value"}]}` |

What Substack's own dashboard sends (HTTP 200) — note the lower `limit` and `offset` pagination:

```
GET https://<pub>.substack.com/api/v1/post_management/drafts?offset=0&limit=25&order_by=draft_updated_at&order_direction=desc
GET https://<pub>.substack.com/api/v1/post_management/published?offset=0&limit=25&order_by=post_date&order_direction=desc
```

Boundary test on `post_management/drafts`:

```
limit=25 => 200
limit=40 => 200
limit=50 => 200
limit=51 => 400
limit=75 => 400
limit=100 => 400
```

### Root cause

Substack added server-side validation that caps `limit` at **50** on these list endpoints. The CLI hardcodes `limit=100`, so `drafts`, `archive`, and `post_management/published` all 400 during `sync`.

### Suggested fix

- Cap the requested page size at 50 (or make it configurable, default `<= 50`), and paginate the rest with `offset` the way the web app does.
- Treat a `400` whose body is `{"errors":[{"param":"limit",...,"msg":"Invalid value"}]}` as a "reduce page size and retry" signal rather than a hard resource failure.

### Secondary note

The default exit policy downgrades partial per-resource failures to warnings and still exits 0, which makes these silent. `--strict` restores a non-zero exit, but a distinct exit code (or a non-zero default) would help schedulers detect the breakage.

MBemera:
Fixes #1524.

## What was broken

Substack validates the `limit` query param server-side on its list endpoints. Anything above the cap is rejected with

```json
{"errors":[{"location":"query","param":"limit","value":100,"msg":"Invalid value"}]}
```

`determinePaginationDefaults()` returns `limit: 100`, and `resourceSupportsPagination()` covers exactly the three resources the issue names — `drafts`, `posts` (`/api/v1/archive`), and `posts-published` (`/post_management/published`) — so all three failed outright.

## The fix

One clamp at the single point of use:

```go
pageSize := determinePaginationDefaults()
pageSize.limit = clampSyncPageSize(resource, pageSize.limit)
```

`pageSize.limit` feeds **three** coupled call sites: the request param (`sync.go:539`), the short-page stop check (`sync.go:731`), and the **offset stride** (`sync.go:739`). Clamping the one value keeps them in agreement — lowering only the request limit would have left the stride advancing by 100 while fetching 50, silently skipping records. That coupling is why the clamp lives at the assignment rather than at the request.

`determinePaginationDefaults()` is left untouched so it stays recognizably generator-emitted.

`/post_management/published` caps lower than the other endpoints, so the clamp is keyed by resource and reuses the existing `publishedPostsPageSize` constant already proven against that endpoint in `portfolio_sync.go` — rather than introducing a second, competing number for the same cap.

## Validation

Boundary measured live against two publications (`www.thefp.com`, `astralcodexten.substack.com`): `limit=50` → 200, `limit=51` → 400. Matches the reporter's boundary test exactly.

End-to-end, pre-fix binary built from `main` vs. patched binary, same publication and flags:

| | `sync --resources posts --strict` | exit | records |
|---|---|---|---|
| `main` | `sync_error` status 400 | 1 | 0 |
| this PR | `sync_complete` | 0 | 23 |

Reproduced on both `astralcodexten` and `platformer`. Rows land in SQLite and are FTS-indexed (`resources: 23 rows`, `resources_fts: 23 rows`).

Seven unit tests in `sync_page_size_test.go` cover the clamp, the per-resource published cap, the "already smaller is left alone" case, a sweep asserting every paginated resource clears the cap, and an `httptest` server reproducing Substack's validation to prove the generated size is rejected and the clamped size is accepted.

Gates: `go build` ✅ `go vet` ✅ `go test ./...` ✅ `gofmt` ✅ (the 5 unformatted files in `internal/cli/` are pre-existing on `main`) `govulncheck ./...` ✅ no vulnerabilities in reachable code. `verify_skill.py` output is byte-identical to the `main` baseline — the 6 `discover` flag findings it reports are pre-existing and untouched by this change.

## Two things deliberately left out of scope

**1. These resources still fetch only their first page.** `extractPageItems` returns `hasMore=false` for a bare JSON array (`sync.go`, strategy 1), and the loop breaks on `!hasMore`, so `/api/v1/archive` never advances past page 1 at any page size — verified live: requesting 12 yields 12 records, requesting 50 yields 23 (Substack's own first-page ceiling). This is a **separate defect** from the limit cap, it sits in the generic generated pagination path, and changing it affects every resource. Worth its own issue; I did not widen this PR to cover it.

**2. The exit-code note.** The issue's secondary note asks for a distinct exit code or non-zero default on partial per-resource failure. That changes documented default behavior for every consumer of `sync`, so it belongs in its own change rather than riding along with a bug fix. `--strict` already restores a non-zero exit today.

## Second commit: a pre-existing CI failure this PR had to clear

`Validate SKILL.md against shipped CLI source` fails on `main`'s content for this CLI — the verifier only runs against CLIs a PR touches, so it stays invisible until someone edits substack, and it blocked this PR.

Cause: `discover patterns` was attached by scanning `rootCmd.Commands()` for the literal name `"discover"` at runtime. Correct at execution time, but `verify_skill.py` resolves documented commands by walking `AddCommand` call sites in source, and a loop-variable receiver isn't statically resolvable — so the subcommand never entered the verifier's tree, its own `--niche/--sort/--since` flags read as "declared elsewhere but not on discover", and `patterns` parsed as a positional arg of `discover`.

Binding the parent to a variable before attaching the child is simpler than the loop and analyzable. Verifier goes from 6 errors + 2 false positives to **all checks passed**; `discover --help` and `discover patterns --help` are byte-identical. Happy to split this into its own PR if you'd rather keep the bug fix isolated.


MBemera:
issue with greptile policygate time out. other checks passed. -MB

tmchow:

MBemera:
@greptileai review

Not triggering 

** @tmchow unsure what is going on here and am away from my computer. 
-MB

PR #1605, opened 2026-07-26
feat(slack): add canvases resource (create, read, edit, delete, share)

dericknwq:
> [!NOTE]
> **Rebased onto post-reprint `main` (2026-08-07).** This was the third of a three-PR stack; #1652
> (the v4.29.0 reprint) landed underneath it and made most of #1594 redundant, so the surviving work
> is consolidated here and this PR now stands alone against `main`. #1594 and #1596 can be closed as
> superseded.

## Summary

Three things, all against the v4.29.0 reprint.

### 1. A `canvases` resource

`slack-pp-cli` has no canvas commands. Not broken — **absent**. The curated spec carries no canvas
endpoints, so the generator emits nothing for them, and canvases are a primary Slack content type.
Their absence was a standing reason to keep the Slack MCP connector running alongside this CLI. The
reprint added 29 endpoints and none of them were canvas ones, so this is still missing on `main`.

| Command | Endpoint |
|---|---|
| `create` | `canvases.create`, or `conversations.canvases.create` with `--channel` |
| `read` | `files.info` → `url_private_download` (see below) |
| `edit` | `canvases.edit`, all six operations |
| `delete` | `canvases.delete` |
| `access-set` | `canvases.access.set`, users and/or channels |
| `sections` | `canvases.sections.lookup` |

Requires `canvases:write`, `canvases:read`, and — for `read` — `files:read`.

**Reading, which has no endpoint.** Slack exposes no get-canvas-content call, and `files.info` returns
no `preview` for a canvas. But a canvas is backed by a file, so `files.info`'s `url_private_download`
serves the document body as HTML. Third-party tooling does this with **browser-session credentials**
(`xoxc` token + `xoxd` cookie) scraped out of Slack Desktop's LevelDB and cookie store. That seemed
worth avoiding, so I checked whether the download works with ordinary OAuth: **it does** — a plain
`xoxp` user token with `files:read` returns HTTP 200 and the full HTML, no cookie involved. So this
needs no session scraping and nothing that breaks when Slack Desktop restarts.

The HTML also carries section ids inline, which makes `read` strictly more useful than `sections` for
targeting edits — `sections.lookup` returns ids with no content, so pairing an id to the text it
covers previously meant guessing.

**Caveat, documented in `--help`, the README and the patch record:** content returns as HTML, not the
Markdown it was created from, so create/read is not lossless. `--format text` is a deliberately crude
tag strip for grepping, not a converter.

**`client.FetchRaw`.** The download needed a new client method: `files.slack.com` is a different host
serving non-JSON, so `do()`'s path joining and envelope handling don't apply. It carries the workspace
credential — in direct contrast to the external upload URL, which must *not* receive one — and its URL
comes from an API response rather than from the operator. So it refuses any host that isn't
`slack.com` or a `.slack.com` subdomain, and refuses non-https schemes. A spoofed or tampered
`files.info` response can't redirect the token elsewhere. Tests cover the suffix-confusion cases
(`slack.com.evil.example.com`, `files.slack.com.attacker.net`) alongside plaintext `http`. No caching,
no rate limiter, no retry: it fetches one document on demand and the response isn't an envelope any of
that machinery understands.

**Structure.** One file behind a shared `postCanvas` helper, rather than the generated
one-file-per-command layout. Six copies of the client/error/envelope boilerplate would be harder to
review and keep consistent, and since no template produces these commands there's no generated output
for a future reprint to diff against. Flagged because it's a deliberate deviation from the surrounding
convention.

Argument validation is done locally wherever Slack would otherwise answer a bare `invalid_arguments` —
the six-value operation enum, the access level, and which operations require `--section` or content. On
an API with six edit operations that each have different requirements, a generic rejection is a
usability trap.

### 2. File upload migrated to the external flow

Slack retired `files.upload`; it answers `method_deprecated` regardless of the request body. On the
reprinted tree the generated command is doubly broken: it posts a JSON body to `/files.upload` **and
declares no `--file` flag at all**, so the only source it can express is inline `--content` — sent to
an endpoint that rejects everything.

The replacement is three steps: `files.getUploadURLExternal` reserves a URL for a filename and length,
the raw bytes go to that URL, `files.completeUploadExternal` registers the file and optionally shares
it. Step 2 deliberately does not carry the workspace credential — the reserved URL holds its own
single-use token and the host is not `slack.com`.

**The MCP surface had the same defect in a different shape.** The reprint replaced the static
`tools.go` upload tool with the code-orchestration descriptor table, which advertises `files.upload`,
so an agent that discovers it through `slack_search` and executes it always fails. The generic
dispatcher can't express the replacement, so `handleCodeOrchExecute` intercepts that endpoint id and
calls the shared `client.UploadExternal`. Both surfaces upload by the same code path rather than the
CLI being fixed and the MCP descriptor left pointing at a dead endpoint.

The source is opened once and sized through that same handle. Sizing a path with `os.Stat` and opening
it later leaves a window in which the file can be appended to, truncated or replaced, so the length
reserved in step 1 wouldn't match the bytes streamed in step 2. The copy is bounded by `LimitReader`
with a probe read after it, so a source that shrank or grew errors rather than uploading a prefix and
reporting success. The bytes stream, so a large upload doesn't have to fit in memory.

`--stdin` is dropped rather than carried over: it fed a free-form JSON body to the retired endpoint,
and the external flow has no equivalent.

### 3. `ok:false` on the command path, and live search

Slack signals application failure with HTTP 200 plus `{"ok":false,"error":"..."}`, so transport-level
handling never sees it. The reprint's own `slack-ok-false-detection` patch guards the **sync loop**
(`sync.go`, `archive_sync.go`) only. Commands run `detectPartialFailure`, which matches the
Google-Ads-shaped `partialFailureError` field and does not recognise Slack's shape, so a write to a
nonexistent channel printed the error envelope as a successful result and exited 0.

Covered here: 53 generated write commands, live search, and per-record during `import` — which
discarded the response body entirely, so a record Slack rejected counted as a success and an import
where every record was rejected reported "N succeeded, 0 failed" and exited 0. The exit-code plumbing
already moved with `failed`; detection was the only missing half.

`checkSlackAPIError` wraps with `apiErr` so every caller matches the documented table (5 = API error)
without touching call sites. Returned bare, read paths reached 5 anyway via `classifyAPIError`'s
default branch while write paths exited 1.

**Live search** separately returned one result for any query: `searchResponsePaths` is empty on `main`,
so `extractSearchResults` falls through every generic wrapper key and hands back the whole envelope as
a single item. Note the symptom **differs from the pre-reprint tree**, where the envelope was also
discarded by `isNilOrEmpty` and the command printed "No results" and exited 0 — the reprint's generic
`hasAnyNonEmptySearchValue` recurse keeps the envelope, so it now renders as one row with no
`text`/`ts`/`permalink` while the real matches stay buried in it. Because that recurse also keeps a
real match, naming `messages.matches` is the entire fix on this tree; `responsePayloadAtPath` already
resolves dotted paths, so no traversal code was added.

`verify_skill.py`: `check_flag_commands` attributed a subcommand's flags to its parent when the
subcommand name contains an underscore, so `messages post_message --text` reported `--text` as
undeclared on `messages`. Resolved as a strict fallback after the parent is ruled out — trying the
child first misreads commands that take a resource name positionally, where `export messages --format`
resolves to a `messages` source file that declares no `--format`. This fixes the check on `main` too;
happy to split it out if you'd rather.

## Verification

`go build`, `go vet`, `gofmt`, `go test ./...` all clean. The SKILL.md verifier and its unit tests pass
locally, as does `govulncheck`.

Canvas and upload surfaces exercised against a built binary: every subcommand registers; the
validation paths each produce a named error and exit 2; `canvases edit --dry-run` emits the correct
nested `document_content`/`changes[]` body; `--dry-run --quiet` prints zero bytes on every branch;
`canvases read --dry-run` and `files upload --dry-run` emit their own multi-host plans; and `--canvas`
is validated even under `--dry-run`. The search fix has an explicit negative control — with the
response path, 2 matches; without it, 1 item that survives the filter.

**Not re-verified against a live workspace.** The original round trip (create from a Markdown file,
sections lookup, `insert_at_end`, `insert_after` a looked-up id, `access-set`, a channel canvas against
a DM, read in both formats, delete, and confirming the workspace canvas list was clean afterwards) ran
on the pre-reprint tree. The patch records state this rather than implying otherwise. Happy to re-run
it before merge if you want that on the record.

## Upstream

Recorded in `.printing-press-patches/` as four `schema_version: 2` records against base run
`20260801-150754-271fdf29`, each carrying a `deferred_to_upstream` note. None filed yet; happy to open
any of them.

1. **The spec catalog has no canvas endpoints**, so a reprint drops this resource entirely. This is no
   longer hypothetical — it is exactly what #1652 did, and why this patch had to be re-applied by
   hand.
2. **The generated body template can't express nested objects.** Canvas endpoints take
   `document_content{type,markdown}` and `changes[]{operation,document_content,section_id}`, while the
   body builder maps flat scalar flags to top-level keys. Even with the endpoints catalogued, these
   commands would still need hand-writing.
3. **No response-envelope error hook.** This is the second hand-patch for `ok:false` on this CLI. A
   per-spec `error_envelope` declaration (field path, truthiness, an access-denied code list) would let
   the sync loop and every generated command fail generically, instead of patching `sync.go` and then
   53 command files by hand — which a reprint silently undoes.
4. **Multipart/streaming and multi-step endpoints aren't modelled.** A POST whose payload is a byte
   stream gets no `--file` flag, and one leg of a three-call flow gets emitted as a standalone command
   pointing at a retired path. Flagging spec endpoints the upstream API has deprecated would also have
   caught this.


tmchow:

dericknwq:
All checks green, Greptile **4/5**. Fixed since the `needs-author` triage:

- **`canvases read --quiet` printed the canvas body** — it ignored `--quiet` in both output branches, so document content landed in scripts and logs that had asked for silence. The fetch still happens, so failures still surface: a nonexistent canvas exits 5 under `--quiet` while a successful one prints zero bytes.
- **The dry-run branch then reintroduced it** — `--dry-run --quiet` printed the plan, because that branch returns before reaching the quiet guard. It checks quiet too now.
- **`--canvas` was skipped under `--dry-run`**, following the generated commands' `!dryRun` gate, which let a dry run report success for a request carrying an empty `canvas_id` — hiding the exact class of mistake a dry run exists to catch. Now validated unconditionally across `edit`, `delete`, `access-set`, `sections` and `read`.
- **`canvases read --dry-run` errored on a valid request** — it makes two requests against two hosts, so without its own dry-run branch `c.Get` returned the synthetic dry-run envelope, which the code parsed as a `files.info` response and rejected for lacking a download URL.

The stale upload limitation you flagged in `README.md`/`SKILL.md` is fixed in #1596, which this branch sits on.

tmchow:

dericknwq:
Greptile is **5/5** on `bb6d3200e` — "no blocking failure remains".

On the finding itself: `postCanvas` was not printing under `--quiet`. That branch ends in
`printOutputWithFlags`, which returns `nil` on quiet before writing anything, so the terminal path
was already silent — and identical in shape to the generated POST commands (`pins add` and the rest
have the same nesting). I checked rather than assumed: `canvases edit --dry-run` under a pty, with
and without `--quiet`, against binaries built either side of the commit. All four outputs match, and
under `--quiet` the only text is the dry-run plan, which `client.dryRun` writes to stderr.

I moved the guard anyway, because the code invited the report. The only visible `quiet` check sat
nested inside the `--json` branch, so the suppression read as conditional on `--json` and you had to
open `helpers.go` to see that it wasn't. It now sits once above both branches, matching
`canvases read`, which already checks quiet ahead of its two output branches. No behaviour change;
the patch record carries the pty comparison and notes why that branch has no unit coverage
(exercising it needs a writer that is both a character device and readable back).

One thing to flag before re-triage: the three `pull_request` checks — **Verify**, **Govulncheck**
and **Validate SKILL.md** — were never dispatched for this commit. Only the two `pull_request_target`
Greptile workflows ran. They aren't queued, in progress, or sitting in `action_required` either; the
runs simply don't exist for `bb6d3200e`, where they fired within seconds of the previous push on
`40d88e63` and all passed. Looks like a dropped Actions dispatch rather than anything in the change.
Happy to close/reopen the PR to force a re-dispatch if that's the easiest fix — say the word and I'll
do it, or re-run from your side if you'd rather not have the churn.

dericknwq:
Correcting my last comment: the three missing checks were **not** a dropped Actions dispatch, and
closing/reopening would not have fixed them. Sorry for the misdiagnosis — no re-run needed on your
side.

The PR was in conflict. `Verify`, `Govulncheck` and `Validate SKILL.md` are `pull_request` workflows,
so they run against the computed merge commit `refs/pull/1605/merge` — which GitHub cannot create
while the merge conflicts, so the runs are never queued. The two Greptile workflows are
`pull_request_target`, run against the base branch ref, and were unaffected. That is exactly the
split I saw and misread as a dropped dispatch.

The conflict came from #1652, the v4.29.0 reprint, which merged 18 minutes before my last push. I
pushed on top of it without noticing the base had moved under me.

I have rebased onto post-reprint `main`. All three checks dispatched within a minute and pass —
including `Validate SKILL.md`, one of the two failures in your triage. All five checks on the PR are
now green, Greptile Review included.

## What the reprint made redundant

Most of the old #1594 was the POST-body-from-flags work. The v4.29.0 generator emits that natively
now, so I dropped it — that alone was ~25 of the conflicting files. What is left is genuinely still
missing on `main`, and I re-derived each against the new tree rather than carrying the old claims
over:

- **`ok:false` on the command path.** The reprint's own `slack-ok-false-detection` patch guards the
  sync loop (`sync.go`, `archive_sync.go`) only. Commands run `detectPartialFailure`, which matches
  the Google-Ads-shaped `partialFailureError` field and does not recognise Slack's shape, so a write
  to a nonexistent channel still printed the error envelope as a result and exited 0. 53 write
  commands, live search, and per-record in `import`, which discarded the response body entirely so a
  rejected record counted as a success.
- **Live search.** `searchResponsePaths` is empty on `main`, so `extractSearchResults` falls through
  every generic wrapper key and returns the whole envelope as one item. Worth flagging that the
  symptom **changed** under the reprint: it is no longer "No results, exit 0" — the new generic
  `hasAnyNonEmptySearchValue` recurse keeps the envelope, so a query with N hits renders one row with
  no text/ts/permalink while the matches stay buried in it. Because that same recurse already keeps a
  real match, naming `messages.matches` is the entire fix here; the hand-rolled envelope struct and
  the `isNilOrEmpty` change the old patch needed are both unnecessary now. I ran the negative control
  rather than assume it: with the path, 2 matches; without, 1 item that survives the filter.
- **File upload.** `main` still posts to the retired `/files.upload` and declares no `--file` flag at
  all, so the only source it could express was inline `--content` — to an endpoint that rejects
  everything.
- **Canvases.** Still absent. The reprint added 29 endpoints, none of them canvas ones, which
  independently confirms the upstream note below.

## One thing that changed shape

The MCP half of the upload fix is different now. The reprint replaced the static `tools.go` upload
tool with the code-orchestration descriptor table, which advertises `files.upload` — so an agent that
finds it via `slack_search` and executes it always fails. The generic dispatcher cannot express the
replacement (one request, JSON body; the flow needs raw bytes to a second host without the workspace
credential), so `handleCodeOrchExecute` intercepts that endpoint id and calls the shared
`client.UploadExternal`. Both surfaces now upload by the same code path rather than the CLI being
fixed and the MCP descriptor left pointing at a dead endpoint.

## Verification

`go build`, `go vet`, `gofmt`, `go test ./...` clean. The SKILL.md verifier and its unit tests pass
locally, as does `govulncheck`. Canvas and upload surfaces re-exercised against a built binary:
validation paths, dry-run plans for both two-host and three-host flows, and `--quiet` silent on every
branch. All argument validation now exits 2 per the documented table — it returned bare errors and
exited 1 before, while the required-flag paths reached 2 only by accidentally matching
`isCobraUsageError`'s prefix.

**Not re-verified live.** The original live-workspace round trip ran on the pre-reprint tree. I have
not re-run it here, and the patch records say so explicitly rather than implying otherwise.

## Stack

#1594 and #1596 are now empty of anything that survives — the upload work is in this branch and the
rest is upstream in the reprint. Happy to close both as superseded, or split this back into three if
you would rather review it that way; say which and I will restructure.

The two upstream items still stand, neither filed yet. The first one is no longer hypothetical: the
spec catalog carrying no canvas endpoints is exactly why the reprint dropped this resource and I
re-applied it by hand. Happy to open both as issues.

tmchow:
Thanks for the follow-up. This now needs maintainer judgment rather than another mechanical triage pass, so I am routing it for a maintainer decision.

PR #1606, opened 2026-07-26
fix(slack): query the messages index in local search

dericknwq:
> [!IMPORTANT]
> **Fourth in a stack.** #1594 → #1596 → #1605 → this. The local-search fix is the single commit on
> top; the diff below is cumulative. It has a real dependency on #1594, described under
> *Depends on #1594*.

## Summary

`search --data-source local` could not find messages, and said so as `No results` with **exit 0** —
a silent wrong answer indistinguishable from a genuine miss.

`store.SearchMessages` existed and was correct. It had **zero callers** anywhere in the tree:

- `search.go`'s empty-type branch queried only `SearchUsergroups` and `SearchFiles`, despite a
  comment claiming it searched *all* FTS-enabled tables.
- `--type messages` fell through to the generic `db.Search`.

Meanwhile `sync` populates `messages` and `messages_fts` either way. So the index held matches no
code path could reach.

Reproduced against a real synced database: `messages_fts` held **40 rows** matching a term the CLI
reported as absent. It now returns exactly 40.

This was called out as deliberately out of scope in #1594, with the local mode's breakage documented
under Known Limitations there. This closes it.

## The fix

Two lines of wiring: a `case "messages"`, and a messages block in the all-types branch.

**Messages are queried first, deliberately.** `outputSearchResults` truncates the aggregate to
`--limit`, so whichever source runs first survives the cut. Appending messages after usergroups and
files would let a `search foo --limit 10` return ten usergroups and zero messages — the same
practical failure as the bug being fixed, with a different cause.

## Depends on #1594

Locally synced messages carry `text` and `ts` but no top-level `title`/`name`/`identifier`/`id`.
Without #1594's `isNilOrEmpty` extension — which taught the filter to recognise a message by
`text`/`ts`/`permalink` — the rows this patch newly returns would be discarded as "empty" and local
search would *still* report nothing.

The two patches are only correct together. Recorded in the patch record so they don't get unpicked
independently.

## Architecture: stayed on typed tables

The current generator favours a single generic `resources` table; this CLI uses typed per-resource
tables. It is already half-converted — `sync` writes users and conversations into the generic table
while messages go to a typed one.

This patch stays on the typed side, deliberately. All eight of this CLI's novel analytics commands
(`digest`, `health`, `quiet`, `trends`, `funny`, `activity`, `response_times`, `threads_stale`) read
the typed store. Converting would break them for no user-visible gain, and a probe of press 4.29.0
output showed the regenerated store drops typed tables entirely along with the messages sync that
fills them — so a reprint would leave nothing for local message search to find at all.

Recorded as `deferred_to_upstream`: the unresolved store architecture is a generator-level decision,
not something to settle inside one CLI.

## On cli-printing-press#1390

#1390 covers this defect family — the generated empty-type branch querying a subset of available
typed FTS sources while claiming otherwise — and was **closed 2026-05-15**.

A probe of press `4.29.0` output shows the mismatch still present and inverted: the regenerated
branch calls only the generic `db.Search` while its comment claims it searches typed tables too.
That's worth reopening or refiling with the evidence, rather than treating this CLI as the outlier.
Referenced from the patch record; happy to file it separately.

## Verification

- `messages_fts` ground truth: 40 rows for the test term → CLI now returns exactly 40, for both the
  all-types branch and `--type messages`
- A term absent from the index (0 rows in SQLite) still reports nothing — the fix does not
  over-report
- `--limit 5` and `--limit 10` return 5 and 10, so aggregate truncation still applies
- A hit and a genuine miss both exit 0; an empty result is legitimate, not an error

`search_local_amend_test.go` adds four cases driving the **real search command** against a throwaway
store seeded through `store.UpsertMessage` — not `SearchMessages` directly. That distinction is the
point: the store helper was always correct, so a store-level test would have passed against the
broken build and proved nothing.

**Confirmed the suite catches the regression.** With the pre-fix `search.go` restored, two of the
four cases fail with the exact production symptom:

```
--- FAIL: TestLocalSearchFindsMessages/all_types_finds_a_message
    expected the seeded message in output, got: No results (source: local)
--- FAIL: TestLocalSearchFindsMessages/explicit_--type_messages_finds_it_too
    expected the seeded message in output, got: No results (source: local)
```

`go build`, `go vet`, `gofmt`, and the full `internal/cli` suite are clean.

## Still not searchable locally

**Users and conversations**, for an unrelated reason: `sync` writes them into the generic `resources`
table rather than their typed tables, so there is no per-type index to query. Separate defect, not
addressed here, noted in the patch record so it isn't mistaken for part of this one.


tmchow:

dericknwq:
All checks green, Greptile **5/5**. Fixed since the `needs-author` triage:

- **`SearchMessages` bypassed FTS5 sanitization** — the sharpest finding of the round. `Search`, `SearchUsergroups` and `SearchFiles` all route through `sanitizeFTSQuery`; this one passed the query raw. Harmless while it had no callers, reachable the moment this PR wired it in, and the same input then meant different things depending on which index answered:

  ```
  unmatched " quote  -> SQL logic error: unterminated string
  task NOT review    -> FTS5 NOT applied, silently changing the query
  ```

  Both are literal text now, while an ordinary term still returns all 40 matches in the test corpus. An empty query short-circuits instead of reaching FTS5. Six syntax cases plus an ordinary-term control added; confirmed the suite catches the regression — reverting `store.go` fails on the unmatched-quote case with the exact production error.

- **Stale limitations** — `README` and `SKILL` still said local search never returns messages, which this PR fixes. Replaced with what's actually true: messages, usergroups and files are covered; users and conversations are not, because `sync` writes them to the generic `resources` table rather than a typed per-resource index.

On the generator side, the `4.29.0` probe evidence behind this is filed as cli-printing-press#3810, since #1390 was closed while the mismatch persists in a different form.

tmchow:
ppl-triage: maintainer-fixing 55fb593e3d2cfde222032f84101a6939693d15a4

I am repairing the merge conflict against current origin/main and will preserve the PR's scoped Slack changes. I will not merge this PR.

tmchow:
ppl-triage: needs-author

I stopped the maintainer repair after the conservative rebase gate. Current origin/main now contains Slack reprint #1652 and its generated tree no longer has the typed messages_fts/SearchMessages architecture this PR targets: current search.go routes local queries through generic db.Search. Rebasing the cumulative stack therefore conflicts across the generated Slack CLI, and porting only #1606 would require author intent about which old-stack behavior to reimplement in the new tree.

Please rebase/reprint this change against current main and resubmit the local-message-search fix against the current Slack architecture, or close this superseded stack. No maintainer code commit or push was made, and this PR was not merged.

dericknwq:
Closing this as superseded, and confirming the bug is actually fixed rather than just dropped.

You were right that the stack targets an architecture that no longer exists. Slack reprint #1652 (v4.29.0) removed `messages_fts`, `SearchMessages` and `sanitizeFTSQuery` outright — `grep -rn "SearchMessages\|messages_fts\|sanitizeFTSQuery" internal/` on current `main` returns nothing. There is no typed messages index left to route to, so there is nothing in this PR to port.

**The user-visible bug is gone too.** The reprint solved it a different way rather than leaving it open: `archive sync` writes message history into the *generic* store as `conversations_history` (and `conversations_replies`), which feeds `resources_fts` — and the no-`--type` branch of `search` calls `db.Search`, which reads that index. I checked that end-to-end rather than by inspection:

```go
db.Upsert("conversations_history", "C999:1720000000.000100",
    json.RawMessage(`{"type":"message","text":"the deployment rollback decision was made here",...}`))

got, _ := db.Search("rollback", 50)   // the exact call `search --data-source local` makes
// => 1 hit, containing the message
```

Passes on `main`. So `search --data-source local` does return messages today. Closing.

The two upstream fixes from the earlier PRs in this stack (#1594, #1596) survived the reprint — `searchResponsePaths = []string{"messages.matches"}` and the `checkSlackAPIError` guard are both in `internal/cli/search.go` on `main`, PATCH comments intact. Nothing lost there either.

### One thing I found while verifying, which is not mine to fix here

`sync`'s own `Example` block advertises a command that cannot work:

```
# Sync specific resources only
slack-pp-cli sync --resources channels,messages
```

Neither `channels` nor `messages` is a sync resource — `knownSyncResourceNames()` lists 14 and includes neither, and messages deliberately come from `archive sync` instead, since `conversations.history?channel=<id>` is a query-param-keyed dependent the generated walker cannot express. Built from `main` and run:

```
$ slack-pp-cli sync --resources channels,messages --dry-run
Error: 2 resource(s) failed to sync: channels, messages
$ echo $?
1
```

A valid resource dry-runs clean by contrast, so this is the example itself, not the environment.

This is **not** a slack problem — it is a generator template with those two names hardcoded, and **301 CLIs in the library ship the identical line**. Slack is just the worst place for it to land, because "channels" and "messages" look entirely plausible here, so a user has every reason to trust it and no reason to suspect the example over their own setup. It also points people away from `archive sync`, which is the thing that actually mirrors messages.

Filing that against `cli-printing-press` rather than patching 301 generated files — will link the issue here once it is up.

dericknwq:
Generator issue filed: mvanhorn/cli-printing-press#4052 — measured at 301 of 301 library CLIs where neither placeholder name is a real sync resource.

PR #1612, opened 2026-07-26
feat(overpass): add overpass

justinwfu:
## overpass

Find things worth photographing by name instead of by OpenStreetMap tag, with automatic failover across Overpass mirrors.

**API:** Overpass | **Category:** travel | **Press version:** 4.29.0
**Spec:** `overpass.yml` (Overpass API 0.7) | **Run:** `20260726-090225`

### Publication Path

New print. No merged CLI and no existing PR matched this slug (verified pre-package). `overpass` is the first Overpass API CLI in the library.

### CLI Shape

```bash
$ overpass-pp-cli --help
Subject CLI — Find things worth photographing by name instead of by OpenStreetMap tag, with automatic failover across Overpass mirrors.

  • near   Finds things worth photographing near a place — water towers, lighthouses, brutalist buildings, piers, viewpoints — by name rather than by OpenStreetMap tag.
  • types   Lists every subject type the CLI knows how to find, with the OpenStreetMap tags behind each one.
  • route   Finds subjects inside a corridor between two places, so a drive can be planned around what is worth stopping for.
  • geojson   Writes results as GeoJSON for opening in a map, a Leaflet page, or any GIS tool.
  • mirrors   Shows which Overpass mirrors are currently answering and how many rate-limit slots each has free.

Health check: run 'overpass-pp-cli doctor' to verify auth and connectivity.

Usage:
  overpass-pp-cli [command]

Available Commands:
  export         Export data to JSONL or JSON for backup, migration, or analysis
  geocoding      Resolve a place name to coordinates via Nominatim, optionally restricted by country
  geojson        Writes results as GeoJSON for opening in a map or GIS tool
  mirrors        Shows which Overpass mirrors are answering and how many rate-limit slots each has free
  near           Finds things worth photographing near a place, by name rather than by OpenStreetMap tag
  route          Finds subjects inside a corridor between two places, for planning what to stop for
  search         Full-text search across synced data or live API
  types          Lists every subject type this CLI can find, with the OpenStreetMap tags behind each
```

### Novel Commands

| `near` | Find photographic subjects near a place | Finds things worth photographing near a place — water towers, lighthouses, brutalist buildings, piers, viewpoints — by name rather than by OpenStreetMap tag. |
| `types` | Browse the subject taxonomy | Lists every subject type the CLI knows how to find, with the OpenStreetMap tags behind each one. |
| `route` | Subjects along a drive | Finds subjects inside a corridor between two places, so a drive can be planned around what is worth stopping for. |
| `geojson` | Export to GeoJSON | Writes results as GeoJSON for opening in a map, a Leaflet page, or any GIS tool. |
| `mirrors` | Mirror health | Shows which Overpass mirrors are currently answering and how many rate-limit slots each has free. |

### What This CLI Does

Overpass is a query language wearing a REST costume, and its public mirrors fall over independently under load. This maps photographic vocabulary — water towers, lighthouses, brutalist buildings, piers, observatories — onto the right tag combinations across 18 subject types in 9 groups, builds valid Overpass QL, retries across mirrors, caches results locally, and exports GeoJSON.

Two deliberate honesty features are worth calling out, because they shape what the tool promises:

- **`types` prints its own coverage limits.** `art_deco`, `brutalist` and `mural` each carry a note that OpenStreetMap style tagging is sparse and volunteer-contributed, so results undercount reality. The taxonomy table is printed verbatim rather than hidden behind the query builder.
- **`route` states its own approximation.** It searches a rectangle around the straight line between two places, not a buffer around the real road, and reports each result's off-line distance so a detour can be judged before committing to it.

Queries cover nodes, ways, and relations. Querying only nodes is the common Overpass mistake and silently loses most large structures.

### Manuscripts

- Research brief: `library/travel/overpass/.manuscripts/20260726-090225/research/`
- Proofs: `library/travel/overpass/.manuscripts/20260726-090225/proofs/`

### Validation Results

| Check | Result |
|-------|--------|
| Manifest | PASS |
| Transcendence | PASS |
| Phase 5 | PASS (16/16, quick) |
| go mod tidy | PASS |
| govulncheck (this CLI only, reachable findings) | PASS — no vulnerabilities |
| go vet | PASS |
| go build | PASS |
| --help | PASS |
| --version | PASS |
| verify-skill | PASS |
| Manuscripts | PRESENT |

`shipcheck` also passes 7/7 legs (verify, validate-narrative, dogfood, workflow-verify, apify-audit, verify-skill, scorecard) at **85/100, Grade A**.

### Notes for review

- **No auth.** Overpass and Nominatim are both keyless; the CLI sends a descriptive User-Agent and paces itself (1 rps to Nominatim, adaptive limiting with typed rate-limit errors to Overpass) rather than hammering a host.
- **Mirror failover is the core reliability feature**, and it is real: during pre-publish testing only 1 of 3 public mirrors was healthy, and the CLI failed over correctly. `mirrors` reports per-mirror state and free rate-limit slots.
- `spots.geojson` at the CLI root is a small sample export (8 public OpenStreetMap lighthouse records) produced by the documented `geojson --out` example. Happy to drop it if the library prefers no sample outputs.


tmchow:

justinwfu:
Both review findings are real and are fixed in ffdad7b.

**Issue 1 — partial warnings corrupt JSON output.** Confirmed. `runQuery` wrote the truncation note to stdout, ahead of the document `near --json`, `route --json` and `geojson` then emit. `runQuery` now *returns* the remark instead of printing it:

- `near --json` / `route --json` carry `"partial": <bool>` and, when truncated, `"partial_remark": "<Overpass's own text>"`.
- `geojson` attaches the same two keys to the FeatureCollection as foreign members (RFC 7946 §6.1). That also fixes a second-order problem the stdout line never solved: a `--out` export loses the stderr warning the moment the file is reopened, and a truncated export is otherwise indistinguishable from a complete one on the map.
- The human-readable line still goes to stderr in every mode, and to stdout only where stdout was already prose (the `near` / `route` tables, and the `geojson --out` confirmation).

**Issue 2 — failover discards caller cancellation.** Confirmed. `context.WithoutCancel` is deliberate — it drops the caller's per-request *deadline* so one slow mirror cannot consume the whole failover budget — but it dropped cancellation with it. The command's root context carries the interrupt and not the deadline, so it is re-attached with `context.AfterFunc`, guarded for the nil `cmd.Context()` that Cobra leaves before `ExecuteC`.

**Verification.** Four tests added in `internal/cli/subject_partial_test.go`, each watched to fail against the old behavior before being kept:

| planted failure | result |
|---|---|
| restore the stdout write | `near --json` stdout → `invalid character 'o' in literal null`; GeoJSON test likewise |
| remove the `AfterFunc` | cancellation takes **5.00s** (the full mirror budget) instead of **0.15s** |

A fourth test pins the other edge — a complete answer must not grow `partial_remark` and must not warn.

One gap stated plainly: `route`'s JSON path takes the same four lines but is **not** covered by these tests. It geocodes through Nominatim before reaching `runQuery`, and the geocoder's base URL is a package constant the tests cannot stub.

Full module `go build` / `go vet` / `go test ./...` are green, `gofmt` is clean, and `verify_skill.py --dir library/travel/overpass --strict` and `verify_manifest.py` (320 manifests, 0 failed) both pass locally. README and SKILL.md document the new `partial` contract.

justinwfu:
@greptileai review

justinwfu:
The policy gate on `ffdad7b` is stale — it lost a race, not a score. Could a maintainer re-run it? I don't have the rights to dispatch or re-run it myself.

Sequence:

- `07:00:09` — `Greptile Review` reported `completed / neutral` (the auto-review declining, this PR is over the size cap).
- `07:00:10` — the gate read the latest Greptile comment, found it still referencing the previous head, and exited: `Latest Greptile confidence-score comment does not reference the current head SHA.`
- A manual `@greptileai review` then ran for 2m25s and **edited that same comment in place** ([#issuecomment-5111680816](https://github.com/mvanhorn/printing-press-library/pull/1612#issuecomment-5111680816)).

That comment now reads **Confidence Score: 5/5**, "The PR appears safe to merge", against `Last reviewed commit: ffdad7bbcc83fef6001b3383ce2dc84e2b9305d7` — the current head. So the gate's own condition is satisfied now; it just evaluated about two minutes early.

Worth noting for the gate itself: because Greptile edits its existing comment rather than posting a new one, the comment's `created_at` stays old while its body updates. A gate that exits the moment the `Greptile Review` check turns `completed` will keep racing the body update, and `neutral` (auto-review declined on size) is not the same signal as `pass`.

Every other check on this head is green: Verify, Govulncheck, Scan, Validate MCPB manifest contract, Validate SKILL.md against shipped CLI source, and Greptile Review.

tmchow:
You were right about the gate. It raced Greptile's in-place comment edit, not the score. Greptile is 5/5 on `ffdad7b`. I'll re-run the gate.

Live dogfood is still the hold. There's a quick 16-pass Phase 5 marker, no `dogfood-results.json`, and no `publish-live-gate.json`. Overpass is public, so a live run with command counts is the remaining ask.

tmchow:
Still waiting on a live Overpass Phase 5 / publish-live-gate run with command counts. Greptile side looks fine; that live proof is the remaining ask. Ping when it is on the branch.

Issue #1635, opened 2026-07-31
ci: 47 published CLIs fail go build and are invisible to govulncheck (redundant toolchain directive)


MBemera:
## Summary

47 published CLIs fail `go build ./...`, `go vet ./...` and `govulncheck ./...` on `main`:

```
go: updates to go.mod needed; to update it:
	go mod tidy
```

## Cause

Their `go.mod` declares both `go 1.26.5` and `toolchain go1.26.5`. A toolchain line equal to the go directive is redundant and tidy wants it removed. `go build` runs `-mod=readonly` by default, so it refuses rather than proceeding. `go mod tidy -diff` shows the entire delta is that one deleted line.

Introduced by `434f4d45b` (`fix(catalog): require Go 1.26.5 across published modules`), which bumped `go 1.26` -> `go 1.26.5` and `toolchain go1.26.4` -> `toolchain go1.26.5`. Before that bump the pair was valid, because the toolchain was newer than the go directive.

## Impact

- `govulncheck` cannot scan these 47 modules at all. It fails to load the package graph, so a reachable vulnerability in any of them goes unreported. This is the part that matters.
- CI does not catch it, because govulncheck is scoped to CLIs a PR touches and nothing has touched these.
- End users are unaffected. `go install <module>/cmd/<cli>@latest` works, verified on `library/ai/context-dev`.

## Fix

Delete the `toolchain` line, or run `go mod tidy`, in each affected module. One line each.

## Affected modules

```
library/ai/context-dev
library/ai/mixlayer
library/auth/1password
library/cloud/cloudflare
library/cloud/vercel-admin
library/commerce/autotempest
library/commerce/daraz
library/commerce/grubhub
library/commerce/shipsgo
library/developer-tools/azure-devops
library/developer-tools/claude-agent-sdk-python-docs
library/developer-tools/github-intel
library/developer-tools/isitagentready
library/devices/gl-inet
library/devices/withings
library/health/oura
library/marketing/amplitude
library/marketing/braze
library/marketing/google-ad-manager
library/marketing/judge-me
library/marketing/kdpnichefinder
library/media-and-entertainment/boardgamegeek
library/media-and-entertainment/metacritic
library/media-and-entertainment/paul-graham
library/media-and-entertainment/podcastindex
library/other/air-quality
library/other/fred
library/other/intervals-icu
library/other/lda-gov
library/other/nhc
library/other/policy-intel
library/other/shelters
library/other/ufo-goat
[...]

PR #1636, opened 2026-07-31
fix(cli): bump golang.org/x/text to v0.39.0 in affected published CLIs

MBemera:
## Bump `golang.org/x/text` to v0.39.0 in affected published CLIs

74 published CLIs carried `golang.org/x/text` below v0.39.0, exposing [GO-2026-5970](https://pkg.go.dev/vuln/GO-2026-5970) / [CVE-2026-56852](https://nvd.nist.gov/vuln/detail/CVE-2026-56852) (CVSS 3.1 **7.5 HIGH**): `norm.Iter` enters an infinite loop on invalid UTF-8 input.

This advisory is reachable. `unicode/norm` enters the package graph via `github.com/enetx/g`, `golang.org/x/net/idna` and `golang.org/x/text/cases`. These CLIs scrape untrusted HTML, so malformed UTF-8 is routine input, not adversarial.

### Versions replaced

| from | count |
|---|---|
| v0.37.0 | 66 |
| v0.38.0 | 5 |
| v0.36.0 | 1 |
| v0.34.0 | 1 |
| v0.14.0 | 1 |

`go get golang.org/x/text@v0.39.0 && go mod tidy` per module. Only go.mod/go.sum change. `table-reservation-goat` also picks up `golang.org/x/sync` v0.20.0 -> v0.21.0, required by x/text v0.39.0.

### Verification

| Check | Result |
|---|---|
| Modules below v0.39.0 remaining | 0 (76 carry x/text: 75 at v0.39.0, 1 at v0.40.0) |
| Bumped modules build / vet / govulncheck | 74/74 PASS / PASS / CLEAN |
| govulncheck findings library-wide | 0 |
| Checksum vs sum.golang.org | matches, one identical hash across all 75 go.sum files |

The floor figure comes from an actual scan of every module in `library/`, not an assertion.

Fix confirmed against the upstream regression input from [CL 794100](https://go-review.googlesource.com/c/text/+/794100): `norm.Iter` over `"\xf3\xcc\x80"` hangs on v0.37.0 and completes on v0.39.0. Only the streaming iterator path is affected; `norm.NFC.String` does not hang.

### Second commit: fix(suno)

`suno` carried `golang.org/x/net` v0.50.0 while the rest of the library sits at v0.55.0 or above. None of v0.50.0's 9 advisories are reachable in suno, so that commit restores the library-wide floor rather than fixing an exploitable path. Kept separate so it can be dropped independently.

### Not in this PR

47 published CLIs fail `go build` on `main` with `go: updates to go.mod needed`, caused by a redundant `toolchain` directive. That blocks govulncheck from scanning them entirely. Filed as #1635.


tmchow:

PR #1664, opened 2026-08-06
feat(google-analytics): add gated GA4 Admin write surface

bobeglz:
## What

`google-analytics-pp-cli` was read-only by construction: every command wrapped the Data API, auth assumed a service-account key, and the help text pinned `analytics.readonly`. There was no way to create a key event, a custom dimension, or a data stream — the GA4 Admin API surface was simply absent.

This adds a **gated** Admin API write surface.

## Changes

**Auth** — `ServiceAccountKey` becomes `Credentials` (type alias keeps the old name working). `Kind()` auto-detects service-account vs ADC `authorized_user`, and `MintToken` grew a refresh-token path. Resolution order: `--credentials` → `GOOGLE_ANALYTICS_ADC` → `GOOGLE_APPLICATION_CREDENTIALS`, with `~/` expansion. No credential path is hard-coded.

**Commands** — `key-events`, `custom-dimensions`, `custom-metrics`, `data-streams`, all with `--agent` JSON. Reads use `analytics.readonly`; writes use `analytics.edit`; clients are cached per scope.

**Escape hatch** — `admin <GET|POST|PATCH|PUT|DELETE> <path>` with `--body`/`@file`, `--update-mask`, `--api v1beta|v1alpha`, `--no-paginate`. `--api` is validated up front so a typo cannot silently fall back to v1beta.

**Write gate** — `requireExplicitYes` checks `cmd.Flags().Changed("yes")`, not the struct field, so `--agent` alone never authorizes a destructive call. Same pattern as `linear-pp-cli`.

```
$ google-analytics-pp-cli key-events delete --property N --name ... --agent
Error: key-events delete is destructive and irreversible: re-run with an explicit --yes (--agent alone does not authorize it)
```

## Two traps encoded in the docs

- **`accessBindings` lives in `v1alpha`, not `v1beta`.** v1beta returns a 404 *HTML* page, not JSON — easy to misread as a bad path.
- **Probing write permission must collide, never invent.** A `create` with a fresh name succeeds and leaves real data behind. Send a `create` that duplicates an existing resource and expect `409 ALREADY_EXISTS`. Found the hard way: a probe with a novel name created three stray key events on live properties.

## Verification

`go build`, `go vet`, `go test` (37 pass), `govulncheck`, `go mod tidy` all green. Admin commands verified live against a real GA4 property: `keyEvents` create/delete, `customDimensions` list, `accessBindings` via v1alpha. The write gate was smoke-tested against the built binary in both directions.

One defect fixed in the new tests: the gate test would have attempted a live DELETE if credentials happened to be exported, so it now clears the credential env vars first.

Docs were also quoted correctly, which takes the repo's own verifier from 6 errors on `main` to all checks passing.

## Not included

`publish validate` has three pre-existing failures — manifest `schema_version` 1-vs-2, a `phase5` marker at schema_version 0, and SKILL.md install-section drift. All three reproduce identically on a pristine `upstream/main`; they are artifacts of the original 4.20.1 print, and the verifier states the install text is generator-owned and must not be hand-edited. The fix is a reprint, which should run after this merges so the write surface isn't overwritten.

### DIFF
--- a/library/marketing/google-analytics/.printing-press-patches/ga4-admin-write-surface.json
+++ b/library/marketing/google-analytics/.printing-press-patches/ga4-admin-write-surface.json
@@ -1,0 +1,30 @@
+{
+  "schema_version": 2,
+  "id": "ga4-admin-write-surface",
+  "applied_at": "2026-08-06",
+  "base_run_id": "ga4-20260612",
+  "base_printing_press_version": "4.20.1",
+  "summary": "feat(google-analytics): add gated GA4 Admin write surface; accept ADC authorized_user credentials",
+  "reason": "The printed CLI was read-only by construction: MintToken hard-coded analytics.readonly and --credentials assumed a service-account key, so ADC authorized_user credentials were rejected and there was no path to create, patch, archive, or delete anything in GA4. Two throwaway Python scripts were filling the gap; this absorbs their surface into the CLI.",
+  "files": [
+    "internal/ga4/types.go",
+    "internal/ga4/auth.go",
+    "internal/ga4/client.go",
+    "internal/ga4/admin.go",
+    "internal/ga4/admin_test.go",
+    "internal/cli/root.go",
+    "internal/cli/health.go",
+    "internal/cli/confirm.go",
+    "internal/cli/admin_write_commands.go",
+    "internal/cli/admin_write_commands_test.go",
+    "internal/cli/agent_context.go",
+    "README.md",
+    "SKILL.md",
+    "AGENTS.md",
+    "CHANGELOG.md",
+    ".printing-press.json"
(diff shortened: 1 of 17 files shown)


tmchow:

- Verify concluded FAILURE

tmchow:
Still the Aug 7 blockers: Verify and the Greptile policy gate are both red on `13af30ce`.

If nothing lands in the next couple of days I'll close this and you can open a fresh PR once those are green.

tmchow:
Closing as stale. The Aug 7 ask still stands: Verify and the Greptile policy gate are both red on `13af30ce`, and nothing has landed since.

Please open a fresh PR once those are green.

PR #1666, opened 2026-08-06
feat(table-reservation-goat): Tock confirm-path forensics instrumentation

teebs4140:
Companion to #1554: when a Tock booking fails between the confirm click and the receipt (`chromedp_book_failed`, receipt never reached), the current failure report says nothing about WHY. This instruments the confirm path — request/response capture armed before navigation, dialog/DOM state snapshots at the failure boundary — so the next silent confirm failure carries actionable evidence instead of requiring a live repro.

- `chrome_book.go`: capture plumbing (armed pre-navigation; day-free availability wrappers ignored), attached to the existing confirm/checkout state machine; no behavior change on the happy path.
- Patch note in `.printing-press-patches/`.
- Tests cover capture arming, failure-boundary snapshots, and no-op on successful confirms.

Motivated by a production booking (2026-08-02) that failed silently at confirm; the instrumentation shipped in our fork on 2026-08-05.

Test note: `TestClickPinnedSlotByTimeText_ExplicitCardInsulatesFromForeignFormLink` is flaky on current `main` in Chrome 150 environments (4/5 failures WITHOUT this change) — that's the pre-existing fixture race fixed by #1659; with #1659 applied this branch passes it 5/5, and the rest of the tock + cli suites are green here.

https://claude.ai/code/session_017ZgvVfyPxcLHYy5DPoDf37

### DIFF
--- a/library/food-and-dining/table-reservation-goat/.printing-press-patches/tock-confirm-forensics.json
+++ b/library/food-and-dining/table-reservation-goat/.printing-press-patches/tock-confirm-forensics.json
@@ -1,0 +1,14 @@
+{
+  "schema_version": 2,
+  "id": "tock-confirm-forensics",
+  "applied_at": "2026-08-05",
+  "base_run_id": "20260712-223206",
+  "base_printing_press_version": "4.28.0",
+  "summary": "Preserve privacy-safe Tock confirm-path forensics across trusted click delivery, redirect-generation-aware network outcomes, retry recovery, target validity, alerts, and payment iframe presence without adding another submission attempt.",
+  "reason": "A Tock checkout can remain on confirm-purchase without revealing whether the trusted click reached the tagged control, whether the form POST was emitted, or whether the one allowed retry dispatched. A reprint must retain target-session Network instrumentation, canonical failure evidence, and the existing single-retry and receipt-deadline safety ceilings.",
+  "files": [
+    "internal/source/tock/chrome_book.go",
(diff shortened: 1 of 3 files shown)


teebs4140:
Addressed the review finding in b1758ef01: the loading-failure flag is now generation-aware (a responding retry clears an initial failure; a retry that dies on the wire records the newest attempt's failure even after an initial response). Two unit tests added covering both directions.

teebs4140:
Addressed in the latest push: a loading-failed newer generation now also zeroes ConfirmPostLastStatus (and advances the status generation), so the snapshot never pairs an older attempt's status with a newer attempt's failure. Test updated to pin 2/0/true for the response-then-retry-failure case.

tmchow:

- Govulncheck concluded CANCELLED

teebs4140:
@tmchow The needs-author state here is stale: the Govulncheck cancellation was resolved and all five checks are green on the current head (`f64b8e0a`) — Govulncheck, Verify, Greptile Review, Greptile policy gate, and the SKILL.md validator all pass. No changes since the review threads were addressed. Requesting re-triage / maintainer routing.

tmchow:
The Aug 7 needs-author note was a cancelled govulncheck run, not a code defect. Current head has Govulncheck, Verify, SKILL, Greptile, and the policy gate all green. Instrumentation only; no happy-path behavior change. Routing to maintainer.

PR #1672, opened 2026-08-07
feat(opensky): add opensky-pp-cli

veltri-23:
## opensky

A live aircraft tracker for the OpenSky Network, the volunteer ADS-B receiver grid that publishes open flight data. No account, no key, no browser. Point it at a bounding box and it tells you what is flying over it right now, then lets you pull the flights that left a specific airport in the last hour or replay one aircraft's full trajectory.

**API:** OpenSky Network v1 (opensky-network.org/api, no auth) | **Category:** travel | **Press version:** 4.28.0

**Spec:** hand-written OpenAPI 3.0 in `spec.yaml`, generated via `cli-printing-press generate` (run id `20260807-162902`)

### CLI Shape

```
Manage opensky resources via the opensky API.

Add --agent to any command for JSON output + non-interactive mode.
Run 'opensky-pp-cli doctor' to verify auth and connectivity.

Usage:
  opensky-pp-cli [command]

Available Commands:
  search         Search locally synced data
  states         Live aircraft states over a bounding box or for one transponder
  version        Print version

Flags:
  -h, --help                 help for opensky-pp-cli
      --no-learn             Disable the teach/recall learning loop for this invocation
      --profile string       Apply values from a saved profile (see 'opensky-pp-cli profile list')
  -v, --version              version for opensky-pp-cli

Use "opensky-pp-cli [command] --help" for more information about a command.
```

### What This CLI Does

- `states` streams live state vectors for any bounding box (`--lamin --lomin --lamax --lomax`) or a single transponder (`--icao24`). Verified live over the JFK metro box: ~40 aircraft with callsign, position, velocity, and altitude, UAL2679 and friends included.
- `flights get-departures` and `flights get-aircraft-trajectory` cover the time-window endpoints. Verified live: 23 departures out of KJFK in the last hour with zero epoch math, and a full trajectory replay for a real transponder (aa07f9, UAL3024 out of EWR).
- The usual printed-CLI extras are all here: `sync` + `search` for an offline SQLite copy, `doctor`, `which`, the teach/recall learning loop, and the MCP server (2 command-mirror tools, runtime Cobra-tree parity confirmed by the MCP surface check).

Two deliberate divergences from a naive spec, recorded as patches:

1. OpenSky requires `begin`/`end` on every `/flights/*` call and anonymous access caps the window at 1 hour. The generated help text already promised defaults ("defaults to now-2h") but the code sent nothing, so bare calls 400'd. Patch 001 makes omitted windows default to the last hour, so `flights get-departures --airport KJFK` just works.
2. OpenSky answers HTTP 404 with an empty array when a window has no flights. The template's generic 404 handler turned that into "resource not found, run the list command", which is wrong and confusing. Patch 002 normalizes it to an empty result with exit 0.

Also worth knowing: `/flights/arrival` and `/routes` are not in the spec. Both were probed live and are currently dead upstream (`/flights/arrival` returns 404-with-empty-array for every airport and window tried, `/routes` returns 400 for every code format). Shipping commands that fail on every live call felt worse than shipping without them; the research notes carry the probe evidence and either endpoint can be added back when upstream recovers.

### Manuscripts

- Research: `library/travel/opensky/.manuscripts/20260807-162902/research/notes.md`, `research.json`
- Proofs: `library/travel/opensky/.manuscripts/20260807-162902/proofs/phase5-acceptance.json`, `phase5-skip.json`, `smoke-test.json`

### Validation Results

| Check | Result |
|---|---|
| Manifest (printer, run_id, novel_features, search_terms) | PASS |
| Phase 5 (dogfood, quick: 25/25 structural checks) | PASS |
| shipcheck (all 7 legs: verify, validate-narrative, dogfood, workflow-verify, apify-audit, verify-skill, scorecard) | PASS |
| go mod tidy | PASS |
| go vet ./... | PASS |
| go build ./... | PASS |
| go test ./... (12 packages, Windows) | PASS |
| --help / --version / doctor | PASS |
| verify-skill | PASS |
| govulncheck | PASS |
| Manuscripts present | PASS |

The Windows note: the generated test helpers isolated `HOME` only, which is a no-op on Windows (`os.UserHomeDir()` reads `USERPROFILE`), so the suite leaked into the real user dirs and failed with entries from actual CLI use. Patch 003 extends the five helpers with `USERPROFILE`/`HOMEDRIVE`/`HOMEPATH` isolation. Test-only change, zero behavior change on POSIX, same fix as the greenhouse and lever prints.

### Publication Path

`library/travel/opensky/` only, per the publishing rules. `registry.json`, `cli-skills/`, README catalog cells, and release ledger files are untouched (bot-regenerated post-merge).

### Novel Commands

- `states` — live ADS-B state vectors over any bounding box (`--lamin --lomin --lamax --lomax`), no auth
- `flights get-departures` — last hour of departures with no epoch math (`--airport` required, `--begin`/`--end` optional)
- `flights get-aircraft-trajectory` — full trajectory replay for one aircraft (`--icao24`)

### Gaps

- `--data-source local` reads work after `sync`, but `/flights/*` rows carry no stable ID so offline lookups against them are incomplete (the store warns about this on sync). Live queries are unaffected.
- Anonymous rate limits apply (roughly a handful of requests per minute on the time-window endpoints). The spec is honest about this; a future auth-token option would lift it.
- State-vector rows come back as positional arrays, so tables show raw columns; `--select` and `--json` give the structured view.

Post-push fixes folded into this PR (patches 004 and 005): the module path was canonicalized to the directory form the CI gate requires, and the `which` index entries are flagless (the generator renders them from `novel_features`, and command strings with example flags do not resolve through cobra's `Find`).


tmchow:

veltri-23:
@tmchow thank you for the merge! :)

PR #1693, opened 2026-08-11
feat(linear): sync prunes upstream deletions and supports incremental crawls

lagiosv:
## Gaps closed

- **GAP-003** Sync never prunes: rows deleted or archived upstream live forever in the local snapshot. (MUST-SHIP 2)
- **GAP-019** No incremental sync: cursors are written empty and only for issues, and the documented `sync --incremental` flag does not exist.
- **GAP-020** Default `--max-pages 10` silently truncates the snapshot at 500 issues and 250 projects with no partial-result signal.
- **GAP-021** `includeArchived` is never passed on any query, so no command can see archived entities.
- **GAP-035** Prune existed only as an external script and only for issues: the other six typed tables drift with no reconciler at all.

## Design

Prune is a live-id-set reconciler, not a timestamp heuristic. Two guardrails from the external `linear-store-prune` script are carried over verbatim because they are load-bearing: an empty live set refuses to prune rather than wiping the table, and the FTS shadow table is reconciled in the same transaction so search can never outlive its rows. Prune only runs off a pass that actually enumerated the workspace. A pass that stopped early on `--max-pages`, or an incremental pass that fetched only what changed, reports no live id set and is refused, which is the second lock on the same door: handing a reconciler the ids that *changed* instead of the ids that *exist* would delete every unchanged row. Incremental crawls narrow through `IssueFilter.updatedAt`, a `DateComparator` verified live against the introspected schema, with a per-table cursor persisted in `sync_state`. Cursor writes moved into the sync orchestration loop so every resource gets one instead of only issues, which is why the `issues search` self-refresh now stamps its own pass: it bypasses that loop.

## Verification (MUST-SHIP 2 proof)

Archive-then-sync against a real workspace:

```
$ linear-pp-cli issues create --team ACC --title "pp-probe prune E" --trust-mode strict --agent
ACC-260, id 31ecdb25-...
$ linear-pp-cli sync --full --max-pages 0
Syncing issues... 229
$ linear-pp-cli sql "SELECT count(*) FROM issues WHERE identifier='ACC-260'"   -> 1
$ linear-pp-cli sql "... issues, issues_fts"                                   -> issues 229, fts 229

$ linear-pp-cli issues archive ACC-260 --yes --agent
archivedAt 2026-08-11T19:22:39.099Z
$ linear-pp-cli sync --full --max-pages 0
Syncing issues... 228
sync: pruned 1 stale issues row(s)
Pruned 1 stale rows: issues 1

$ linear-pp-cli sql "SELECT count(*) FROM issues WHERE identifier='ACC-260'"   -> 0
$ linear-pp-cli sql "... issues, issues_fts"                                   -> issues 228, fts 228
```

The row is gone and the FTS shadow table stayed in lockstep, so the delete trigger fires on prune.

The empty-set refusal fired for real during the same run rather than being asserted in a unit test only:

```
Syncing labels... 6
sync: pruned 3 stale labels row(s)
Syncing cycles... 0
sync: cycles returned an empty live set, prune skipped
Syncing issues... 224
sync: pruned 7 stale issues row(s)
Pruned 10 stale rows: labels 3, issues 7
```

Incremental pass, cursor honored and prune correctly skipped:

```
Syncing cycles since 2026-08-11T19:47:36Z... 0
Syncing issues since 2026-08-11T19:47:36Z... 3
Synced 92 items in 2.875s
Incremental (updatedAt window, prune skipped): cycles, issues
```

`--include-archived` reaches rows that were previously invisible:

```
without --include-archived : 226 rows, ACC-263 absent
with    --include-archived : 263 rows, ACC-263 present with archivedAt 2026-08-11T19:55:24.821Z
```

Build, vet and the full test suite are green.

---

Stacked on #1692. Review the diff against `feat/linear-state-groups`.


tmchow:

PR #1725, opened 2026-08-15
feat(ynab): add ynab

mikesnowbie:
## ynab

Every YNAB feature, plus a local mirror that makes reconciliation, payee trend analysis, and a ready-made ProjectionLab export instant instead of manual.

**API:** ynab | **Category:** other | **Press version:** 4.30.2
**Spec:** https://api.ynab.com/papi/open_api_spec.yaml (official, v1.86.0)

### Publication Path

Update existing PR #1725

This update addresses Greptile's outstanding review finding (confidence 4/5, "not yet safe to merge"): `accounts reconcile`'s candidate-transaction search omitted `since_date`, so YNAB's API silently capped it to a one-year lookback while the reported discrepancy was computed from the account's real all-time cleared balance — a culprit transaction older than a year could never appear in candidates. `accounts reconcile` now defaults `since_date` to `2000-01-01` (effectively full account history) whenever `--since-date` is not supplied. It also bumps `go.mod` from `go 1.26.5` to `go 1.26.6` to match the library's current toolchain floor, which was blocking the Scan check independently of the reconcile fix.

### CLI Shape

```bash
$ ynab-pp-cli --help
Ynab CLI — Every YNAB feature, plus a local mirror that makes reconciliation, payee trend analysis, and a ready-made ProjectionLab export instant instead of manual.

  • export balances   One-shot export of current account balances reshaped into ProjectionLab's expected schema, ready to sync into a net-worth projection tool.
  • accounts reconcile   Diff the local cleared-transaction total against a bank statement balance and surface the transaction(s) most likely responsible for the gap.
  • payees profile   Aggregated monthly spend stats, average transaction size, and typical categories for a single payee.

Health check: run 'ynab-pp-cli doctor' to verify auth and connectivity.

Usage:
  ynab-pp-cli [command]

Available Commands:
  accounts               The accounts for a plan.
  agent-context          Emit structured JSON describing this CLI for agents
  api                    Browse all API endpoints by interface name
  auth                   Manage authentication for Ynab Endpoints
  categories             The categories for a plan, organized into category groups.
  category-groups        Create and update category groups
  completion             Generate the autocompletion script for the specified shell
  doctor                 Check CLI health
  export                 export subcommands: balances
  feedback               Record feedback about this CLI (local by default; upstream opt-in)
  help                   Help about any command
  learnings              Inspect or forget the local search_learnings table
  money-movement-groups  Returns all money movement groups
  money-movements        Returns all money movements
  months                 Each plan contains one or more months, which is where Ready to Assign
  payee-locations        When you enter a transaction and specify a payee on the YNAB mobile apps
  payees                 The payees for a plan.
  plans                  Plans are the top-level container for your YNAB data; accounts, categories, payees
  playbook               Inspect or amend stored CLI playbooks
  profile                Named sets of flags saved for reuse
  recall                 Check prior learnings for a query before running discovery (LLM-fired, pre-discovery)
  scheduled-transactions The scheduled transactions for a plan
  settings               Returns settings for a plan
  teach                  Record a query -> resource mapping for future recall (LLM-fired, silent)
  teach-lookup           Install a manual entity-lookup row (kind, canonical, value)
  teach-pattern          Install a manual generalization pattern (query_template, resource_template, entity_kind)
  teach-playbook         Record a CLI playbook + free-text notes for a query family
  transactions           The transactions for a plan. Transaction amounts are specified in [milliunits format](https://api.ynab.com/#formats).
  user                   Returns authenticated user information
  version                Print version
  which                  Find the command that implements a capability
  workflow               Compound workflows that combine multiple API operations

Flags:
      --agent                   Set all agent-friendly defaults (--json --compact --no-input --no-color --yes)
      --allow-partial-failure   Downgrade response-body partial-failure (e.g. partialFailureError) to a warning instead of a non-zero exit
      --audit-dir string        Aggregate the receipt and index under this audit directory
      --client-profile string   Select the tenant-gated client profile (env: PRINTING_PRESS_CLIENT_PROFILE)
      --compact                 Return only key fields (id, name, status, timestamps) for minimal token usage
      --config string           Config file path
      --csv                     Output as CSV (table and array responses)
      --data-source string      Data source for read commands: auto (live with local fallback), live (API only), local (synced data only) (default "auto")
      --deliver string          Route output to a sink: stdout (default), file:<path>, webhook:<url>
      --dry-run                 Show request without sending
  -h, --help                    help for ynab-pp-cli
      --home string             Root directory for config, data, state, and cache files
      --human-friendly          Enable colored output and rich formatting
      --idempotent              Treat already-existing create results as a successful no-op
      --ignore-missing          Treat missing delete targets as a successful no-op
      --json                    Output as JSON
      --max-age duration        Maximum acceptable age of local-store data before a stderr hint suggests sync; 0 disables (default 30m0s)
      --no-cache                Bypass response cache
      --no-color                Disable colored output
      --no-input                Disable all interactive prompts (for CI/agents)
      --no-learn                Disable the teach/recall learning loop for this invocation
      --plain                   Output as plain tab-separated text
      --profile string          Apply values from a saved run profile; this does not select a client (see 'ynab-pp-cli profile list')
      --quiet                   Bare output, one value per line
      --rate-limit float        Max requests per second (0 to disable)
      --receipt                 Write an atomic private run receipt
      --receipt-file string     Override the run receipt destination
      --select string           Comma-separated fields to include in output (e.g. --select data)
      --timeout duration        Request timeout (default 1m0s)
  -v, --version                 version for ynab-pp-cli
      --yes                     Skip confirmation prompts (for agents and scripts)

Use "ynab-pp-cli [command] --help" for more information about a command.
```

### Novel Commands

| `export balances` | Export balances (ProjectionLab-shaped) | One-shot export of current account balances reshaped into ProjectionLab's expected schema, ready to sync into a net-worth projection tool. |
| `accounts reconcile` | Reconciliation helper | Diff the local cleared-transaction total against a bank statement balance and surface the transaction(s) most likely responsible for the gap. |
| `payees profile` | Payee spend profile | Aggregated monthly spend stats, average transaction size, and typical categories for a single payee. |

### What This CLI Does

This CLI mirrors your budget locally so bank reconciliation and payee spend history become instant, offline, agent-native commands instead of manual math against a live API and a 200-request-per-hour ceiling. It also ships a one-shot export shaped for downstream net-worth tools like ProjectionLab.

### Manuscripts

- [Research Brief](https://github.com/mvanhorn/printing-press-library/blob/582804e32013c1f19d437b0026e8010f5dbe6c89/library/other/ynab/.manuscripts/20260812-093755-e1ecbcb1/research/2026-08-12-093755-feat-ynab-pp-cli-brief.md)
- [Absorb Manifest](https://github.com/mvanhorn/printing-press-library/blob/582804e32013c1f19d437b0026e8010f5dbe6c89/library/other/ynab/.manuscripts/20260812-093755-e1ecbcb1/research/2026-08-12-093755-feat-ynab-pp-cli-absorb-manifest.md)
- [Novel Features Brainstorm](https://github.com/mvanhorn/printing-press-library/blob/582804e32013c1f19d437b0026e8010f5dbe6c89/library/other/ynab/.manuscripts/20260812-093755-e1ecbcb1/research/2026-08-12-093755-novel-features-brainstorm.md)
- [Source OpenAPI Spec](https://github.com/mvanhorn/printing-press-library/blob/582804e32013c1f19d437b0026e8010f5dbe6c89/library/other/ynab/.manuscripts/20260812-093755-e1ecbcb1/research/ynab-openapi-spec.yaml)
- [Shipcheck Report](https://github.com/mvanhorn/printing-press-library/blob/582804e32013c1f19d437b0026e8010f5dbe6c89/library/other/ynab/.manuscripts/20260812-093755-e1ecbcb1/proofs/2026-08-12-1435-fix-ynab-pp-cli-shipcheck.md)
- [Phase 5 Acceptance](https://github.com/mvanhorn/printing-press-library/blob/582804e32013c1f19d437b0026e8010f5dbe6c89/library/other/ynab/.manuscripts/20260812-093755-e1ecbcb1/proofs/phase5-acceptance.json)

### Validation Results

| Check | Result |
|-------|--------|
| Manifest | PASS |
| Phase 5 | PASS |
| go mod tidy | PASS |
| go vet | PASS |
| go build | PASS |
| --help | PASS |
| --version | PASS |
| Manuscripts | PRESENT |

### Publish Live Gate

Full live dogfood reran at publish time against the real YNAB API and passed (146/146 evaluated checks, remainder skipped as mutating/blocked-fixture by design), specifically re-verifying `accounts reconcile`'s corrected `--since-date` default. Proof: [Phase 5 Acceptance](https://github.com/mvanhorn/printing-press-library/blob/582804e32013c1f19d437b0026e8010f5dbe6c89/library/other/ynab/.manuscripts/20260812-093755-e1ecbcb1/proofs/phase5-acceptance.json).

### Hand-authored customizations

Four patches beyond generator output, recorded in `.printing-press-patches/`:

- `ynab-envelope-unwrap` — YNAB wraps every API response in its own `{"data": {...}}` envelope; this print's spec absorption never classified that wrapper, so generated read/write commands doubly nested output as `results.data.<field>` instead of `results.<field>`. Fixed at the display layer only (after local-cache writes, which already handled the wrapper correctly) via a new `unwrapYNABEnvelope` helper.
- `ynab-dead-code-and-annotations` — removed a verified-dead helper function and added missing `pp:typed-exit-codes` annotations to four parent commands to match sibling convention, applied during a `/printing-press-polish` pass.
- `ynab-novel-command-freshness` — the three novel financial commands bypass the client's 5-minute response cache on their live fetches (`GetWithHeadersNoCache`), since they're declared `pp:data-source live` and a cached response could silently show pre-change numbers.
- `ynab-reconcile-full-history-default` — `accounts reconcile` now defaults `since_date` to `2000-01-01` (effectively full account history) instead of leaving it unset, fixing the one-year-lookback gap described above.

### Known gap (not addressed in this PR)

This print has no `sync`/local-mirror command — no resource was classified as syncable during generation, so all novel commands hit the live API directly rather than reading a local SQLite mirror. An attempt to add `x-cache`/`x-pp-syncable` spec enrichment and regenerate via `generate --force` corrupted the tree (deleted hand-authored novel-command files and project metadata) and was reverted. A `/printing-press-reprint` is planned separately to address this properly.

### Gaps

- `spec_url` is unset in the manifest; the CLI was generated against the official YNAB OpenAPI spec at `https://api.ynab.com/papi/open_api_spec.yaml` (v1.86.0), linked above.


mikesnowbie:
### Govulncheck: known library-wide gap, not a defect in this CLI

This CLI's `govulncheck` check is expected to keep failing, and I want to flag why before it looks unaddressed.

The repo's `.go-version` pins CI to Go 1.26.5 with `GOTOOLCHAIN=local` (no automatic toolchain downloads). Go 1.26.6 fixed four standard-library CVEs, all reachable from this (and presumably most) generated CLI's ordinary `net/http` usage in `internal/cliutil/probe.go`'s reachability probe:

- [GO-2026-6218](https://pkg.go.dev/vuln/GO-2026-6218) — `net/url` quadratic complexity
- [GO-2026-6090](https://pkg.go.dev/vuln/GO-2026-6090) — `crypto/tls` post-handshake message limit
- [GO-2026-5972](https://pkg.go.dev/vuln/GO-2026-5972) — `encoding/asn1` recursion depth
- [GO-2026-5026](https://pkg.go.dev/vuln/GO-2026-5026) — `net/http`/`golang.org/x/net/idna` Punycode label rejection

I tried both states of `go.mod`:
- `go 1.26.6` → CI's `GOTOOLCHAIN=local` refuses to build at all (`go.mod requires go >= 1.26.6 (running go 1.26.5; GOTOOLCHAIN=local)`), failing before govulncheck even runs.
- `go 1.26.5` (current state of this PR) → the build proceeds, but govulncheck then correctly reports the four CVEs above, since it evaluates against the `go` directive's declared stdlib version regardless of the actual toolchain invoking it — confirmed by reproducing the identical finding set locally.

There's no third option from inside `library/other/ynab/`: the vulnerable code is standard-library internals bundled with the pinned Go 1.26.5 toolchain itself, reached through ordinary HTTPS client usage that any API-wrapping CLI needs. This isn't ynab-specific — any CLI in the catalog whose PR touches `go.mod` or Go source right now should hit the same wall, since the fix requires the actual Go binary CI compiles with to change.

I left `go.mod` at `go 1.26.5` (matching what CI can actually build) rather than `1.26.6` (which fails harder, before analysis even starts) — but neither state gets a clean govulncheck run today. The real fix is bumping this repo's root `.go-version` to `1.26.6+`, which is outside this PR's scope (`library/other/ynab/` only) and a maintainer call. Happy to open that as a separate PR/issue if useful — let me know.

tmchow (changes requested):
Reconcile `--since-date` is addressed. After the branch update, Scan is failing because `library/other/ynab/go.mod` still says `go 1.26.5` and the repo floor is now 1.26.6. Bump that one directive and this should be ready.

tmchow (changes requested):
Scan is red on this head. `library/other/ynab/go.mod` still has `go 1.26.5` and the repo floor is 1.26.6. Greptile is also 4/5: `accounts reconcile` still searches only one year of transactions by default, so an older discrepancy can be missing from the ranked results. Bump the go directive and fix that lookback.

mikesnowbie:
@tmchow both items from your last review are addressed in adb2abebc:

1. **go.mod floor** — `library/other/ynab/go.mod`'s `go` directive is now `1.26.6`, matching the repo floor. `go build ./...`, `go vet ./...`, and `go test ./...` all pass clean against that toolchain.
2. **Reconcile lookback (Greptile 4/5)** — `accounts reconcile` now defaults `since_date` to `2000-01-01` when `--since-date` isn't supplied, instead of omitting the parameter (which let YNAB silently apply its own one-year default). The candidate search now covers an account's practical full history by default; `--since-date` still narrows it when the caller wants that, and `result.Note`'s caveat now only fires in that narrowed case. Patch recorded at `.printing-press-patches/ynab-reconcile-full-history-default.json` / `.md`.

Re-requesting your review.

tmchow:
Both items from the Aug 17 review are on this head. `library/other/ynab/go.mod` is `go 1.26.6` and Scan is green. `accounts reconcile` now defaults `since_date` to 2000-01-01 instead of letting YNAB apply a one-year lookback. Greptile's remaining 4/5 is that a transaction older than 2000 would still be missed; that is not a real YNAB account. Live proof is pass / full / 146 of 146. Routing to maintainer.

tmchow (commented):
Both Aug 17 asks are met on this head: go.mod is 1.26.6 (Scan green) and reconcile defaults to a 2000-01-01 lookback. The leftover Greptile 4/5 about pre-2000 transactions is not a real YNAB case. Clearing the changes request so this can go to maintainer review.

tmchow (approved):
Approving to lift my own Aug 17 changes request. go.mod is 1.26.6 and the reconcile lookback is fixed. Merge still belongs to the maintainer queue.

PR #1758, opened 2026-08-19
feat(scientific-consensus): add DeepSeek provider, pin sampling temperature to 0

laci141:
## Summary

Adds DeepSeek as a sixth optional LLM stance provider, and pins `temperature: 0` for all OpenAI-compatible providers.

One file, `internal/scengine/llm_stance.go`, 29 insertions and 7 deletions. No new dependencies.

## Why DeepSeek

Stance classification issues **one API call per work**, so a 25-work run is 25 calls and a 300-work regression sweep is 300. Cost per call is therefore the dominant constraint on how much measurement is affordable.

DeepSeek is placed after `anthropic` and `openai` but before `gemini`/`groq`/`mistral`. A user who deliberately set an Anthropic or OpenAI key still gets what they asked for; a user with several keys configured gets the low-cost path by default.

The model name is a compile-time constant rather than an env var. A stance corpus measured under one model is not comparable to one measured under another, and silent model drift between runs would make any future regression unattributable.

## Why temperature: 0

This is the more consequential half of the change, and it affects the existing OpenAI, Groq and Mistral paths as well as DeepSeek, because it lives in the shared `openAICompatible` helper.

Without it, the same work and the same claim can receive different stance labels on different runs. In a consensus engine, reproducibility is the product: a user who reruns `consensus "does X cause Y"` and gets a different verdict has no way to tell whether the evidence changed or the sampler did.

Flagging this explicitly as a deliberate behaviour change to the three incumbent providers, not an accidental side effect.

## Measurement

I ran the LLM path against the heuristic across 13 stored corpora to check the provider works end to end before proposing it. DeepSeek `deepseek-chat`, temperature 0, 295 works, 5m22s wall time, 0 call errors.

| Probe | LLM | Heuristic |
| --- | --- | --- |
| Unrelated-claim control (correct answer is inconclusive) | 36/39 (92.3%) | 12/39 (30.8%) |
| Mean reported confidence | 0.883 | 0.692 |
| Label disagreement vs heuristic | 155/295 (52.5%) | n/a |

The unrelated-claim control is the one worth reading: works are scored against a claim they have nothing to do with, so the correct label is `inconclusive`. The heuristic answers from paper tone and gets it wrong most of the time; the LLM path abstains.

Two caveats I am not hiding:

- Mean confidence of 0.883 is higher than the heuristic's, and I have not yet established that it is calibrated rather than merely confident. That is separate work and not part of this PR.
- No hand-labelled gold set exists yet, so the table above measures behaviour and agreement, not ground-truth accuracy.

This PR does not change the default. The keyless heuristic remains the default path when no provider key is set, and `stance_method` still reports honestly which path produced a result.

## Verification

| Check | Result |
| --- | --- |
| `go build ./...` | exit 0 |
| `go vet ./...` | clean |
| `go test ./internal/scengine/ -count=1` | ok, 0.513s |
| `gofmt -l` on the LF copy of the file | clean |
| `git diff --stat` | 1 file changed, 29 insertions(+), 7 deletions(-) |

### DIFF
--- a/library/other/scientific-consensus/.printing-press-patches/deepseek-provider-and-temperature-zero.json
+++ b/library/other/scientific-consensus/.printing-press-patches/deepseek-provider-and-temperature-zero.json
@@ -1,0 +1,11 @@
+{
+  "schema_version": 2,
+  "id": "deepseek-provider-and-temperature-zero",
+  "applied_at": "2026-08-19",
+  "base_run_id": "20260620-165352-e55cec69",
+  "base_printing_press_version": "4.24.0",
+  "summary": "DeepSeek registered as an optional stance provider (DEEPSEEK_API_KEY, compile-time deepseek-chat) in the shared OpenAI-compatible helper, placed after Anthropic/OpenAI and before Gemini/Groq/Mistral; all OpenAI-compatible providers now send temperature 0 so stance labels are reproducible.",
+  "reason": "A reprint must keep DeepSeek in that priority slot (DEEPSEEK_API_KEY, compile-time deepseek-chat) and keep temperature 0 on the shared OpenAI-compatible helper — otherwise the optional low-cost path disappears and the same work/claim can receive different stance labels across runs.",
+  "files": ["internal/scengine/llm_stance.go"],
+  "validated_outcome": "Author measured DeepSeek deepseek-chat temperature 0 across 13 stored corpora / 295 works, 0 call errors, 5m22s; unrelated-claim control 36/39 vs heuristic 12/39. No new dependencies."
+}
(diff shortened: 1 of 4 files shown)


tmchow:
The DeepSeek path and `temperature: 0` look fine, and the 295-work run is enough testing for an existing-CLI change. What's missing is the `.printing-press-patches/` record, so a reprint doesn't wipe it. I'll add that.

ppl-triage: maintainer-fixing fda06d7507f4

tmchow:
Added the patch ledger record at `.printing-press-patches/deepseek-provider-and-temperature-zero.json`, and updated the existing README/SKILL provider lists with DeepSeek (`DEEPSEEK_API_KEY`), priority (after Anthropic/OpenAI, before Gemini/Groq/Mistral), and temperature 0. Waiting on CI/Greptile.

tmchow:
Patch ledger is in, CI is green, Greptile is 5/5. DeepSeek + temp 0 was already proven by the 295-work run. Merging.

PR #1759, opened 2026-08-19
feat(scientific-consensus): exclude retracted works from the consensus score

laci141:
## Summary

Excludes retracted works from the consensus score. Four files, 765 lines, of which about 600 are a stored API response used as test data. The engine code is roughly 150 lines.

## The problem

A retracted paper keeps its citation count, and citation count is what weights a work in the consensus aggregator. A study withdrawn for fabricated data can therefore outweigh several sound ones, purely because it was cited heavily both before and after retraction. Wakefield-class papers are the obvious case, but the mechanism is general.

## The approach

`DetectRetraction` reads two independent signals:

- a **title marker**, matched by a conservative anchored regex that only fires on `RETRACTED`/`WITHDRAWN` at the start of a title, followed by a short bracketed or colon-terminated span
- an **index flag**, supplied by the caller from source metadata

The regex is deliberately narrow. A paper *about* retraction ("Retraction rates in clinical trials, 2000–2020") must not be flagged, so the pattern is anchored to the start and bounded to 20 characters before the delimiter.

## Retraction is not a stance

This is the design decision worth reviewing. A retraction notice is not evidence against the claim the retracted paper discussed; it is a statement about a paper. Modelling it as a `refuting` stance would introduce a different error from the one being fixed: a single withdrawal would read as refuting evidence and move the consensus score in a direction the literature does not support.

`Retraction` is therefore its own type with `ExcludeFromScore()`, separate from the stance enum. `TestRetractionSignalIsNotAStance` guards this.

Excluded works are counted and reported separately, so a user sees that exclusion happened rather than silently receiving a smaller corpus.

## normalize.go

38 lines, one function. Folds typographic dashes and related punctuation before matching, so an en dash or em dash in a title does not defeat the marker regex. Included here because the retraction path is its first caller.

## Test data

`testdata/corpora_full/vaccines.json` is a stored OpenAlex response containing two genuinely retracted works. It is a real API capture rather than a hand-written fixture, so the regex is exercised against titles exactly as they arrive from the source, including their punctuation and bracketing.

This is the only corpus file in this PR. I have further stored corpora locally for regression work on the stance classifier, but those belong with that change, not this one.

## Tests

| Test | What it pins |
| --- | --- |
| `TestDetectRetraction` | title marker and index flag, positive and negative cases |
| `TestRetractionExcludedFromScoring` | a retracted work does not contribute weight |
| `TestRetractionSignalIsNotAStance` | retraction never becomes a stance label |
| `TestKnownGap_RetractionNotices` | documents a case not yet handled, as a named gap rather than a silent one |

## Verification

| Check | Result |
| --- | --- |
| `go build ./...` | exit 0 |
| `go test ./internal/scengine/ -count=1` | ok, 0.586s |
| `git diff --stat` | 4 files changed, 765 insertions(+) |

No dependency changes. No existing behaviour changes for works that are not retracted.

### DIFF
--- a/library/other/scientific-consensus/internal/scengine/normalize.go
+++ b/library/other/scientific-consensus/internal/scengine/normalize.go
@@ -1,0 +1,38 @@
+package scengine
+
+import "strings"
+
+// normalizeText folds the typographic characters that scientific publishers use
+// into their ASCII equivalents, so that a plain strings.Contains match behaves
+// the same on "Omega-3" and "Omega-3" with a typographic hyphen.
+//
+// Measured on the twelve full corpora: 107 occurrences of six distinct non-ASCII
+// dash characters. One of them cost a real exclusion - "Omega-3 polyunsaturated
+// fatty acids and inflammatory processes" (1332 citations, supporting) uses
+// U+2010 HYPHEN, a different byte sequence from the U+002D HYPHEN-MINUS in the
+// claim "omega-3 improves cardiovascular health".
+//
+// Only dashes are folded. Accents, ligatures and quotation marks are left alone:
+// they were not measured to cause misses, and folding them would be a larger
+// change than the evidence supports.
+//
+// Both readers of a work's text call this - the PICO gate and the stance
+// classifier - so the two can never disagree about what the text says.
+func normalizeText(s string) string {
+	return dashFolder.Replace(s)
+}
+
+// dashFolder maps the dash-like runes found in the corpora to ASCII "-".
+//
+// Every entry below was counted in testdata/corpora_full except U+2012, which
+// is included to close the Unicode dash block; adding further speculative
+// entries would widen the fold without evidence.
+var dashFolder = strings.NewReplacer(
+	"\u2010", "-", // HYPHEN               30 occurrences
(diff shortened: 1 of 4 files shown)


tmchow:
The detector and tests are fine, but `DetectRetraction` / `ExcludeFromScore` never run on a real consensus fetch, so retracted works still score. Wire it through OpenAlex `is_retracted` and the scoring path, and add the patch-ledger entry, before this can merge.

laci141:
Closing this as superseded. All three review comments are now addressed on
main, but through #1841 and #1760 rather than here.

Retraction exclusion is connected: filterRetracted has four call sites in
scwork.go on main, and is_retracted is requested explicitly in the OpenAlex
select so DetectRetraction gets both signals rather than degrading to the
title rule.

normalizeText has a caller: the PICO gate folds text through it before
matching, so the gate and the stance classifier cannot disagree about what a
work's text says.

Both patch ledger entries exist:
retraction-excluded-from-the-scored-corpus.json and
pico-gate-on-the-production-relevance-path.json.

This branch holds an earlier, unwired version of the same work, so merging it
would only reintroduce the disconnected detector. Thanks for the review — the
three comments are what shaped the wiring that landed.

PR #1760, opened 2026-08-19
feat(scientific-consensus): refutation override and PICO relevance gate

laci141:
## Summary

Two connected changes to the keyless heuristic stance classifier, plus the stored corpora that pin their behaviour.

18 files, 5516 insertions, 22 deletions. About 400 lines are engine code; the remaining ~5100 are 13 stored API responses used as test data.

## Refutation override

The counting pass tallies positive and negative cues across an abstract and takes the majority. That fails on a specific and common shape: a paper whose **conclusion directly refutes** the claim, while its **background section** is full of harm vocabulary.

`The evidence does not support a causal association between X and Y` is a direct answer to the claim. It must not be outvoted by the word "risk" appearing four times in an introduction that is merely summarising why the question was asked.

`refutationOverride` runs after the counting pass and can override it:

- **Strong cues** match over the full text without pairing. They are direct answers, so they legitimately outrank a vocabulary tally.
- **Meta cues** describe a controversy rather than a finding, so they are gated by pairing to the claim's intervention and outcome. A paper reporting that a debate exists is not itself refuting evidence.

A framing gate guards both: refutation phrasing posed as a *question* is not a refutation.

## PICO relevance gate

Scoring every retrieved work against the claim treats an off-topic paper with heavy citation mass as if it were evidence, and citation mass is what weights a work in the aggregator.

The gate requires the claim's intervention side and outcome side to both appear before a work contributes. Content tokens are truncated to a fixed stem length so inflected forms still match.

## Sentence scoping

Positive cues are paired to the claim within the enclosing sentence, intersected with a 120-character window. The window is a **cap on an already sentence-scoped search**, never a way to reach across a sentence boundary.

Negation is searched 25 characters back. Deliberately short: long enough for `does not increase the risk`, short enough that `no benefit was seen, but mortality increased` is not misread as a single negated clause.

## Test data

`testdata/corpora/` holds 13 stored OpenAlex responses covering claims that are true, false, and genuinely contested. They are real API captures rather than hand-written fixtures, so cue matching is exercised against abstracts exactly as they arrive, including their punctuation, hedging, and section structure.

I am aware 5100 lines of JSON is a lot to add. The alternative was hand-written fixtures, and those would have quietly passed every one of these gates, because I would have written them to match what the code already does. These captures are the reason the four dead ends below were measurable at all.

Two tests use them:

| Test | What it pins |
| --- | --- |
| `corpus_regression_test` | current verdict, score and label distribution across all 13 corpora |
| `pico_impact_test` | how many works the gate excludes per corpus, so a future change that quietly widens or narrows it is visible |

## What did not work

Four approaches were tried against these corpora and abandoned because the measurement did not support them:

- **Epidemiological cue expansion** — 38 raw signals added, only 3 survived the existing gates
- **Widening the pairing window** — the vaccines corpus moved from 6 to 15 works scored positive, the wrong direction
- **Requiring all claim stems rather than any** — broke claim-aware classification, because real abstracts write "Sweetener" where the claim says "artificial sweeteners"
- **A head-stem rule using the longest stem per side** — the stemmer truncates to 5 characters, making `artif` and `sweet` the same length, so coffee and sweetener corpora dropped to zero

Recording these because the corpora are what made them falsifiable, and because a future contributor is otherwise likely to try the same four.

## Verification

| Check | Result |
| --- | --- |
| `go build ./...` | exit 0 |
| `go test ./internal/scengine/ -count=1` | ok, 0.843s, 0 failures |
| `gofmt -l` on the staged (LF) copy | clean |
| `git diff --stat` | 18 files changed, 5516 insertions(+), 22 deletions(-) |

No dependency changes. The keyless heuristic remains the default path.


tmchow:
Same shape as the retraction PR: the PICO helpers and stored corpora are test-only. Production still uses the old title/topic overlap filter. Connect `IsPICORelevant` on the consensus path, and add a `.printing-press-patches/` record. `normalize.go` also lands in #1759 — keep those from diverging.

laci141:
Addressed in 996657223 (gate) and a5271fdbf (ledger), rebased onto current main.

IsPICORelevant is now called from filterRelevant, the single place the consensus
and compare paths already pass through, so the gate reaches production without a
second relevance rule appearing anywhere.

The two gates are alternatives rather than layers. PICO runs when the claim
splits around a polarity verb; when PICOTokens returns nothing the older
title/topic overlap rule runs instead. Layering them would apply a title-only
rule on top of an abstract-aware one and re-drop works PICO deliberately kept,
and removing the older rule outright would leave unsplittable claims with no
relevance filtering at all.

The tests pin which gate runs, not that filtering happens: every work in the
PICO case carries an intervention token in its title, so the overlap rule would
keep all of them. Forcing the PICO branch off fails the test with both works
kept.

normalize.go does not exist on main, so there is no divergence with #1759 to
reconcile — this branch introduces it.

laci141:
Both addressed in 5b9b3b3c0.

The gate reached only consensus. compare, controversies and report called
scoreWorks directly, so off-topic works still moved their verdicts — and in
controversies and report the mismatch was visible in the output, since
StudyCount and Analyzed read len(works) and counted works the stance tallies
never described. All four paths now gate before scoring, verified by grep for
the call site in each file.

The gate stays at the call sites rather than moving inside scoreWorks precisely
because of those two figures: filtering inside would leave the reported count
and the stats summarising it describing different sets. report keeps
TotalMatches ungated on purpose — it reports what the query matched upstream,
not what the run analyzed.

The ledger now covers the refutation override as well: stance.go and its cue
structure, the pairing rule for meta cues, and the framing gate. It listed only
the PICO changes before, so a regenerated package would have kept the gate and
lost the override.

tmchow:
Thanks — the production-path wiring looks right. `IsPICORelevant` is on every scoring entry point now, and Greptile is 5/5 with CI green.

Not merging yet only because #1841 (retraction exclusion on the live path) is still open on the same CLI and should land first so we do not force a double rebase. Once that is in, this is ready to merge as-is (or with a trivial rebase onto main).

Clearing `needs-author` → `awaiting-maintainer`.

tmchow:
Rebased onto main after #1841. Conflict in `compare.go` `computeConsensus` resolved by keeping both gates in the consensus order: `filterRelevant` then `filterRetracted`. Local `go test ./internal/cli/ ./internal/scengine/` passed. Waiting on CI, then merging.

tmchow:
Your PICO / refutation work looked good (Greptile 5/5), but it conflicted with #1841 which just landed the retraction-exclusion leftover on the same files.

Opened origin follow-up #1851 with your changes rebased onto current main. Will merge that and close this as superseded once CI is green.

tmchow:
Merged via origin follow-up #1851 (after #1841 landed first). Closing this as superseded — thanks @laci141.

PR #1770, opened 2026-08-20
fix(scientific-consensus): let a stronger heuristic outrank a generic review pubtype

laci141:
fix(scientific-consensus): let a stronger heuristic outrank a generic review pubtype

ClassifyDesign returned on the first authoritative publication type it
recognised. Every entry in pubtypeMap names a specific design except one:
"review", which a source can apply to a work whose own title and abstract
call it a meta-analysis. When that generic label was all that came back, the
cascade stopped there and the title/abstract heuristics never ran.

Measured on 10.3389/fnut.2022.1084455, a 2023 pooled analysis of 55 RCTs
whose title reads "A systematic review and meta-analysis" and whose abstract
repeats it twice. Tagged only "Review", it classified as narrative-review at
tier 1. With the pubtypes removed so the heuristics could run, the same
title and abstract classified as meta-analysis at tier 9 — an eight-point
swing on a work the consensus engine should weight most heavily.

The generic label now defers: the heuristics run, and the stronger of the
two answers wins. A specific pubtype still returns immediately and is never
overridden by a weaker heuristic, and "review" still stands when the
heuristics find nothing better. Covered by five table cases.

### DIFF
--- a/library/other/scientific-consensus/.printing-press-patches/generic-review-pubtype-defers-to-heuristic.json
+++ b/library/other/scientific-consensus/.printing-press-patches/generic-review-pubtype-defers-to-heuristic.json
@@ -1,0 +1,13 @@
+{
+  "schema_version": 2,
+  "id": "generic-review-pubtype-defers-to-heuristic",
+  "applied_at": "2026-08-20",
+  "base_run_id": "20260620-165352-e55cec69",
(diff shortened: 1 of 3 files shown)


tmchow:
Generic `review` deferring to a stronger title/abstract heuristic is right, and the table tests pin the specific-pubtype path. Merging.

PR #1773, opened 2026-08-20
fix(peloton): repair auth, offline/sync store gaps, and add bounded/resumable dependent sync

jrmii:
## Summary

A round of live-verified fixes to `peloton-pp-cli`/`peloton-pp-mcp`: authentication was silently broken on every single-fetch and MCP tool call, `sync`/`offline` wrote to and read from different tables, `workflow_archive` archived nothing regardless of flags while exposed as an MCP tool, and the dependent-sync fan-out (performance/workout_details, one HTTP request per workout) was unbounded and unresumable — meaning a real account's backlog could never complete inside an MCP tool call's timeout. This PR fixes all of that.

A companion PR (#TBD, `feat/peloton-classes-detail-sync`) adds a new `classes_detail` dependent resource on top of this branch — it's stacked on this one and will show a smaller diff once this merges.

## What Changed

**Auth (was silently broken)**
- Single-fetch commands and MCP tool calls never attached the session cookie — `/api/me` and `/api/workout/{id}/performance_graph` returned 401 while `sync` succeeded in the same process with the same credentials. Managed auth is now wired into every call path.
- The real auth model is `POST /auth/login {username_or_email, password}` → a `peloton_session_id` cookie, plus a `peloton-platform: web` header some endpoints require. **Not OAuth.** The prior "managed OAuth bundle" / "oauth2_refresh" vocabulary in `doctor`/`agent-context` was actively misleading (cost real debugging time) and has been corrected throughout — please don't reintroduce that framing in future changes to this CLI.

**Offline/sync store gaps**
- `sync` wrote synced data into one table; the `offline` read commands consulted a different one that stayed empty. Reconciled onto a single write-through path (`UpsertBatchWithFacts`/`UpsertWithFacts`) so every sync path lands in both.
- `doctor` status reporting was wrong as a downstream consequence; fixed alongside.
- Performance graphs were stored at ~1/50th resolution (25 samples for a 1200s workout; the live endpoint returns 1s) — the dependent fetch never set `every_n=1` the way the single-fetch command already did.
- `workout_details`/`strength` made syncable (backs `offline workout`/`intervals`/`repeat`/`strength`).
- `resolveLocal` 404'd on any endpoint with a trailing sub-resource path segment.
- `ride.id` (nested) vs `ride_id` (top-level, promoted at write time) path bugs across workouts/classes/workout_details — the API nests class association under `ride.id` and the top-level `ride_id` field several read paths assumed only exists if something promotes it there first.

**Bounded/resumable dependent sync**
- `--max-parents`, `--max-pages`, `--stale-before` on `sync`, with oldest-fetched-first candidate ordering so a capped call always makes progress on the most-overdue records. `--latest-only` scopes a dependent's fan-out to only what the current run's flat phase actually touched.
- `workflow_archive` had a hardcoded-empty resource list (`resources := []string{}`) — every invocation silently reported `{"resources_synced":0}` and archived nothing, on any account, regardless of flags, while being exposed as an MCP tool (an agent calling it saw a clean success response for a no-op). Now delegates to `sync` and inherits its bounding flags.
- Freestyle/non-class workouts (Just Run, Outdoor Running, Just Ride) carry Peloton's all-zeros sentinel ride id in place of a real class association. `offline_repeat`/`offline_intervals` now treat it as "no class" instead of a false same-class match or a misleading "class data unavailable" caveat.
- MCP context tool rewritten off stale CRM-domain scaffolding that didn't describe this CLI's actual (health/fitness) domain; redundant catalog vocabulary stripped from `classes_catalog`/`classes_search` responses.

## Validation

Reproducible by anyone, run from `library/health/peloton/`:
```
go build ./...
go vet ./...
go test ./...
govulncheck ./...
python3 .github/scripts/verify-skill/verify_skill.py --dir library/health/peloton/
```
All green at this branch's tip (`4c36507b7`), independently re-checked (not just at the overall working branch's later tip).

**Live verification** (separate, explicitly caveated category — this was black-box testing against a real account, not a code review, and doesn't attest to whether a fix generalizes beyond what was exercised): every fix above was reproduced live before the fix and confirmed resolved after, across several verification rounds against a real Peloton account (auth 401→200, `workflow_archive` 0-synced→populated store, performance samples 25→1200 for a 1200s workout, etc.). Tested against one account, US/imperial/English/all-access content only — no coverage of metric units, non-English captions, or non-US content availability.

## Open question for a maintainer

Some of the touched helpers (`--select` field projection, `--compact` recursion, caveat handling) live in patterns this CLI's generator likely stamps into other printed CLIs too. Everything in this PR is local to `library/health/peloton/`; I haven't checked whether the equivalent code in sibling CLIs has the same defects. If the underlying pattern is generator-shared rather than local, the analogous fix likely belongs in `cli-printing-press` as well — flagging rather than guessing.

## Patches index

Every code-level customization in this PR is recorded under `library/health/peloton/.printing-press-patches/` per this repo's convention, one file per patch.


tmchow:
The auth, offline/sync, and bounded-sync work looks right, and you already reproduced it live. One leftover on my side: after the capped fan-out, `SaveSyncStates` errors are still discarded, so a failed checkpoint can report progress. I will surface that, then this can land.

#1774 is stacked on an older tip of this branch (through 4c36507) and still needs this PR first.

tmchow:
Pushed to this branch: `SaveSyncStates` errors after capped/`--full` dependent fan-out are no longer discarded. If the atomic cursor/backlog/turn checkpoint fails, the sync command returns that error instead of reporting progress. Unit test covers the failure via a SQLite trigger on those keys. CI is rechecking.

tmchow:
@greptileai review

Leftover head is `5347b163e1518492dbaa2454c8cd6f2df52c7fde`. Previous review on this SHA skipped (error `f179a476-168e-4bc1-a7bb-b1cc6e01144b`). Please score this SHA so the policy gate can recheck.

tmchow:
Squash-merged. The leftover on `SaveSyncStates` is in: a failed cursor/backlog/turn checkpoint now fails the sync command instead of reporting progress. Greptile scored that head 5/5 and the policy gate is green.

Landed as `f45ab2336707acf9b267d59512a069cccec2f2e9`.

PR #1774, opened 2026-08-20
feat(peloton): add classes_detail dependent sync, and the bugs found building it

jrmii:
## Summary

Adds `classes_detail`, a new parent-keyed dependent sync (alongside the existing `performance`/`workout_details`), plus the bugs found while building and live-verifying it.

**Stacked on #1773** (`fix/peloton-auth-and-bounded-sync`) — this branch includes all of that PR's commits. The diff here will shrink to just this PR's own delta once #1773 merges and this branch is retargeted/rebased.

## Why

The bulk classes catalog endpoint (`GET /api/v2/ride/archived`) never returns `segments`/`target_metrics_data` — only the per-class detail endpoint (`GET /api/ride/{ride_id}/details`) does. On a real verification account, 99.8% of synced classes (48,645 of 48,743) had no segment data as a result, making `offline_classes_structure`/`offline_intervals` unusable for almost the entire synced catalog. The caveat text also actively misled: "stored class has no comparable segment list" read as "this class has no segments" when the truth was "this class was synced in catalog-list form and its segments were never fetched."

## What Changed

**The `classes_detail` dependent**
- Fans out over the distinct classes an account holder actually took (`workouts.ride_id`), not the full catalog — hundreds of ids on a real account, not tens of thousands.
- Content-based pending check (does the record already carry `segments`), not presence-based — every taken class already has a `classes` record from the flat catalog sync, so a presence check would see it as permanently done.
- Merges into the *same* `classes` record the flat catalog sync already created, via the same write-through path a live `classes show`/`classes structure` call uses (including its nested `ride.id` id resolution).
- Cascades automatically whenever `classes` is synced, the same way `workouts` cascades into `performance`/`workout_details`.
- Bounded/resumable via the same `--max-parents`/`--stale-before` machinery #1773 adds.

**Bugs found building and verifying it**
- The flat classes catalog sync was wholesale-overwriting `classes_detail`'s merged records on every routine `sync`/`workflow archive` call, silently stripping segments back out (confirmed live: 187→156 synced-with-segments across one archive call). Fixed with merge-on-write: a list-shaped incoming item now preserves any detail-only field an existing detail-shaped record already has.
- The typed `workouts.ride_id` column (populated by a write-time promotion) only covered ~60% of ride-bearing workouts on a real account (511 of 843) — older or otherwise-written rows kept the nested `ride` object with no corresponding column value. `classes_detail`'s parent-id source now falls back to the nested `ride.id` field when the column is empty.
- The freestyle/non-class sentinel ride id (see #1773's `offline_repeat`/`offline_intervals` fix) wasn't excluded from this dependent's own fan-out — wasted one API call and logged one anomaly on every single invocation, and since the pending check is content-based, could never be satisfied, so it stayed permanently pending.
- MCP `search --select` was a silent no-op: the hand-written `search` tool never declared a `select` parameter, so passing one did nothing (`--select results.id,results.title` still returned full ~15 KB records).
- `offline_history` returned `"caveats": null` instead of omitting the key when there was nothing to report, so callers couldn't test for presence without also null-checking.
- Offline class readers (`classes show`/`structure`/`intervals`) 404'd with "stored catalog_classes fact not found" for ids the generic `resources` table actually had synced, because the secondary `provider_payloads` cache index had a small (97-row) gap against it. Now falls back to `resources` on a cache miss.
- A general `--select` projection bug: any object with a non-matching selector and *any* array-valued field (e.g. `segments`, `related_rides`) tripped an envelope-fallback heuristic meant for this CLI's own thin wrapper shapes (`{"items":[...]}`), and leaked every other field completely unfiltered — including large nested objects (playlist, instructor bio, `workout_share_images`). Bounded the heuristic to objects with at most 5 top-level keys (every real wrapper shape here tops out at 4; a raw upstream detail payload has far more).
- `workflow_archive`'s `--max-pages`/`--max-parents` now default to bounded values (5/500) specifically when invoked as an MCP tool call with no explicit flags, so an agent that doesn't already know about those flags still gets a timeout-safe call. A direct shell invocation keeps today's unbounded-by-default behavior.

## Test corpus

The account used for live verification is worth offering independent of this specific patch — its long tail is what surfaced several of these bugs (the freestyle sentinel from years of Just Run/Outdoor sessions, the sparse `ride_id` column from older synced rows):

```
3,596 workouts, 2016-01-23 through 2026-08-19
9 disciplines: stretching 1561, running 1067, cycling 601, strength 199,
               yoga 119, walking 16, meditation 12, circuit 3, cardio 2
10 device/platform combinations, including a 3rd-party Garmin Connect sync
   and a Guide (t21n8m2/tiger) device
48,894 catalog classes synced; 315 distinct classes actually taken
Gen-1 Bike (lookup-table power, not strain gauge)
```

## Open question for a maintainer

The `--select` envelope-fallback bound above (≤5 top-level keys) is a heuristic tuned to this CLI's own wrapper shapes. If `filterFieldsRec`/`filterFields` in `internal/cli/helpers.go` is boilerplate the generator stamps into every printed CLI rather than local to peloton, the same leak likely exists in sibling CLIs, and the fix probably belongs in `cli-printing-press` too. I haven't checked other CLIs — flagging rather than guessing.

## Validation

Same commands as #1773, run from `library/health/peloton/` — all green at this branch's tip:
```
go build ./...
go vet ./...
go test ./...
govulncheck ./...
python3 .github/scripts/verify-skill/verify_skill.py --dir library/health/peloton/
```

**Live verification** (separate, explicitly caveated — black-box testing against one real account, not a code review): each fix above was reproduced live before the fix and confirmed resolved after, across three verification rounds, culminating in:
```
sync --resources classes_detail --max-parents 150
-> 130 stored, no anomaly event
(re-run) -> already_up_to_date, total 0
classes with segments: 232 -> 362
```
Two prior root-cause theories from that verification session were wrong and were corrected during this work: a timestamp-format theory for `--stale-before`'s ordering (actual cause was candidate ordering, fixed in #1773) and a "resources-fallback bypasses projection" theory for the `--select` leak above (actual cause was the envelope-fallback heuristic). Noting this so the fixes are read as verified-by-reproduction, not verified-by-explanation.

Same locale/account caveat as #1773 applies: one account, US/imperial/English/all-access only.

## Patches index

Every code-level customization in this PR is recorded under `library/health/peloton/.printing-press-patches/` per this repo's convention, one file per patch.


tmchow:
Do not merge this before #1773. You branched from 4c36507, then both sides kept moving — this one has classes_detail, 1773 has the later doctor/docs commit. After 1773 lands this will need a rebase.

Same leftover as 1773: the discarded `SaveSyncStates` error after a capped `--full` fan-out. I will fix that here too so the rebase is smaller. Live reproduction on classes_detail (130 stored, then already_up_to_date) is enough for an existing-CLI change.

tmchow:
Pushed a maintainer fix to this branch for the leftover Greptile P1: a capped `--full` fan-out now fails the resource if `SaveSyncStates` cannot persist the cursor/backlog/turn checkpoint, instead of reporting success. Unit tests passed locally (`go test ./...` under Go 1.26.6). CI is rechecking this push. This PR still must wait for #1773 — no rebase or merge from here.

tmchow:
Rebased onto main after #1773 and squash-merged. The leftover `SaveSyncStates` work was already in 1773, so it dropped off this branch. classes_detail is in: fan-out for taken-class details, merge that keeps segments, the null-ride fix, and the freestyle sentinel skip. Live run was 130 stored, then already_up_to_date. Greptile 5/5 on eb3187a.

Landed as `468965a6256a732add10cf35f442a9f52166e634`.

tmchow:
Rebased onto main after #1773 (`f45ab233`) landed. Dropped the leftover `SaveSyncStates` fatal commit (`9d0b28a`) — that fix is already on main via #1773. CI green, Greptile 5/5 on `eb3187a`. Squash-merging.

PR #1775, opened 2026-08-21
feat(travelclick): add travelclick

enlewof:
## travelclick

Search any TravelClick-powered hotel's own rates directly -- room types, rate plans, fee-inclusive pricing, and the cheapest night to book -- without opening a browser tab per property.

**API:** travelclick | **Category:** travel | **Press version:** 4.30.3
**Spec:** Not specified

### Publication Path

New print

### CLI Shape

```bash
$ travelclick-pp-cli --help
Travelclick CLI — Search any TravelClick-powered hotel's own rates directly -- room types, rate plans, fee-inclusive pricing, and the cheapest night to book -- without opening a browser tab per property.

Highlights (not in the official API docs):
  • rates compare   Check several boutique hotels for the same dates in one call, ranked by lowest fee-inclusive total.
  • rates cheapest-night   Scan several hotels' calendars at once and return the single best hotel+date combination.
  • hotels alias   Give a memorable name to a TravelClick hotel ID instead of memorizing a 6-digit number.
  • analytics price-drift   Track how a hotel's rates move over time from your own saved search history.
  • codes check-all   Test one corporate or group code against every saved hotel at once.

Agent mode: add --agent to any command for JSON output + non-interactive mode.
Health check: run 'travelclick-pp-cli doctor' to verify auth and connectivity.
See README.md or the bundled SKILL.md for recipes.

Usage:
  travelclick-pp-cli [command]

Available Commands:
  agent-context  Emit structured JSON describing this CLI for agents
  analytics      Local state that compounds
  api            Browse all API endpoints by interface name
  auth           Manage authentication for Travelclick
  codes          Validate special rate codes (corporate/rate-access or group-attendee) against a hotel before applying them to a search.
  completion     Generate the autocompletion script for the specified shell
  doctor         Check CLI health
  feedback       Record feedback about this CLI (local by default; upstream opt-in)
  help           Help about any command
  hotel          Fetch a hotel's public profile: name, address, geolocation, check-in/out times, accepted cards
  hotels         Local state that compounds
  learnings      Inspect or forget the local search_learnings table
  playbook       Inspect or amend stored CLI playbooks
  profile        Named sets of flags saved for reuse
  rates          Search room availability and rate plans for specific dates, or scan a date range for the cheapest night.
  recall         Check prior learnings for a query before running discovery (LLM-fired, pre-discovery)
  teach          Record a query -> resource mapping for future recall (LLM-fired, silent)
  teach-lookup   Install a manual entity-lookup row (kind, canonical, value)
[...]

PR #1776, opened 2026-08-21
docs(travelclick): document the corporate-code-to-rates-search gap

enlewof:
## Summary

Found this gap firsthand today while checking a real corporate rate's price for another project: `codes validate-corporate`/`validate-group` can only confirm whether a rate-access code is valid — `rates search` has **no parameter at all** to actually filter results to that code's rate plan once validated. The existing README `## Known Gaps` entry noted the success-response shape was unconfirmed but didn't call out this concrete consequence, and the "Check a corporate code before searching" recipe's own wording ("before building a search around it") implies a chained workflow that doesn't actually exist in the command surface.

## Evidence

- Read the actual discovery sample for `codes validate-corporate` (`travelclick-browser-sniff-spec-samples/get__ibe-codes_v1_hotel_hotel_id_specialcodes_corporate_TEST123__2f95cd16.json`): the only captured response is a 404 (`INVALID_CORP_ID`, from an intentionally-fake `TEST123` code). A successful validation's response shape was never observed during discovery — so there was never a field available to wire into `rates search` even if someone tried.
- Read `rates_search.go` directly: confirmed its `params` map has no corporate-code, rate-plan-id, or similar field among `dateIn`, `dateOut`, `adults`, `infants`, `rooms`, `currency`, `lang`, `partnerIdentifier`, `bookerIdentifier`.

## What this PR adds

- A `## Known Limitations` section in SKILL.md explaining the gap precisely, with the exact discovery-sample evidence above.
- A **verified workaround**: the booking widget's own page accepts a `RatePlanId` query parameter that pre-selects a specific corporate rate directly — `https://reservations.travelclick.com/{hotel_id}?RatePlanId={id}` (redirects to `bookings.travelclick.com`). Confirmed working today against hotel `102306` / `RatePlanId=5835835` — loads with the correct corporate rate ("Magnite Inc") pre-selected under Accommodations, real price readable from the rendered page.
- Corrected the misleading recipe text in both SKILL.md and README.md (same text was duplicated in both).
- Enhanced the existing README `## Known Gaps` bullet with the concrete consequence and the verified workaround, rather than adding a duplicate/conflicting entry.

## What this PR deliberately does not do

- Does not add a guessed `--corporate-code` or `--rate-plan-id` flag to `rates search`. The `RatePlanId` parameter is confirmed working at the **page level** (the rendered widget), not confirmed as an **API-level** query parameter on `/ibe-shop/v1/hotel/{hotel_id}/avail` — it was observed through the UI, not captured at the network layer. Wiring a flag onto an unconfirmed parameter name would be a guess dressed up as a fix. The path to a durable fix is named explicitly in the new section: re-run discovery's browser-sniff against `codes validate-corporate` with a real (non-`TEST123`) code, and re-sniff `avail` while the widget has a `RatePlanId` selected, to capture both missing shapes for real.
- Docs-only change; no Go code touched.

## Verification

- `cli-printing-press verify-skill --dir library/travel/travelclick` passes clean after the edit.

### DIFF
--- a/library/travel/travelclick/README.md
+++ b/library/travel/travelclick/README.md
@@ -213,7 +213,8 @@ Returns the lowest rate for every day in the range so you can sort for the minim
 travelclick-pp-cli codes validate-corporate 102306 ACME2026
 ```
 
-Confirms a rate-access code is live for this hotel before building a search around it.
+Confirms whether a rate-access code is live for this hotel. **Validating a code does not
+currently let you search rates for it** — see Known Gaps below.
 
 ### Compare three boutique hotels for the same weekend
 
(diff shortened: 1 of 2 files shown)


tmchow:
Merged. Docs-only follow-up to #1775: rates search still has no --corporate-code or --rate-plan-id (confirmed in rates_search.go), the corporate-code recipe no longer implies a chained search, Greptile is 5/5, and library CI is green.

ppl-triage: merged b6357e247080787066b80ee9efa239ff5f8bbe0e

PR #1780, opened 2026-08-21
test(peloton): add live-round-14 turn-bit regression test

jrmii:
## peloton

Follow-up to #1773/#1774's --full dependent-sync resumability work. A live-account verification round (7 consecutive `--full` calls against `workout_details`, ~3,596 workouts, `--max-parents 40` x5 then `2` x2) confirmed the sweep-cursor guarantee end to end -- the offset advanced monotonically by exactly `--max-parents` every call, cross-checked against a direct `COUNT(*)` of newly-synced rows with zero divergence at any checkpoint.

That same round flagged something that looked like a discrepancy: the persisted turn bit (`sync_state` row `<resource>:full_turn`, used to alternate priority between the sweep and the failed-id backlog at `--max-parents 1`) appeared to flip once then stay flat across several subsequent calls, which looked inconsistent with the documented "flips unconditionally every call" behavior.

**What this PR adds:** `TestFullSyncTurnBitFlipsEveryCallAcrossMixedCapSizes` reproduces the exact live call sequence (500 synthetic ids, `--max-parents 40` x5 then `2` x2, nothing failing) through the real fetch/store path. It reproduces the live report's offset progression exactly (40, 80, 120, 160, 200, 202, 204), and the turn bit alternates 1,0,1,0,1,0,1 correctly on every call -- so the live-account discrepancy was not reproduced here. The code appears to behave as designed; the live observation is more likely an artifact of how that round's ad hoc SQL tool read the database (e.g. a stale read across separate queries) than a code defect, but I'm not able to confirm that without another live round, so this PR only adds the passing regression test and documents the round's findings -- it does not claim to have fixed anything or fully explain the original observation.

Also updates the existing `full-sync-offset-persists-after-success` patch entry with the round-14 results: guarantee (1) (sweep cursor) confirmed live; guarantee (2) (failed-id backlog retry) was not naturally exercised (zero fetch failures occurred across all 204 requests), so it remains verified only by unit tests; and confirms `last_synced_at` updates on all three `sync_state` rows on every call regardless of content change (intended -- one timestamp per atomic transaction, not a per-row dirty flag).

### Validation Results

| Check | Result |
|---|---|
| `go build ./...` | PASS |
| `go vet ./...` | PASS |
| `go test ./...` | PASS |
| `gofmt -l .` | clean (no new findings; one pre-existing unrelated file already had drift before this PR) |

No functional code changes -- test and patch-documentation only.


tmchow:
Round-14 turn-bit test drives the same 7-call mixed-cap sequence through the real fetch/store path and pins the offset (40 through 204) and the 1/0 flip. Real regression test, CI green, Greptile 5/5. Squash-merging.

PR #1786, opened 2026-08-22
fix(thelancet): decode HTML entities in curate titles, topics, and journals

laci141:
OpenAlex returns HTML-encoded strings in some title and display_name fields — for example 'Bile Acid \&amp; Tryptophan' instead of 'Bile Acid \& Tryptophan'. CurateLive passed these raw to WorkRow without decoding, so the entity appeared verbatim in every downstream export (JSON, CSV, XLSX, BibTeX).

Apply cliutil.CleanText (html.UnescapeString + TrimSpace) to Title, Topic DisplayName, and Journal DisplayName at the point of assignment to WorkRow.

Measured: 'Bile Acid \&amp; Tryptophan Metabolism' in reading-list JSON export. After this fix the value is 'Bile Acid \& Tryptophan Metabolism' in all four export formats.


laci141:
Closing in favour of a clean single-purpose PR. This branch carried the entire feat/thelancet history, which is why Greptile scored the whole package rather than the two-file fix. Reopening from a branch cut off main with only the CleanText change and its patch-ledger entry.

PR #1789, opened 2026-08-22
fix(thelancet): decode HTML entities in curate titles, topics, and journals

laci141:
OpenAlex returns HTML-encoded strings in some title and display_name fields. CurateLive assigned them to WorkRow without decoding, so the raw entity surfaced in every downstream export.

Measured: a reading-list export carried "Bile Acid &amp;amp; Tryptophan Metabolism" verbatim into JSON, CSV, XLSX, and BibTeX. In the BibTeX case bibEsc then escaped the ampersand of the entity itself, producing \\&amp;amp; in the .bib file.

cliutil.CleanText already exists and is applied at every other extraction point in this CLI — CurateLive was the one call site constructing rows without it. Three calls added: Title, primary_topic.display_name, primary_location.source.display_name.

Second commit records the change in the per-CLI patch ledger, since AGENTS.md requires generated-tree patches to document why they belong here.

go build ./... and go vet ./... clean.

### DIFF
--- a/library/developer-tools/thelancet/.printing-press-patches/thelancet-curate-html-entities.json
+++ b/library/developer-tools/thelancet/.printing-press-patches/thelancet-curate-html-entities.json
@@ -4,11 +4,13 @@
   "applied_at": "2026-08-22",
   "base_run_id": "20260617-072858-2f9fe13e",
   "base_printing_press_version": "4.24.0",
-  "summary": "Title, Topic, and Journal are run through cliutil.CleanText once when Curate or CurateLive builds a WorkRow. The local store persists raw OpenAlex display fields so a regen must not re-introduce ingest-time decoding.",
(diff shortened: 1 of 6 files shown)


tmchow:
The CurateLive decode is the right call — `&amp;` in the export was real. After a sync, curate reads the store, and that path still copies OpenAlex title/topic/journal through raw. I'll apply the same CleanText there.

tmchow (approved):
Approved. One-pass CleanText on WorkRow construction matches CurateLive; ingest no longer pre-decodes, so nested &amp;amp; does not double-unescape on the store path.

tmchow (approved):
Re-approved on leftover e6d84f4: CleanText once at ingest (decodeWork/StoreWorks); Curate reads store as-is; TestCurateMatchesDecodedAmpersandQuery covers '&' LIKE against a title that arrived as &amp;. Greptile LIKE thread resolved as addressed.

tmchow:
Squash-merged leftover: persist CleanText once at ingest, store reads as-is, `&` query matches a title ingested from `&amp;`. CI green (Verify, Govulncheck, SKILL.md, Greptile + policy gate).

PR #1794, opened 2026-08-22
feat(jules): add jules

wryenmeek:
## jules

Jules Planning & Progress API for async coding tasks

**API:** jules | **Category:** developer-tools | **Press version:** 4.31.1
**Spec:** Not specified

### Publication Path

New print

### CLI Shape

```bash
$ jules-pp-cli --help
Drive Jules (async coding task) sessions, sources, and activities from the
command line, with quota-aware dispatch, continuous session monitoring,
working-tree checkpoints, and pre-submission compliance/diff gating on top.

Add --agent to any command for JSON output + non-interactive mode.
Run 'jules-pp-cli doctor' to verify auth and connectivity.

Usage:
  jules-pp-cli [command]

Available Commands:
  agent-context  Emit structured JSON describing this CLI for agents
  analytics      Run analytics queries on locally synced data
  archive        Archive stale sessions to free quota
  auth           Manage authentication for Jules
  checkpoint     Save and restore working tree checkpoints for sessions
  completion     Generate the autocompletion script for the specified shell
  compliance     Verify session changes against governance policies
  diff-validate  Validate session diffs before submission
  doctor         Check CLI health
  export         Export data to JSONL or JSON for backup, migration, or analysis
  feedback       Record feedback about this CLI (local by default; upstream opt-in)
  help           Help about any command
  import         Import data from JSONL file via API create/upsert calls
  learnings      Inspect or forget the local search_learnings table
  monitor        Monitor session progress with reconciliation support
  persona        Record and reuse Jules work patterns as personas
  playbook       Inspect or amend stored CLI playbooks
  profile        Named sets of flags saved for reuse
  recall         Check prior learnings for a query before running discovery (LLM-fired, pre-discovery)
  search         Search locally synced data
  sessions       List, get, create, delete, and send sessions
  sources        List and get sources
  sync           Sync API data to local SQLite for offline search and analysis
  tail           Stream live changes by polling the API at regular intervals
  teach          Record a query -> resource mapping for future recall (LLM-fired, silent)
  teach-lookup   Install a manual entity-lookup row (kind, canonical, value)
  teach-pattern  Install a manual generalization pattern (query_template, resource_template, entity_kind)
  teach-playbook Record a CLI playbook + free-text notes for a query family
  trigger        Chain cron schedules with GitHub workflow triggers
  version        Print version
  which          Find the command that implements a capability
  workflow       Compound workflows that combine multiple API operations

Use "jules-pp-cli [command] --help" for more information about a command.
```

### Novel Commands

[...]

PR #1796, opened 2026-08-23
chore: gitignore raw live-dogfood transcript in manuscript proofs

wryenmeek:
## Summary

Adds a `.gitignore` entry for `publish-live-gate.json` under any CLI's `.manuscripts/**/proofs/`, matching the existing pattern for raw `.log` dogfood output right above it.

## Why

`publish-live-gate.json` is the raw `dogfood --live --json` transcript written during `/printing-press-publish`'s Step 4.5 ("Live End-to-End Gate"). It samples real live-API response bodies into each test's `output_sample` field by design — useful for local review, but this file lives under a CLI's own `.manuscripts/` tree, which `publish package` bundles into the public PR without distinguishing "safe metadata" from "raw live-account dump."

This surfaced concretely on the `jules` CLI's publish PR (#1794): a real account's private org/repo names, an internal branch name, and a proprietary source-code diff ended up committed to a public fork before being caught and force-pushed away.

`phase5-acceptance.json` (also written by Step 4.5, via `--write-acceptance`) is unaffected — it's metadata-only (pass/fail counts, schema version, source-file hashes) and stays tracked as the canonical proof artifact.

## Scope

This is a backstop, not the root-cause fix. The actual fix belongs in `printing-press-publish`'s Step 4.5, which should stop writing this transcript under `.manuscripts/` in the first place — tracked separately as #4319. This `.gitignore` entry protects:

- CLIs printed/published before that fix lands
- any operator running an unpatched copy of the publish skill

## Related issues

- #4319 — the root-cause fix (Step 4.5 should not write the raw transcript into a manuscript path at all)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,6 +58,12 @@ Thumbs.db
 # canonical archive; raw .log files are local-only generation/dogfood output.
 library/**/.manuscripts/**/proofs/*.log
 
+# Raw live-dogfood transcript — samples real live-API response bodies
+# (account data, private repo/org names, source diffs). Never canonical
+# evidence; phase5-acceptance.json (metadata-only) is the safe artifact.
+# See #4319.
+library/**/.manuscripts/**/proofs/publish-live-gate.json
+
 # Local Git worktrees
 .worktrees/
 


tmchow:
This only gitignores the raw live-gate transcript under manuscript proofs, same pattern as the `.log` files above it. `phase5-acceptance.json` stays. CI green, Greptile 5/5. Squash-merging.

tmchow:
Landed as `f7db5d8eb607ea1b3c27393d27591efd92593014`.

PR #1805, opened 2026-08-24
feat(synology): add synology

smoochy:
## What this is

`synology-pp-cli`, a CLI for the Synology DSM WebAPI, printed from the locally installed `synology-mcp` server as the capability blueprint and verified end to end against a live DS415+ running DSM 7.1.1.

Surfaces covered: File Station (list, stat, search, download, and the mutating create/rename/copy/delete task commands), storage (volumes, disks, SMART), system (info, network, services, UPS, utilization), users, groups, and shared-folder permissions.

## Why it needed ten patches

DSM is not a REST API in the shape the generator assumes, and each mismatch is recorded under `.printing-press-patches/`. The three that matter most:

- **Every call returns HTTP 200.** Success and failure are signalled inside a JSON envelope, and the error codes are namespace-scoped, so code 400 means different things under `SYNO.API.Auth` and `SYNO.FileStation.List`. The transport unwraps the envelope and resolves codes per API instead of trusting the status line.
- **Auth is a session handshake, not a token.** The generated `auth login` surface is the framework's generic browser-cookie flow and does not apply; `session login --account --passwd [--otp-code]` does. Session credentials are redacted from output and deliberately kept out of the local store.
- **DSM tunnels mutations through GET.** The generator infers `mcp:read-only` from the HTTP method, so six destructive File Station commands shipped annotated read-only, which an MCP client that auto-approves read-only tools would have acted on. They are corrected to `false`.

The remaining seven cover endpoint corrections, a required parameter DSM does not document as required, a binary-response endpoint that was being parsed as JSON, a command that could never succeed and was dropped, and replacing placeholder examples with values DSM actually accepts.

## Verification

- `publish validate`: all eleven gates pass.
- Phase 5 live dogfood, `--level full`, against a live DS415+: **192 passed, 0 failed**, 101 skipped, `Verdict: PASS (with skips)`.

Two notes on the live run, both recorded in the patch entries rather than hidden:

1. The dogfood harness executes a command's `example:` string verbatim against the real target. The four mutating File Station examples therefore created, renamed and copied real entries on the NAS, and consecutive runs were not idempotent. Their examples now carry `--dry-run`, which also happens to be the right thing to show a human reading `--help` for a destructive command. The tradeoff is stated in `synology-mutating-file-station-safety.json`: those four commands now exercise request construction rather than a live mutation.
2. Phase 4.85 was skipped during the run because a host firewall was blocking outbound connections from freshly built Go binaries. That block was diagnosed and lifted, and Phase 5's live matrix then ran green, so the payload-shape concern is covered by evidence even though 4.85 itself was not re-run. The proof file states exactly what remains unverified.

## Manuscripts

- [build log](https://github.com/mvanhorn/printing-press-library/blob/4022eadbb1dd9f8b0cd5a8e90d93ed76bdb41133/library/devices/synology/.manuscripts/20260824-135140/proofs/2026-08-24-160642-fix-synology-pp-cli-build-log.md)
- [shipcheck](https://github.com/mvanhorn/printing-press-library/blob/4022eadbb1dd9f8b0cd5a8e90d93ed76bdb41133/library/devices/synology/.manuscripts/20260824-135140/proofs/2026-08-24-174500-fix-synology-pp-cli-shipcheck.md)
- [polish](https://github.com/mvanhorn/printing-press-library/blob/4022eadbb1dd9f8b0cd5a8e90d93ed76bdb41133/library/devices/synology/.manuscripts/20260824-135140/proofs/2026-08-24-190500-fix-synology-pp-cli-pp-cli-polish.md)
- [Phase 4.85 skip record](https://github.com/mvanhorn/printing-press-library/blob/4022eadbb1dd9f8b0cd5a8e90d93ed76bdb41133/library/devices/synology/.manuscripts/20260824-135140/proofs/phase-4.85-skip.md)
- [Phase 4.95 findings](https://github.com/mvanhorn/printing-press-library/blob/4022eadbb1dd9f8b0cd5a8e90d93ed76bdb41133/library/devices/synology/.manuscripts/20260824-135140/proofs/phase-4.95-findings.md)
- [Phase 5 acceptance marker](https://github.com/mvanhorn/printing-press-library/blob/4022eadbb1dd9f8b0cd5a8e90d93ed76bdb41133/library/devices/synology/.manuscripts/20260824-135140/proofs/phase5-acceptance.json)


tmchow:
Retitled to the usual add-synology shape. Live phase5 is 192/192 against a real NAS. Landing this.

PR #1822, opened 2026-08-26
feat(scientific-consensus): carry authors and journal into study briefs

laci141:
## Summary

The consensus output listed study titles with no authors, so downstream LLM
synthesis invented them: three runs over the same 2011 review produced three
different first authors. BibTeX export was missing `author` and `journal` for
the same reason.

## What Changed

The OpenAlex `select` in `scwork.go` already requested `authorships` and
`primary_location`, and both were parsed. The loss happened one layer later, at
normalization: `scWork` kept only `FirstAuthor`, so the full author list never
survived the decode. The fix therefore needed two hops, not one.

- `scWork` gains `Authors []string`, populated verbatim from every authorship in
  OpenAlex order. No truncation, no "et al.", no initials conversion.
- `workBrief` gains `Authors` and `Journal`, both `omitempty`. A work with no
  authorships, or no `primary_location.source`, is normal rather than an error —
  it emits nothing instead of an empty string or a placeholder.
- `compare` and `controversies` build their briefs through `topByStance` /
  `allStudyBriefs`, so they inherit both fields with no change of their own.
- `curateItem` gains `Authors`, and `renderCurateBibtex` emits
  `author = {A and B and C}`. `journal` was already wired from `Venue`.

Two things deliberately left alone:

- The `landmark` command prints `"would list landmark papers"` and returns no
  works, so there is nothing there to populate.
- `export.go` / `export_out.go` are the JSONL and xlsx paths; the BibTeX renderer
  lives in `analytics_works.go` behind `curate --format bibtex`.

This PR also realigns an anonymous struct in `newCitedBy` that `gofmt` flagged.
That nit predates this change and is unrelated to it; it is fixed here because
the CI gofmt gate aborts the run and this commit already touches the file.

## Manuscripts

Not applicable — this is a fix to an existing CLI, not a new publication.

## Validation

- [x] `python3 .github/scripts/verify-skill/verify_skill.py --dir library/other/scientific-consensus/`
- [x] `go build ./...` from the touched CLI root
- [x] `go vet ./...` from the touched CLI root
- [x] `go test ./...` from the touched CLI root
- [x] `govulncheck ./...` from the touched CLI root
- [x] `git diff --check`

`gofmt`: the worktree is CRLF, so a repo-wide `gofmt -l .` is not meaningful
here. Checked against LF copies of the staged files instead — all clean.

`govulncheck`: no vulnerabilities reachable from this code. It does report one
vulnerability in an imported package and one in a required module, neither of
which this code calls. Both predate this PR.

Tests added in `internal/cli/authorship_test.go`: a work with several authors, a
work with none, a work with no `primary_location.source`, and the BibTeX
rendering of a multi-author entry. `go test -count=1 ./internal/cli/` passes.

## Gaps / Follow-Up

The author list is passed through verbatim. If a downstream consumer needs a
shortened form ("Smith et al."), that belongs in the consumer, not here — the
CLI's job is to report what OpenAlex returned.

### DIFF
--- a/library/other/scientific-consensus/.printing-press-patches/bibtex-field-values-and-cite-keys-must-be-normalized.json
+++ b/library/other/scientific-consensus/.printing-press-patches/bibtex-field-values-and-cite-keys-must-be-normalized.json
@@ -1,0 +1,14 @@
+{
+  "schema_version": 2,
+  "id": "bibtex-field-values-and-cite-keys-must-be-normalized",
+  "applied_at": "2026-08-26",
+  "base_run_id": "20260620-165352-e55cec69",
+  "base_printing_press_version": "4.24.0",
+  "summary": "Anything the BibTeX export interpolates into a braced field value (title, each author name before joining, journal) must be normalized first, and the citation key must be a whitelist of ASCII letters, digits, '-' and '_' derived from the DOI with a refN fallback and per-render uniqueness — a raw value or a raw DOI silently corrupts the whole export, not just its own entry.",
+  "reason": "OpenAlex titles, author names and source names are free text: a brace terminates the field early and every subsequent entry is swallowed, a stray backslash starts a bogus LaTeX control sequence, and '&', '%', '$', '#', '_' are LaTeX specials. Braces and backslashes are dropped rather than escaped because a literal brace has no reliable representation inside a braced field value and is bad OpenAlex data, not content. Separately, the key was a two-character replace over the DOI, so a real production DOI (10.1016/s2213-8587(21)00051-6) produced '10_1016_s2213-8587(21)00051-6' — parentheses that break BibTeX parsing and Zotero import. Distinct DOIs can also normalize onto one key, and duplicate keys are an import error, so collisions are disambiguated in appearance order.",
+  "files": [
+    "internal/cli/analytics_works.go",
+    "internal/cli/bibtex_escape_test.go"
+  ],
+  "validated_outcome": "Tests pin a braced name, a backslashed name, '&'/'%'/'$'/'#'/'_' escaping, newline/CR/tab folding to a single space with runs collapsed and trimmed, typographic quotes left untouched as source data, the parenthesised production DOI yielding a legal key, an all-illegal DOI falling back to refN, and three colliding DOIs rendering as key, key_2, key_3. go build/vet/test green."
+}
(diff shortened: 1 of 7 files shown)


laci141:
All five checks pass.

On the escaping comment — the concern is valid and I'd rather fix it than
argue it, but scoping it to `author` alone would leave the file
inconsistent: `title` and `journal` are interpolated into
`renderCurateBibtex` exactly the same way and always have been. A name
with a brace is no more dangerous than a title with one.

Some data on how this actually shows up: across 99 exported entries from
a real run, no title contained a BibTeX-significant character
(`{ } \ % & $ #`). What did appear were five typographic quotes — `’`
three times, `“` and `”` once each — which LaTeX renders incorrectly
rather than breaking the parse. So the realistic failure mode here is
malformed output, not a corrupted file.

Two ways I can take this, happy with either:

1. Add a single `bibtexEscape` helper applied to `title`, `journal` and
   `author` together, in this PR. Slightly widens the diff beyond the
   author work, but leaves one consistent rule instead of one escaped
   field and two unescaped ones.
2. Keep this PR to the author propagation and open a follow-up that
   escapes all three fields at once.

I'd lean toward (1) since the escaping is a few lines and splitting it
means shipping a knowingly inconsistent renderer in between. Let me know
which you prefer and I'll push it.

laci141:
Fixed in 0218bfaaa, with the scope widened a little.

Escaping only `author` would have left the renderer inconsistent — `title`
and `journal` are interpolated the same way — so `bibtexEscapeValue` now
applies to all three, and to each author name before the join rather than
to the joined string.

Braces and backslashes are dropped rather than escaped. There is no
reliable way to emit a literal brace in a BibTeX field value, and an
unbalanced one terminates the field early and corrupts every entry after
it, so dropping is the safer failure. A brace in an OpenAlex author name
is bad data rather than content.

While in there I found a second unsafe interpolation that this comment
did not cover: the citation key. It was built by replacing only `/` and
`.` in the DOI, so a real production DOI `10.1016/s2213-8587(21)00051-6`
produced the key `10_1016_s2213-8587(21)00051-6` — parentheses are not
legal in a BibTeX key and break Zotero import. `bibtexCiteKey` now
whitelists `[A-Za-z0-9-_]` and disambiguates collisions deterministically.

Typographic quotes are left as-is. Across 99 entries from a real run they
were the only non-ASCII characters that appeared (`’` ×3, `“`, `”`); they
render imperfectly in LaTeX but they are the source data, so changing them
felt like a separate decision.

tmchow:
Authors and journal now survive into the study briefs, and the BibTeX escaping plus cite-key cleanup looks right. Merging.

PR #1823, opened 2026-08-26
feat(bmw-cardata): add bmw-cardata

tmchow:
Origin follow-up of #1417.

#1417 still has the CLI and the live proofs (phase5 full, 12/12), but it is 1103 commits behind `main` and dirty: the fork also carries old `.github/workflows/*` and `.go-version` edits that already landed on origin.

This branch is that CLI tree (`library/devices/bmw-cardata`) replayed onto current `main`. No rebase was pushed to the fork.

Original printer: @jvm


tmchow:
Superseded by #1828.

PR #1819, opened 2026-08-25
feat(carol-bike): add carol-bike

bricenice17:
## carol-bike

Read-only CLI and MCP server for CAROL Bike riders to access account statistics, ride history, calendars, trends, and a local SQLite mirror of their own workout data.

**API:** CAROL Bike Rider API | **Category:** health | **Press version:** 4.31.1

**Spec:** Sanitized browser-observed OpenAPI 3.0.3 specification for seven authenticated GET endpoints. The API is unofficial and unsupported by CAROL.

### Publication Path

Generated and validated through CLI Printing Press 4.31.1, then packaged under `library/health/carol-bike/`. This PR does not edit `registry.json` or `cli-skills/`; post-merge automation owns those generated surfaces.

### CLI Shape

```text
Carol Bike CLI — Read and locally mirror personal CAROL Bike workout data.

Highlights (not in the official API docs):
  • sync   Mirror personal CAROL ride history into local SQLite for dependable offline search and analysis.

Agent mode: add --agent to any command for JSON output + non-interactive mode.
Health check: run 'carol-bike-pp-cli doctor' to verify auth and connectivity.
See README.md or the bundled SKILL.md for recipes.

Usage:
  carol-bike-pp-cli [command]

Available Commands:
  agent-context  Emit structured JSON describing this CLI for agents
  api            Browse all API endpoints by interface name
  auth           Manage authentication for CAROL Bike
  completion     Generate the autocompletion script for the specified shell
  doctor         Check CLI health
  feedback       Record feedback about this CLI (local by default; upstream opt-in)
  help           Help about any command
  learnings      Inspect or forget the local search_learnings table
  playbook       Inspect or amend stored CLI playbooks
  profile        Named sets of flags saved for reuse
  recall         Check prior learnings for a query before running discovery (LLM-fired, pre-discovery)
  ride           Read personal CAROL Bike ride history.
  search         Search locally synced data
  stats          Read aggregate CAROL Bike rider statistics.
  sync           Sync API data to local SQLite for offline search and analysis
  teach          Record a query -> resource mapping for future recall (LLM-fired, silent)
  teach-lookup   Install a manual entity-lookup row (kind, canonical, value)
  teach-pattern  Install a manual generalization pattern (query_template, resource_template, entity_kind)
  teach-playbook Record a CLI playbook + free-text notes for a query family
  trends         Get rider trend series
  version        Print version
  which          Find the command that implements a capability
  workflow       Compound workflows that combine multiple API operations

Flags:
[...]

PR #1810, opened 2026-08-25
feat(is-agentic): add Is Agentic CLI

SomSamantray:
## is-agentic

Turn Is Agentic reports into durable, scriptable readiness evidence.

**API:** is-agentic | **Category:** developer-tools | **Press version:** 4.31.1  
**Spec:** Not specified

### Publication Path

Update existing PR #1810

### CLI Shape

```bash
$ is-agentic-pp-cli --help
Is Agentic CLI — Turn Is Agentic reports into durable, scriptable readiness evidence.

Highlights (not in the official API docs):
  • history   Keep a local, timestamped ledger of readiness reports and their provenance.
  • diff   See which readiness scores and findings changed between two retained audits.
  • check   Fail CI with an explicit, machine-readable readiness policy decision.
  • portfolio   Compare a fleet of public sites in one sortable score and issue matrix.
  • issues   Track when readiness findings first appeared, last appeared, were fixed, or regressed.
  • portfolio   Refresh many targets with bounded concurrency, deduplication, and Retry-After-aware pacing.
  • evidence   Package a report and its provenance into a portable evidence artifact.

Agent mode: add --agent to any command for JSON output + non-interactive mode.
Health check: run 'is-agentic-pp-cli doctor' to verify auth and connectivity.
See README.md or the bundled SKILL.md for recipes.

Usage:
  is-agentic-pp-cli [command]

Available Commands:
  agent-context  Emit structured JSON describing this CLI for agents
  check          Fail CI with an explicit, machine-readable readiness policy decision.
  completion     Generate the autocompletion script for the specified shell
  diff           See which readiness scores and findings changed between two retained audits.
  doctor         Check CLI health
  evidence       Package a report and its provenance into a portable evidence artifact.
  export         Export data to JSONL or JSON for backup, migration, or analysis
  feedback       Record feedback about this CLI (local by default; upstream opt-in)
  help           Help about any command
  history        Keep a local, timestamped ledger of readiness reports and their provenance.
  issues         Track when readiness findings first appeared, last appeared, were fixed, or regressed.
  learnings      Inspect or forget the local search_learnings table
  playbook       Inspect or amend stored CLI playbooks
  portfolio      Compare a fleet of public sites in one sortable score and issue matrix.
  profile        Named sets of flags saved for reuse
  recall         Check prior learnings for a query before running discovery (LLM-fired, pre-discovery)
  report         Read completed public Is Agentic reports.
  search         Search locally synced data
  sync           Sync API data to local SQLite for offline search and analysis
[...]

PR #1830, opened 2026-08-26
chore: drop leftover apify audit JSON

tmchow:
Drops `apify-actor-audit-report.json` from carol-bike, is-agentic, and dndbeyond. That file shouldn't ship; it landed on main with #1819 / #1810 / #1809 after the origin leftover follow-ups were closed.


tmchow:
Just dropping the leftover apify audit JSON that rode in with the three new CLIs. CI is green — merging.

PR #1841, opened 2026-08-27
feat(scientific-consensus): wire retraction exclusion into the live path

laci141:
The retraction detector (retracted.go) existed but was never called from
a real consensus fetch, so retracted works still contributed to the score.
A heavily cited retracted meta-analysis can outweigh several sound studies
purely on citation mass.

Two signals are threaded through:

1. A title-marker regex anchored at the start, matching RETRACTED/WITHDRAWN
   followed by a short separator (RetractionDeclared).
2. is_retracted from the OpenAlex index (RetractionFlagged). This field was
   absent from the select list, so it never arrived even after the detector
   was added. It is now requested explicitly, with a comment explaining that
   removing it silently degrades the gate to title-only detection.

filterRetracted is the single detection site. It calls DetectRetraction with
both signals, routes via ExcludeFromScore, and returns the excluded slice so
ExcludeFromScore works (excluded works stay visible in the output, flagged,
rather than vanishing).

The gate runs after the relevance filter and before enrichPubTypes in
consensus.go, for the same ordering rationale as the relevance gate:
excluded works must not inflate StudyCount, TotalCitations, apex design, or
dispersion. compare.go receives the same gate so batch and compare paths do
not diverge.

retracted_excluded is reported separately, matching the shape of the other
exclusion counters, so a reader can see how many works were set aside.

## Validation

- [ ] `python3 .github/scripts/verify-skill/verify_skill.py --dir library/other/scientific-consensus/`
- [ ] `go build ./...`
- [ ] `go vet ./...`
- [ ] `go test -count=1 ./...`
- [ ] `govulncheck ./...`
- [ ] `git diff --check`

`gofmt`: worktree is CRLF; checked only the 5 touched files against LF
copies — all clean.

Tests added in `internal/cli/retraction_gate_test.go`: title-marker only,
index-flag only, neither (kept), a false-positive anchoring case
("Retracted Science..."), select-list/normalization, and end-to-end scoring.


laci141:
Both review comments are addressed in ae06be4c2, and the branch is rebased
onto current main (the scwork.go conflict was #1822 landing in between; the
resolution keeps both the authors field and the retraction fields).

On the misclassified empty result: the note is now built by a single
function, consensusNote, and the call site is one line. The wording
distinguishes the two populations, which are not interchangeable — dropped
counts what the relevance gate removed from everything fetched, and the
retraction gate runs afterwards over what survived it. So a work counted as
retracted was by construction already judged relevant, and the empty-result
branch says "all N relevant work(s) were excluded as retracted" rather than
implying the whole fetch was discarded.

Fragments are joined from a slice rather than appended with a repeated
separator check; that repetition is what let the two stages drift apart in
the first place.

On the ledger: internal/scengine/retracted.go is added as a fifth entry.

One deliberate non-change: compare.go still says "one or both claims
returned no works". Its note is about two claims rather than one result
set and does not fit the consensusNote signature. The same misclassification
can occur there, so I would rather fix it properly in a follow-up than force
it into this shape.

tmchow:
The consensusNote fix is good. I'll take the leftover compare.go case (all-retracted still reads as "no works") so this can land.

tmchow:
Landed the leftover compare empty-result note on this branch (`ca7f0a777`), matching the `consensusNote` wording you already used.

When `StudyCount` is 0 but `RetractedExcluded` / `AllStudies` still list works, compare now says the relevant work(s) were excluded as retracted instead of "returned no works". Genuine empty fetches keep the original wording. Focused tests are in `compare_test.go`.

laci141:
Sorry for the force-push that dropped your compare.go fix (ca7f0a777).
Restored the branch. The fix is back on top.

laci141:
Tests updated in 57e10d4b2 — four assertions carried the old "relevant"
string, all four now match the "fetched" wording. `go test ./internal/cli/`
is green.

PR #1857, opened 2026-08-29
fix(scientific-consensus): raise claimStemLen to 6 so common prefixes stop matching

laci141:
## Summary

`claimStemLen` goes from 5 to 6, plus the corpus work that makes the change
measurable.

Not a CLI publish: no new package, no manifest, no manuscripts. The Published
CLI Metadata section of the template is omitted for that reason.

## Why five was wrong

Five characters is short enough that a claim's own token collapses onto a
prefix the literature uses for something else.

`intermittent fasting improves insulin sensitivity` stems its intervention
side to `inter`. That string appears in 52 of 57 archived works — as
*intervention*, *interaction*, *interval*, *interquartile*, *interplay*,
*internal*, *interpret*, *interested*. Never once as *intermittent*.

`coffee improves alertness` stems its outcome side to `alert`, which matches
the **Safety Alert** in a review of *Candida* species used as starter cultures
in fermented foods. That review scored as supporting evidence at confidence
0.90.

## Why the suite could not see it

The twelve corpora on main share a property nobody chose: every claim's
intervention side stems to something that occurs almost nowhere else —
`melat`, `saffr`, `vacci`, `coffe`. None of them collide with anything.

Measured: setting `claimStemLen` to **3** — collapsing coffee to `cof` and
alertness to `ale` — left the entire package green.

Two reasons. `TestTokenLeakage` computed its drift marker and sent it to
`t.Logf`, which `go test` discards on a passing run. And no corpus existed
where the collapse could show up.

## What Changed

| commit | |
|---|---|
| `22e86ed1` | archive `fasting`, a corpus whose stem is a common prefix |
| `490f29c5` | make leakage drift fail instead of log |
| `df3214de` | wire `vitaminc` in — it shipped with the gate but was never registered, and in the wrong envelope |
| `0aa86a6e` | raise the constant |
| `46e98ac7` | ledger entry |

Each is separately reviewable, and the fix lands on an instrument that can
already prove it works: `claimStemLen 6` now fails on `fasting` with
`5 -> 25 of 57`, and on nothing else.

## Validation

- [x] `go build ./...` from `library/other/scientific-consensus` — clean
- [x] `go vet ./internal/scengine/` — clean
- [x] `gofmt` on an LF copy of both touched `.go` files — `stance.go` clean;
      `corpus_regression_test.go` reports the same single pre-existing
      complaint it reports on `main` (an aligned struct field at the
      `omega3` entry), unchanged by this PR
- [x] mutation check: `claimStemLen = 6` fails `TestTokenLeakage` on
      `fasting` only; `claimStemLen = 3` used to pass silently and no longer
      can

Only `fasting` moves. Leakage `5/57 -> 25/57`; the gate's drop count `3 -> 20`,
nine of those decided on a full abstract rather than a truncated stump — at
five, **zero** of the three drops were trustworthy. No verdict changes on any
corpus. No stance changes on any surviving work.

`vitaminc` is the control: its glued phrase `[vitamin c]` is never truncated,
and it holds at `0/3` under both values.

Live replay at `--limit 100` on eight claims: coffee 21 works → 17, fasting 57
→ 39. Vitamin D, screen time, omega 3, meditation and probiotics unchanged in
both count and verdict.

The four works that leave the coffee run — the *Candida* review, a 1976
theophylline compliance study, a general "Coffee and health" overview, and a
coffee-berry-extract trial — were already being discarded by the downstream
LLM relevance filter. The gate now discards them deterministically instead.

## Gaps / Follow-up

`vitaminc.json` landed with the PICO gate in `64a9473f` but was never added to
`allCorpora`, and it was saved as the raw CLI result rather than the export
envelope. `loadCorpus` reads `response.result`, which that file did not have —
registering it would have yielded an empty `corpusResult` with no error. Fixed
here rather than left for later, since it belongs to the same file family. Its
data is unchanged; only the envelope is new.

`TestTokenLeakage` measures the claim's *first* content token, while the gate
keys on the whole intervention side (`[inter fasti]`, not just `inter`). The
two therefore disagree on multi-word interventions. Widening the measurement
was tried and rejected here: requiring every claim token would fail 19/19 on
melatonin and 12/12 on saffron, which measures nothing. Left as a known gap
rather than a half-fix.


laci141:
Ledger entry trimmed in 4396ec2, from 2852 characters to about a third of
that — in line with the rest of the directory.

What stays is the decision a regeneration needs: why five collides, why six
rather than seven, that the constant is shared with the stance classifier,
and that the corpus suite passing is not sufficient grounds to change it,
since it passed at claimStemLen 3.

The per-corpus counts, the live replay figures and the misfiled review's
confidence score moved to the commit messages and the PR description, where
they are dated and searchable.

PR #1860, opened 2026-08-29
fix(scientific-consensus): a host trial's randomisation is not this work's design

laci141:
## Summary

A nested study names the trial it sat inside, and the design heuristics cannot
tell whose design they just read.

Not a CLI publish: no new package, no manifest, no manuscripts.

## The case

`10.1371/journal.pone.0045231` — title ends **"A Case Control Study"**.

Its abstract says the participants were *"within a phase IV cluster-randomised
trial of HPV vaccination"*. The RCT heuristic matched that phrase, and the work
was promoted **eight tiers** — from case-control to the apex of the pyramid —
where it then carried a 0.90-confidence stance on a claim about HPV vaccines and
infertility.

## The guard

`hostTrialMention` matches a phrase crediting a **different** study for the
randomisation: `within`, `nested in`, `participants in`, `recruited from`,
`enrolled in`, `sub-study of`, `secondary analysis of`, `ancillary to` — each
followed by `a`/`an`/`the` and within 60 characters of a randomisation word.

`hostTrialOnly` then removes the matched spans — not the sentences they sit in —
and asks whether an RCT signal survives in what is left. The tier is skipped
only when none does, and only when the title does not itself say randomised.

Three narrowings, each load-bearing:

- **the determiner** separates a host trial from a work saying *"this randomised
  trial"* about itself
- **the span removal** lets one sentence name a host AND report this work's own
  allocation without losing the second
- **the title exception** stops a work that states its own design in the title
  from being overruled by anything in the abstract

## Review findings addressed

Four, all correct.

**Participant wording implied a false host.** The pattern treated
`participants in this randomised trial` the same as `participants in a
randomised trial`. The first is a work describing itself. Measured on the
saffron corpus, `enrolled in this prospective, comparative, randomized` was the
pattern's **only** match across 246 works — every match it had was a false one.
The determiner requirement excludes *this*, *our*, *the present* and *the
current* without listing them.

**The suppression was work-wide.** An abstract can name the trial its
participants came from AND describe its own randomisation, and one flag
discarded both.

**The suppression was then sentence-wide.** The first fix dropped the whole
sentence a host phrase sat in, which fails when both signals share one sentence.
`hostTrialOnly` now removes only the matched spans. On *"participants were drawn
from within a larger placebo-controlled trial and we then randomised 200 of them
in a double-blind design"* the remainder still reads *"and we then randomised
200 of them in a double-blind design"*, and the tier holds. This also drops the
sentence splitting, which was the weakest part of that version — splitting on
`.` cuts abbreviations in half, and the fragments were never the unit the
question was about.

**The ledger reason was six sentences.** Down to two, naming the three
narrowings that must survive a regeneration rather than restating the
measurements.

The 60-character bound stays a character count rather than becoming a word
count: `cluster-randomised` is one token, and a word-stepping pattern never
reaches inside it. A six-word version was tried and stopped matching the case
this guard exists for.

## Two wider rules, measured and rejected

| rule | works changed in `testdata/corpora` | direction |
|---|---|---|
| title beats title+abstract | 3 of 249 | all wrong |
| title beats publication type | 4 of 249 | all wrong |

Both demote works whose titles legitimately name two designs —
*"systematic review and meta-analyses"* becomes systematic review. The pyramid
order already picks the right one. Recorded in the ledger so the next reader
does not retry them.

## Also in this branch: gap-finding wording

`a246efe7` also touched `gaps.go`, which that commit's message does not mention.
`615bb875` records why rather than rewriting a pushed commit.

The gap findings state properties of the literature while measuring a sample.
`gaps` fetches by `relevance_score:desc` and reads the first 100 of what can be
120,000 matches. On *"vitamin C prevents the common cold"*: `gaps` reports 2022
as the latest year; `consensus` — same claim, same sort, 20 works — reports
2018. Two numbers, one literature.

Three of the four findings share this: `missing-strong-designs` reads the
sample's designs, `stale-evidence` its years, `understudied-population` its
text. Only `thin-literature` is safe, since it counts total matches rather than
the sample. The `well-covered` message needed it too — *"no obvious gaps
detected"* is as strong a claim as any of them.

Each now names the sample it read, and `stale-evidence` says outright that a
newer study ranking lower on relevance would not appear.

Re-sorting by `publication_date` was measured and rejected: on that claim it
returns five 2026 papers about SGLT2 inhibitors, tau seeding and meteorology.
Sorted by date, OpenAlex full-text search returns the newest *anything*, so
`stale-evidence` would never fire again — worse than the current state, not
better. The sampling limit is not fixable from inside this command; saying so
is.

## Validation

- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] `gofmt` on an LF copy of every touched file — clean

**Zero matches** across the 246 works in `testdata/corpora`. No design or
verdict change on any corpus.

Seven tests, each mutation-checked:

| mutation | fails |
|---|---|
| guard never fires | `IgnoresHostTrialRandomisation` |
| tier skip disabled | `IgnoresHostTrialRandomisation` |
| title exception removed | `KeepsRCTWhenTitleSaysSo` |
| determiner removed | `KeepsRCTWhenPhraseSaysThis` |
| work-wide flag restored | `KeepsRCTWhenOwnRandomisationIsSeparate` |
| sentence-level version restored | `KeepsRCTWhenBothSignalsShareASentence` |
| `hostTrialOnly` forced true | all four RCT-keeping tests |

## Known limit

This sits on the heuristic path. An authoritative publication type still wins
ahead of it, and OpenAlex indexes this particular work under its host trial's
type — so a live query still reports it as an RCT.
`TestClassifyDesignHostTrialGuardDoesNotOverridePubType` pins that rather than
leaving it to be discovered.

The guard helps the works that arrive without a publication type, which is most
of them: on one measured query, 34 of 100 retrieved works carried one.


laci141:
Addressed in 3c702548.

Correct finding. hostTrialOnly now removes the matched spans and asks whether an
RCT signal survives in what is left, instead of dropping the sentence they sit
in. On "participants were drawn from within a larger placebo-controlled trial
and we then randomised 200 of them in a double-blind design", the remainder
still reads "and we then randomised 200 of them in a double-blind design" and
the tier holds. On the Tanzania abstract nothing survives, and the tier is
skipped as intended.

This also drops the sentence splitting, which was the weakest part of the
previous version — splitting on "." cuts abbreviations in half, and the
fragments were never the unit the question was about. The function is shorter
than what it replaced.

Seventh test added, mutation-checked: restoring the sentence-level version fails
it and nothing else. Forcing hostTrialOnly to return true now fails four tests
rather than three.

laci141:
Closing this rather than taking a fifth narrowing pass at the design guard.

Four review rounds, four correct findings — the last one being that
`hostTrialMention` cannot separate "enrolled in a randomised controlled trial"
said about the work itself from the same phrase said about a host. That is
right, and I do not have a way to measure a fix for it: across the 246 works in
`testdata/corpora` the pattern matches **zero** times, so every rule I have
proposed in this PR was argued rather than measured.

That mattered three times already here. Title-priority looked correct and moved
3 works the wrong way. My "zero false positives" claim was wrong — the pattern's
one match in the corpora was a false one. Sentence-level scoping looked correct
and failed when both signals shared a sentence.

A guard that measurably helps nothing and can demote a genuine RCT is worse than
its absence, so it should not go in on argument alone.

The one clean part of this branch — the gap-finding wording, which is measured
and drew no review comment — is now its own PR: #<NEW>.

The design defect itself is real and I have opened #<ISSUE> with the case, the
four measurements, and the three approaches that were tried and rejected, so
whoever picks it up does not repeat them.

laci141:
Closing this rather than taking a fifth narrowing pass at the design guard.

Four review rounds, four correct findings — the last one being that
`hostTrialMention` cannot separate "enrolled in a randomised controlled trial"
said about the work itself from the same phrase said about a host. That is
right, and I do not have a way to measure a fix for it: across the 246 works in
`testdata/corpora` the pattern matches **zero** times, so every rule I have
proposed in this PR was argued rather than measured.

That mattered three times already here. Title-priority looked correct and moved
3 works the wrong way. My "zero false positives" claim was wrong — the pattern's
one match in the corpora was a false one. Sentence-level scoping looked correct
and failed when both signals shared a sentence.

A guard that measurably helps nothing and can demote a genuine RCT is worse than
its absence, so it should not go in on argument alone.

The one clean part of this branch — the gap-finding wording, which is measured
and drew no review comment — is now its own PR: #<NEW>.

The design defect itself is real and I have opened #<ISSUE> with the case, the
four measurements, and the three approaches that were tried and rejected, so
whoever picks it up does not repeat them.

laci141:
Two placeholders survived in the comment above.

The gap-finding wording is #1861.

There is no separate issue, and on reflection there should not be: everything
that would go in one is already here. The case is
`10.1371/journal.pone.0045231`, whose title ends "A Case Control Study" and
whose abstract puts its participants "within a phase IV cluster-randomised
trial of HPV vaccination" — eight tiers of promotion from one phrase, and a
0.90-confidence stance at the apex of the pyramid.

Three approaches were measured and rejected, each in a commit on this branch:

- title priority over title+abstract — moves 3 of 249 works, all wrong
- title priority over publication type — moves 4, all wrong
- a host-phrase pattern — matches 0 of 246 works in the corpora, so nothing
  about it could be measured, and four review rounds each found a real defect
  in it that measurement would have caught

The blocking constraint is that the pattern has no coverage in
`testdata/corpora`. Anyone picking this up needs a corpus where it fires before
a rule can be argued from evidence rather than from reasoning. That may be the
more useful first step than the guard itself.

Worth noting separately: on the live query this work is classified from
OpenAlex's publication type, not the heuristics, so no heuristic-path guard
would have changed its verdict anyway.

PR #1861, opened 2026-08-29
fix(scientific-consensus): make each gap finding name the sample it read

laci141:
## Summary

The gap findings state properties of the literature while measuring a sample.

`gaps` fetches by `relevance_score:desc` and reads the first 100 of what can be
120,000 matches, so *"most recent analyzed study is from 2022"* is a fact about
those 100 works, not about the field.

Four lines of wording. Not a CLI publish: no new package, no manifest, no
manuscripts.

## The measurement

On *"vitamin C prevents the common cold"*:

| command | works read | reports latest year |
|---|---|---|
| `gaps` | 100 | **2022** |
| `consensus` | 20 | **2018** |

Same claim, same sort. Two numbers, one literature. A reader acting on the
first would conclude there has been no work since 2022 and go looking for a gap
the sample invented.

## Which findings share this

| finding | reads | sample-bound |
|---|---|---|
| `missing-strong-designs` | the sample's designs | yes |
| `stale-evidence` | the sample's years | yes |
| `understudied-population` | the sample's text | yes |
| `thin-literature` | `total` matches | **no** |

`thin-literature` is safe because it counts what the query matched, not what was
retrieved. The `well-covered` message needed the same treatment — *"no obvious
gaps detected"* is as strong a claim as any of them, and the one most likely to
be trusted.

Each now names the sample it read, and `stale-evidence` says outright that a
newer study ranking lower on relevance would not appear.

## Re-sorting by date, measured and rejected

The obvious fix is a second query sorted by `publication_date:desc`. On this
claim that returns five 2026 papers about SGLT2 inhibitors, tau seeding, YouTube
recovery interviews, gene therapy and meteorology.

OpenAlex full-text search sorted by date returns the newest *anything*, so
`stale-evidence` would never fire again — worse than the current state, not
better.

The sampling limit is not fixable from inside this command. Saying so is.

## Validation

- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] `gofmt` on an LF copy — clean

No behaviour change: the same findings fire under the same conditions, and only
the wording differs.

### DIFF
--- a/library/other/scientific-consensus/.printing-press-patches/gap-findings-name-their-sample.json
+++ b/library/other/scientific-consensus/.printing-press-patches/gap-findings-name-their-sample.json
@@ -1,0 +1,13 @@
+{
+  "schema_version": 2,
+  "id": "gap-findings-name-their-sample",
+  "applied_at": "2026-08-29",
+  "base_run_id": "20260620-165352-e55cec69",
+  "base_printing_press_version": "4.24.0",
+  "summary": "Each gap finding names the sample it read. gaps looks at the works one relevance-ranked page returned, so a finding drawn from that page is a fact about those works and must not be worded as a fact about the literature.",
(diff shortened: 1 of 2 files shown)


laci141:
Added in the latest commit: `.printing-press-patches/gap-findings-name-their-sample.json`.

It records what a regeneration would undo — the four messages each stating a
property of the literature that was measured on one relevance-ranked page — and
the one approach that was tried and rejected, so the next reader does not
re-attempt the date-sorted query.

PR #1837, opened 2026-08-26
feat(kuma): add kuma

keithah:
## kuma

A focused Uptime Kuma v2 operator CLI for monitor inventory, heartbeat timelines, incident context, and one explicitly gated retry mutation. It uses Kuma's observed Socket.IO protocol, including streamed `heartbeatList` responses.

**API:** Uptime Kuma v2 internal Socket.IO | **Category:** monitoring | **Press version:** 4.0.0

**Spec:** internal integration; `spec_source: internal-yaml`; live protocol evidence is archived in `.manuscripts/20260826-kuma-rewrite/research/brief.md`.

### Publication Path

New print published from the `feat/kuma` branch into the catalog package at `library/monitoring/kuma/`.

### Novel Commands

| Command | Name | Description |
|---------|------|-------------|
| `incident-context` | Incident context | Combine monitor metadata and recent heartbeat history into an incident-oriented brief. |
| `set-retries` | Safe retry mutation | Change max retries only after an explicit `--yes` confirmation; otherwise produce a dry-run. |
| `heartbeats` | Socket.IO v2 client | Decode Kuma's streamed heartbeatList events and preserve monitor IDs from tuple envelopes. |

### CLI Shape

```text
kuma-pp-cli — Uptime Kuma v2 operator CLI

Usage:
  kuma-pp-cli <command> [flags]

Commands:
  health            connectivity + auth check
  monitors          list monitors (--query substring)
  heartbeats        recent beats across monitors (--hours, default 3)
  incident-context  composite brief for one monitor (--monitor id|name)
  set-retries       change a monitor's maxretries (dry-run unless --yes)
  version           print version

Global flags:
  --url, --username, --password   connection overrides (default from env)
  --json                          JSON output
```

### What This CLI Does

- Authenticates to Kuma v2 over Socket.IO.
- Lists and filters monitors.
- Retrieves and normalizes streamed heartbeat history.
- Produces incident context from monitor state and recent beats.
- Provides `set-retries`, dry-run by default, with a complete-object edit and readback verification when `--yes` is supplied.

### Manuscripts

- Research: `.manuscripts/20260826-kuma-rewrite/research/brief.md`
- Phase 5 acceptance: `.manuscripts/20260826-kuma-rewrite/proofs/phase5-acceptance.json`
- Publish live gate: `.manuscripts/20260826-kuma-rewrite/proofs/publish-live-gate.json`

### Validation Results

| Check | Result |
|---|---|
| Manifest | PASS |
[...]

PR #1883, opened 2026-08-30
docs(kuma): document native kumactl HTTP MCP integration

keithah:
## Kuma catalog docs

The published Kuma package is a generated Go CLI (`kuma-pp-cli`) with a custom Socket.IO client. It has no Python project metadata, `uptime-kuma-rest-api` dependency pin, or MCP transport integration point.

This PR documents the truthful split rather than inventing a Go dependency or rewriting the client:

- Keep `kuma-pp-cli` and its existing Socket.IO workflow unchanged.
- Point MCP users to the separately published native `kumactl` service.
- Document the required Streamable HTTP invocation:
  `kumactl-mcp --transport streamable-http --host 127.0.0.1 --port 40108 --path /mcp`
- Preserve `UPTIME_KUMA_URL`, `UPTIME_KUMA_USERNAME`, and `UPTIME_KUMA_PASSWORD` for both workflows.

### Validation

- `go test ./...`
- `go vet ./...`
- `go build ./...`
- `gofmt -l .` (no output)
- `git diff --check`
- `go run ./tools/generate-registry/main.go --validate` (466 entries)
- `go run ./tools/generate-skills/main.go --validate` (466 entries)
- Go module path invariant across `library/*/*/go.mod`


tmchow:
Thanks for the kumactl split docs and the 10 MiB Socket.IO read bound. Those two late commits are useful.

This PR itself is a full reprint of the kuma package against a pre-merge base, so GitHub reports it as 23 files / +3433 and it conflicts with main. The CLI already landed in #1837 on Aug 28, so I am not merging the reprint. I am extracting just the unique README/SKILL/client.go/client_test.go delta onto current main as an origin follow-up and will close this once that lands.

Credit stays with you on the leftover.

keithah:
whoops, sorry about that, thanks! I'll close this out then.

tmchow:
The leftover from this PR (kumactl HTTP MCP docs and the 10 MiB Socket.IO read bound) landed in #1886. Thanks for those two commits, @keithah.

PR #1896, opened 2026-09-01
feat(artistly): add artistly

tmchow:
Origin leftover from #1211 (@vcolombo).

Takes the Artistly CLI tree onto current main and drops the competing repo-root workflow / `.go-version` pins. phase5-acceptance.json is already a pass (6/6, cookie session).

Close #1211 as superseded after this merges.


tmchow:
Greptile came back 3/5 with two P1s I am not going to bypass: batch submit without `--wait` ignores concurrent/daily quota, and polling can pick up another session's finished designs. Library CI is green. I'll fix those on this leftover rather than land it as-is.

tmchow:
Quota pacing and prompt correlation landed on this leftover. Greptile is 4/5 with one remaining P1: two sessions submitting the same prompt after the same max design ID can still grab each other's result. That's a real race, but it is leftover round 2 and I'm not doing a third chase tonight. Library CI is green. Leaving this open for the babysitter rather than merging over that P1.

tmchow:
Closing this maintainer leftover after two fix rounds. The remaining identical-prompt concurrency P1 needs to be solved on the original #1211 before we replay it again.

PR #1899, opened 2026-09-01
feat(bonusly): fix broken user-lookup endpoints, validate mentions, require PAT-only auth

enlewof:
## Summary

Fixes 6 bugs found during a live dogfood session against bonusly-pp-cli,
plus one direct-input architecture change. All 7 findings were
independently verified against Bonusly's official, live API reference at
docs.bonus.ly (fetched during this run — the CLI's spec.yaml pre-dates
that reference and marks the affected paths as `# inferred`).

## Findings

| ID | Type | Summary |
|----|------|---------|
| F1 | bug | `users search` called nonexistent `/users/search?query=` (always 404). Real endpoint: `/users/autocomplete?search=`. |
| F2 | bug | `users get` called nonexistent `/users/show?identifier=` (always 404). Real endpoint: `/users/{id}` (path param, ObjectId only) — now resolves email/display-name input through autocomplete first. |
| F3 | bug | `org top --search` was a silent no-op (the real `/users` list endpoint has no search param) and `--cursor` pagination was dead (real API paginates via `limit`/`skip`). |
| F4 | bug | `give`/`recognition create` silently accepted free-text recipient mentions containing spaces, only failing opaquely at submit time; `--dry-run` gave a false-positive preview. |
| F5 | bug/doc | `give_amounts` reads like a hard allow-list but is UI-suggestion data only — verified live that the API accepts amounts outside that list. |
| F6 | bug | Browser-cookie session auth (`auth login --chrome`) authenticated GETs but 403'd with a CSRF error on every write (POST /bonuses). |
| F7 | direct-input / architecture | Removed browser-cookie auth entirely in favor of Personal Access Tokens. Bonusly's documented API ([help.bonus.ly](https://help.bonus.ly/en/articles/15264520-bonusly-api-access-for-admins)) confirms PATs are the sole auth mechanism — no cookie/session security scheme exists in the API at all. This retracts and deletes the prior `bonusly-cookie-session-auth.json` patch, whose customization (Chrome profile discovery, pycookiecheat shell-outs, cookie-jar seeding) no longer exists in this source. |

## Changes

```
 .../bonusly-cookie-session-auth.json               |  23 -
 library/productivity/bonusly/README.md             |  19 +-
 library/productivity/bonusly/SKILL.md              |  17 +-
 library/productivity/bonusly/internal/cli/auth.go  | 849 +--------------------
 library/productivity/bonusly/internal/cli/give.go  |   3 +
 .../productivity/bonusly/internal/cli/org_top.go   |  56 +-
 .../bonusly/internal/cli/recognition_create.go     |   5 +
 .../productivity/bonusly/internal/cli/users_get.go |  63 +-
 .../bonusly/internal/cli/users_search.go           |  47 +-
 .../bonusly/internal/client/bonusly_cookiejar.go   |  77 --
 .../productivity/bonusly/internal/client/client.go |   5 -
 .../bonusly/internal/cliutil/credentials.go        |   4 +-
 .../productivity/bonusly/internal/config/config.go |  66 +-
 13 files changed, 201 insertions(+), 1033 deletions(-)
```

Plus 4 new/1-retracted `.printing-press-patches/` entries and 1 new
in-repo plan doc under `.manuscripts/`.

## Verification

- `go build ./...` — PASS
- `go vet ./...` — PASS
- `cli-printing-press publish validate` — 11/12 checks PASS (manifest,
  transcendence, go mod tidy, module path, govulncheck, go vet, go build,
  `--help`, `--version`, verify-skill, manuscripts). The `phase5` source
  fingerprint check fails — **confirmed pre-existing**: it fails
  identically (same exit code, same "changed files" list) on a completely
  untouched, freshly-cloned copy of `upstream/main`, so this is not
  something this patch introduced or is responsible for fixing.
- Live smoke tests against the real Bonusly API (6/6 pass, independently
  re-verified after implementation):
  - `users search --search "Jane"` → real matching users returned
  - `users get <email>` → full user record (previously always 404)
  - `org top --search "<name>"` → correctly filtered subset (previously
    ignored, returned all 100 rows every time)
  - `give --to "Test Person" ... --dry-run` → fails fast with an
    actionable error instead of a false-positive preview
  - `auth login --chrome` → prints PAT onboarding instructions, touches
    no Chrome/sqlite3/pycookiecheat
  - `auth status` → still correctly reports credential presence via the
    bearer-token path only
- Documentation: `README.md` and `SKILL.md`'s Auth Setup sections
  rewritten to describe only the PAT flow.

## Evidence

- Plan doc: https://github.com/mvanhorn/printing-press-library/blob/f03c77cb958e9a3a87ad0e54077f55c047da9235/library/productivity/bonusly/.manuscripts/20260810-153232-4e95f81e/proofs/2026-09-01-171800-amend-bonusly-pp-cli-plan.md
- Patch manifests: https://github.com/mvanhorn/printing-press-library/tree/f03c77cb958e9a3a87ad0e54077f55c047da9235/library/productivity/bonusly/.printing-press-patches

## Notes

- `comp:bonusly` / `priority:P*` labels don't exist in this repo, so this
  PR is labeled `bug` + `enhancement` instead (both exist).
- No open or recently-merged PR overlapped with these findings (checked
  before starting).
- Names/emails in the plan doc and error-message examples are scrubbed
  placeholders (`jane.doe@example.com` etc.) — the session that surfaced
  these findings involved a real company's real employees; none of that
  is reflected in the shipped diff.


enlewof:
Fixed in 2a5fd8d — `org top --search` now walks the full top-level population (looping `limit=100` pages via increasing `skip` until a short page ends the walk) before filtering, instead of only scanning the single page selected by `--limit`/`--skip`. Live-verified by finding a user who sat at the tail end of what was previously the only page ever fetched. The no-search path is unchanged and still honors `--limit`/`--skip` as normal pagination.

enlewof:
Fixed in e021ea6 — extracted a shared `validateMentionReason()` helper and call it from both the `--reason` flag path and the `--stdin` JSON-body path, so the space-after-`@` check applies regardless of input source. Live-verified both paths now reject the same malformed pattern (`{"reason": "+15 @ test #teamwork"}` via stdin, and the equivalent `--reason` flag value) consistently.

Worth noting while re-verifying: this check catches a literally-empty mention token (a space immediately after `@`). It deliberately does *not* flag a bare single-word mention followed by a capitalized word (e.g. `@Colin Payne-Rogers`) — that pattern is syntactically identical to this CLI's own documented example (`+50 @jane.doe Great work...`), so a broader heuristic would false-positive on legitimate single-token mentions. That specific failure mode still surfaces as Bonusly's own live 400 error rather than a pre-empted client-side one on the `--reason`/`--stdin` surface. `give`'s `--to` flag (a structured, non-free-text input) has no such ambiguity and remains fully covered for F4's original motivating scenario. Documented in the patch record.

enlewof:
Fixed in 49cd8455 — this is a fair catch, and worth explaining why the fix changed shape rather than just patching the same check again.

The whitespace-after-`@` check (from the prior two commits) was correct as far as it went, but it only catches a **literally empty** mention token. It can't catch `@Jane Doe` because that pattern is syntactically identical to this CLI's own documented valid example: `+50 @jane.doe Great work...`. Both are "a token, then a space, then more text" — there is no syntactic rule that flags one without also flagging the other. I considered a capitalization heuristic (bare word + following capitalized word) and rejected it for the same reason: it would false-positive on that exact documented example.

Replaced it with a **live resolution check** instead, using the `/users/autocomplete` endpoint this same PR already fixed (F1): every `@token` extracted from the reason is looked up against the real user directory (username, full email, or email local-part, case-insensitive) before the mutating `POST /bonuses` can fire. This applies to `recognition create` (`--reason` and `--stdin`, both) and to `give --to` (which had the same latent gap for a bare unresolvable single token with no embedded space).

Live-verified end-to-end, not dry-run:
```
$ recognition create --reason "+15 @Jane Doe test #teamwork"
Error: mention "@Jane" does not match any user by username or email — did you mean
one of: Jane Chan (jchan@...); Jane Hanley (jhanley@...); Janek Nieberg (jnieberg@...)?
Use the full email or username, e.g. @jane.doe@example.com

$ echo '{"reason": "+15 @Jane Doe test #teamwork"}' | recognition create --stdin
Error: mention "@Jane" does not match any user by username or email — did you mean ...
```
Both rejected before any POST — no opaque 400, no false preview. A genuinely valid mention (tested against my own account's email) correctly passes the check and reaches the real API, which fails for an unrelated, expected reason (self-recognition) — confirming this doesn't over-block legitimate mentions.

One honest limitation, documented in the patch record: under `--dry-run`, this check can't actually run, because `--dry-run` intercepts *every* client request including reads (`client.Client.dryRun`) — there's no live round trip to validate against. The dry-run preview now says so explicitly ("could not verify ... running under --dry-run") rather than silently claiming a pass. The real submission is the enforcement point either way, and that path is fully covered.

enlewof:
Fixed both in ddcd818 —

**Dry-run skips mention enforcement** (`recognition_create.go`): the live check added last commit was itself intercepted by `--dry-run` (the shared client's `DryRun` flag short-circuits *every* request, reads included). `resolveMentionCandidate` now flips `c.DryRun` off for the duration of its own read-only lookup, then restores it immediately after — the mutating `POST /bonuses` later in the same command is still correctly blocked if `--dry-run` was requested, but the validation read now always genuinely reaches the live API. Verified:
```
$ recognition create --reason "+15 @Jane Doe test #teamwork" --dry-run
Error: mention "@Jane" does not match any user by username or email — did you mean one of: ...
```
Now caught under `--dry-run` too, not just live. A valid mention still shows a normal dry-run preview (verified no false-positive block).

**Search silently stops at 5,000** (`org_top.go`): added a `truncated: true` field in the JSON envelope plus a stderr warning when the 50-page cap is hit on a still-full page, so an incomplete result is never silently indistinguishable from a complete one. Verified the normal (non-truncated) case correctly omits the field. The 5,000-user cap path itself is verified by code inspection rather than a live reproduction — no company in the test account has anywhere near that many top-level users to exercise it end-to-end; noted in the patch record.

enlewof:
Fixed in f15a4c4 — the narrow pre-scan only checked for a literal space or tab immediately after `@`. Removed it entirely rather than extending its character list: `strings.Fields` (already used by the main validation loop) splits on *every* Unicode whitespace character, so the loop's existing empty-token check now catches an empty mention for any whitespace type — changed from a silent `continue` to a hard error.

Live-verified:
```
$ python3 -c 'import json; print(json.dumps({"reason": "+15 @\ntest #teamwork"}))' | recognition create --stdin --dry-run
Error: reason contains an invalid mention with no text after @ — use an email or username with no space, e.g. @jane.doe@company.com or @janedoe

$ recognition create --reason $'+15 @\x0btest #teamwork' --dry-run   # vertical tab after @
Error: reason contains an invalid mention with no text after @ — use an email or username with no space, e.g. @jane.doe@company.com or @janedoe
```
Both rejected. A valid mention still passes through to a normal dry-run preview, confirmed unaffected.

PR #1936, opened 2026-09-04
feat(concur): add browser-automation fallback for report creation on policy-gated tenants

### LINKED ISSUE #1935: feat(concur): add browser-automation fallback for report creation on policy-gated tenants
Captured during a /printing-press-amend run on concur-pp-cli.

Findings:
- F1 (feature): Live v4 API rejects `reports create` with HTTP 400 "policyId is required" on Concur tenants that require explicit expense-policy selection. No command in the CLI (account, api account, api reports, lists) exposes a policy ID lookup. The CLI's own manifest already names the fix template: hotels_search.go's agent-browser fallback pattern.

PR with the proposed fix follows (auto-detect-and-fallback: pure HTTP tried first, transparent browser-driven retry only on that specific validation-error shape).

enlewof:
## Summary

`reports create` fails live on Concur tenants that require explicit expense-policy
selection (HTTP 400 "policyId is required"), and no command in this CLI exposes a
policy ID to supply. This adds a transparent, auto-detecting browser-automation
fallback: the existing pure-HTTP path is tried first and unchanged for every other
tenant; only on that specific validation-error shape does it drive the real Concur
UI to complete the creation.

**Update 1:** a live end-to-end run surfaced a real bug -- a one-time-per-session
promotional interstitial dialog can render on top of the Create Report modal and
silently block the click. Fixed by mirroring `hotels_search.go`'s overlay-clearing
convention (close button, with an Escape-key fallback since the close click alone
can report success without actually dismissing the dialog).

**Update 2 (PR review):** Greptile flagged two P1 findings, both fixed:
- *Fallback Targets Wrong Region* -- host derivation silently defaulted to
  `us2.concursolutions.com` for any base URL it couldn't confirm. Fixed via
  `resolveReportsUIHost`: derive only from a trusted host, honor an explicit
  `CONCUR_UI_BASE_URL` override, hard-error otherwise.
- *Successful Creation Reports Failure* -- a failure after the irreversible click
  looked like a safely-retryable error, risking a duplicate report. Fixed via
  `reportsCreatePartialSuccessError`.

**Update 3 (PR review, security):** Greptile caught a real bug in Update 2's own
fix -- *Hostname Boundary Is Missing*. `resolveReportsUIHost`'s trust check used a
raw `strings.HasSuffix(h, "concursolutions.com")` with no domain-boundary
requirement, so a lookalike host like `evilconcursolutions.com` -- which literally
ends with that substring while being a totally unrelated domain -- was incorrectly
accepted as trusted, and the fallback would open it in the potentially-authenticated
browser. Fixed to require the exact domain or a `.""-prefixed suffix, matching
`client.go`'s `baseURLIsTrustedConcurHost` check this function's own docstring
already claimed to mirror "in spirit" but hadn't actually implemented correctly.
Two new regression tests: the bare lookalike, and the same lookalike surviving the
`www-`/`.api.` cleaning transform first.

## Findings

|---|---|---|---|
| F1 | report-creation-browser-fallback | feature | Live v4 API rejects `reports create` with HTTP 400 policyId required; the CLI's own manifest already names `hotels_search.go`'s agent-browser pattern as the template. |

## Verification

- Build/vet: PASS. Tests: PASS -- full suite, every package, 8 `resolveReportsUIHost` cases
  (including 2 security regressions), 2 interstitial scenarios, 1 partial-success-error case.
- `publish validate`: PASS on every check this change touches. 2 checks fail (phase5 fingerprint
  drift, verify-skill install-section drift) -- confirmed via git-stash comparison against a
  clean `upstream/main` that both fail identically with zero changes applied at every commit in
  this PR; pre-existing, unrelated.
- Live smoke: the original fallback and interstitial fix were both exercised end-to-end against
  the real tenant (created and submitted a real report). The region-guessing, duplicate-risk,
  and hostname-boundary fixes were driven by PR review rather than further live runs, each
  verified via unit tests written to fail against the bug first, then pass against the fix.
- Patch contract: 1 patch file (updated in place across all 4 commits), `// PATCH(...)` source
  comments at every changed call site.

## Evidence

- Patch record: https://github.com/enlewof/printing-press-library/blob//library/productivity/concur/.printing-press-patches/report-creation-browser-fallback.json
- Local plan doc (not in this PR): `~/printing-press/manuscripts/concur/amend-2026-09-04T2146/proofs/2026-09-04-amend-concur-pp-cli.md`
- `deferred_to_upstream` in the patch record notes the browser-fallback shape may generalize to
  any printed CLI hitting a tenant-config-dependent required field a generator can't discover
  ahead of time -- kept CLI-specific for now.

Closes #1935


tmchow:
Taking the remaining Greptile P1 (port-strip on CONCUR_BASE_URL host trust) on an origin leftover: #1943. Will close this as superseded once that lands.

PR #1942, opened 2026-09-05
feat(drug-enforcement): emit full code_info as wrapped Lots/Expiry

laci141:
openFDA enforcement records carry lot numbers and expiry dates in `code_info`, which the CLI dropped from human-readable output entirely.

Measured over 100 live records: median 50 chars, p90 261, max 829. Truncating at a fixed width loses data in ~22% of records. Unlike Product and Reason, which are descriptive prose that stays useful when clipped, `code_info` is identifying data — a partial lot number is not partially useful, and reads as a valid value when it is a fragment.

Adds `wrapField`, which word-wraps to the block's existing 16-column indent and never splits a token, so a lot number is never broken across lines.

## Summary

- `recallRecord` gains `CodeInfo` (`json:"code_info"`); the decoder previously discarded the field
- The record block prints a `Lots/Expiry:` line between `Product:` and `Reason:`
- `wrapField` wraps on whitespace only, counts runes, and lets an oversized single token overflow its line rather than hard-breaking it — splitting a lot number would defeat copy-paste and exact matching just as thoroughly as truncating it, and would do so silently
- `Product:` and `Reason:` are deliberately unchanged and keep `clip`

## Published CLI Metadata

Not applicable — this PR does not add or modify `library/<category>/<slug>/`.

## What Changed

`internal/cli/recalls.go`: struct field, one output line, `wrapField` plus its two width constants.
`internal/cli/recalls_test.go`: `TestWrapField`, table-driven, expectations hand-derived from the algorithm rather than captured from a run. Covers empty input, whitespace-only input, a value that fits, multi-line wrapping, an oversized single token, a zero-width geometry, and a multi-byte case where a byte count would wrap but a rune count does not.

## CLI Shape

Output shape only; no flags or subcommands added.

Before (`reference D-0617-2026`) — the field was absent entirely.

After:

```
[D-0617-2026] Class II
  Firm:         Direct Rx
  Initiated:    2026-06-22
  Status:       Ongoing
  Product:      DULOXETINE D/R, a) 30 mg (NDC 61919-482-30), 30 Caps…
  Lots/Expiry:  Lot: a) 09JA2530, 31JA2507, expires: 04/30/2027; b) Lot:
                09DE2412, 09JA2528, 29JA2511, expires: 04/30/2027.
  Reason:       CGMP Deviations: Presence of N-nitroso-duloxetine impurity…
```

The multi-line path on the longest `code_info` in a 100-record sample (D-0690-2026, CareFusion 213 / BD ChloraPrep, 965 runes, 96 lot numbers across 18 expiry groups):

```
  Lots/Expiry:  Lots: 4062535, 4065201, 4066446, Exp. 2/28/2027; 4180288,
                4197695, 4197697, 4180291, Exp. 6/30/2027; 4219454, 4221945,
                ...
                6111107, Exp. 4/30/2029.
```

Round-tripped mechanically rather than by eye: the raw `code_info` was re-fetched and the printed block's tokens re-joined and compared. Lossless on both records. Every continuation line starts at exactly column 16.

## What This CLI Does

Unchanged — wraps the openFDA `/drug/enforcement.json` endpoint with pre-built search expressions for recall-lookup workflows.

## Validation

- [x] `go build ./...` from the touched CLI root
- [x] `go vet ./...` from the touched CLI root
- [x] `go test ./...` from the touched CLI root — `internal/cli` passes including `TestWrapField`
- [ ] `python3 .github/scripts/verify-skill.py --dir library/<category>/<slug>/` — n/a, no skill dir touched
- [ ] `go run ./tools/generate-skills/main.go` — n/a, no `SKILL.md` changed
- [x] `git diff --check`

Five pre-existing test failures are untouched and unrelated: four in `internal/cliutil` and one in `internal/mcp`, all the same Windows home-directory resolution issue.

## Gaps / Follow-Up

The record block is now internally inconsistent: `Lots/Expiry:` reflows to 80 columns while `Product:` and `Reason:` still emit lines past it, which are in fact the widest lines in the output above. That is a deliberate scope call — this PR fixes a data-loss bug with one justification, and reflowing the prose fields is a separate cosmetic change. Happy to fold it in if you'd prefer the block coherent in one pass.

`strings.Fields` collapses internal whitespace runs, so D-0690's 965 runes print as 963 (two `";  "` became `"; "`). Whitespace runs are not identifying, and the same normalization protects the block from `code_info` values that carry embedded newlines. Flagging it because "prints the value in full" is true of the tokens, not byte-for-byte.

### DIFF
--- a/library/health/drug-enforcement/internal/cli/recalls.go
+++ b/library/health/drug-enforcement/internal/cli/recalls.go
@@ -14,6 +14,7 @@ import (
 	"net/http"
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	"github.com/mvanhorn/printing-press-library/library/health/drug-enforcement/internal/client"
 
@@ -45,6 +46,7 @@ type recallRecord struct {
 	ReasonForRecall      string `json:"reason_for_recall"`
 	DistributionPattern  string `json:"distribution_pattern"`
 	ProductDescription   string `json:"product_description"`
+	CodeInfo             string `json:"code_info"`
 	RecallInitiationDate string `json:"recall_initiation_date"`
 	ReportDate           string `json:"report_date"`
 	State                string `json:"state"`
@@ -258,6 +260,7 @@ func runRecallSearch(cmd *cobra.Command, flags *rootFlags, search string, limit 
 		fmt.Fprintf(w, "  Initiated:    %s\n", dash(normalizeRecallDate(r.RecallInitiationDate)))
 		fmt.Fprintf(w, "  Status:       %s\n", dash(r.Status))
 		fmt.Fprintf(w, "  Product:      %s\n", dash(clip(r.ProductDescription, 100)))
+		fmt.Fprintf(w, "  Lots/Expiry:  %s\n", dash(wrapField(r.CodeInfo, recallLineWidth, recallLabelWidth)))
 		fmt.Fprintf(w, "  Reason:       %s\n", dash(clip(r.ReasonForRecall, 100)))
 		fmt.Fprintln(w)
 	}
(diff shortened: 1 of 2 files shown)


tmchow:
Library Actions are still queued with no runner pick-up on this head, so parking this as awaiting-maintainer for now. Greptile already passed; once Verify / Govulncheck / SKILL (and the policy gate) go green I'll merge from the next recheck — nothing needed from you.

PR #1999, opened 2026-09-14
clinical-trials: make report a testable view

laci141:
`report` assembled its briefing inside `RunE` behind a live client, so nothing could assert that it reached `tallyPhases` at all. #1990 removed the four copies of the loop and named this as what it left open. This makes one of the three remaining callers reachable.

## What Changed

- `internal/cli/report.go`: `buildTopicReport(term string, total, recruiting int, trials []Trial, topN int) topicReport` now sits beside `topicReport`. The twelve lines that were in `RunE` moved into it byte for byte; `RunE` calls it. No behaviour change.
- `internal/cli/report_view_test.go`: new, hand-authored. Five assertions over a six-trial corpus.
- `.printing-press-patches/phase-tally-counts-every-sampled-trial.json`: the record now covers the country list too, and its closing sentence no longer says report is out of reach.

## Where the cut is

Everything above `buildTopicReport` in the command talks to the network — two `ctgovCount` calls and one `ctgovFetch`. Everything from there down is arithmetic on trials the caller already holds. That is the same split `buildTrialListView` makes in `recruiting.go`, and it is why recruiting's tally was the only one a test could reach.

`total` and `recruiting` stay parameters rather than being derived from the sample. They describe the whole literature; `len(trials)` is a different number, and the briefing prints all three side by side. A caller that passed `len(trials)` for `total` would look plausible on screen.

## Why not phaseTallyCorpus

That corpus exists to separate a per-trial phase tally from a per-entry one, and it carries no `Countries` at all — a country assertion written against it would pass no matter how the geo loop broke.

The new corpus carries every asserted field at zero, one and more than one: one trial with no countries, three with one, two with two or three; one trial with no phases, four with one, one with two. That is the multiplicity #1971 showed a single-valued fixture cannot test.

`Sponsor` is deliberately different: it reads one scalar per trial and can only ever contribute one. The test pins that, so if `Trial` ever gains a `Sponsors` slice the number stops matching and says so.

## Mutations

Each fails only what it should.

| Mutation | Result |
|---|---|
| geo loop restricted to a trial's first country | country total 8 → 5, Canada 3 → 1 |
| `SampleSize: total` instead of `len(trials)` | sample assertion alone |
| `geo.top(8)` hardcoded over `topN` | bound assertion alone |
| `len(t.Phases) == 0` → `== -1` in `tallyPhases` | **seven tests across three files**, including the new one |

The last row is the one that matters. `TestBuildTopicReportCountsPhaseEntries` fails alongside the four in `phase_tally_test.go` and the two in `tally_test.go`. It could not, if `report` had kept a private loop.

Mutations changed a condition rather than commenting out a block — an orphaned `}` stops the file compiling and proves nothing.

## Validation

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./... -count=1`
- [x] `govulncheck ./...` — no vulnerabilities found
- [x] `git diff --check` — clean
- [ ] `verify-still` — publish-flow check for new-CLI submissions; this is not one
- [ ] `generate-skills` — no SKILL.md change

`gofmt` was checked on the staged LF copy of both touched Go files, not the CRLF worktree.

`internal/mcp/tools_test.go` fails locally on Windows for an unrelated reason: the real `~/.config/clinical-trials-pp-cli/` exists on this machine and defeats the test's temp-dir override. Unchanged from #1971 and #1990, and green in CI.

## Gaps / Follow-Up

- `compare` and `emerging` still tally inside `RunE` behind a live client. Neither is the same shape as this one. `compare` already has `buildDrugProfile`, but its network calls and its arithmetic interleave — `ctgovCount` → field, `ctgovCount` → field, `ctgovFetch` → tally — and only six lines at the end are pure, against a function that already takes seven parameters. `emerging` tallies inside a loop shared with five other counters and a year cutoff. Each is its own change, and folding them in here would have held up a clean cut behind a contested one.
- The UI surplus note is unchanged: a reader seeing 311 beside a sample of 300 still needs the consuming application to explain it. Carried from #1973.


laci141:
Fixed in a3e90d155. The record was at ~3400 characters across summary, reason and validated_outcome, against a directory median near 1160 — changelog altitude, as AGENTS.md:290 puts it. Now 1801.

Dropped: the four-row mutation table, the 8→5 and Canada 3→1 figures, the six-trial corpus composition, the tirzepatide 244/282 reading. All of it lives in the commit messages and in #1990 and this PR, where it is dated and searchable.

Kept, as what a regeneration must not simplify away: that the distribution counts phase entries rather than trials; what N/A means and that the bucket is lossy by design; that one helper serves all four commands; that the country list follows the same per-entry rule while sponsor, reading a scalar, does not; and the fixture rule that a corpus where every item has exactly one of something cannot detect a per-item multiplicity bug.

Still above the median, but this record now carries the contract of three changes (#1971, #1990, this one) rather than one. Happy to cut the last two sentences of the summary if that is still too long.

